Issue #7, opened 2021-10-15
Should `epi_signal` inherit from `tsibble`?


ryantibs:
The [tsibble package](https://tsibble.tidyverts.org), as @jacobbien mentioned, looks like it has a structure that would be helpful to have `epi_signal` inherit, rather a plain `tibble`.  A `tsibble` has the following structure: 
- `key`: for us this would be `geo_value`, I think;
- `index`: for us would be be `time_value`, I think;
- any other columns as measured variables: for us, we need at least `value`, here.

Generally, we should also check out the [tidyverts](https://tsibble.tidyverts.org), and yes, that's not a typo, click on the link.  It would be nice to see if anything here just takes care of things we might need (or might have already implemented?) in `epitools`, and soon, in `epipred`.

@jacobbien @dajmcdon @capnrefsmmat @brookslogan Looking through all of this would be a big job, but do any of you guys just want to each take a look at some part, and report back what you think here, on this issue?

PR #9, opened 2021-10-15
add more complex slide example

### LINKED ISSUE #5: Add advanced slide example
We should add a more advanced example of sliding with a function to the [slide vignette](https://cmu-delphi.github.io/epitools/articles/epitools.html), very bottom.

Originally this was about sensorization (in the old `modeltools` vignette), but I think that this example should be in its own vignette, if it exists at all; see #6.

dajmcdon:
Closes #5. 

Added two "local forecasters" to the vignette.

1. The baseline.
2. An AR.

Both do arbitrary aheads. A few notes:
* I mentioned in the vignette that these will, by design, not account for data versioning.
* They use an odd amount of data. There is an arbitrary `min_train_window` argument in both functions that returns `NA` forecasts if the slide window is not long enough yet. After that, they do whatever. This seemed easiest to illustrate the behavior without too much data processing that would be needed in a real forecaster. Let me know if you have a better idea. I didn't bother to filter out the `NA`s this induces at the beginning of the series, but that doesn't seem key to the illustration either.

Two things that occur to me related to #7 : 
1. Is there any "filling" of missing `time_values`? `tsibble::fill_gaps()` fixes this, but I couldn't tell if it's part of the class.
2. Does the class enforce sequential `time_values`? Not clear to me if `tsibble` does this. But maybe we should?


ryantibs:
@dajmcdon Awesome!  I was just looking through the code now and these are nice examples of slide indeed. 

Couple more things come to mind.  (Sorry to drag this out ...)

1. Given the AR example, is there any reason to do the baseline example?  Doesn't seem to add much and it only makes the vignette longer.  How do you feel about commenting out the baseline for now (we could maybe reuse it elsewhere)?
2. What is going on with AR over the full year?  I just extended the plotting range and look what I get (below).  Something very weird is going on the point predictions in CA in January; do you have a sense for whether this is real (just AR behaving badly), or some numerical bug (matrix inversion problem?) in the implementation?  And what's going on with its quantiles (no blue fan) at various other points in time? 

Screen Shot 2021-10-16 at 11 13 05 PM

PR #22, opened 2021-11-09
Add `epi_archive`

brookslogan:
This is a work in progress.  Opening the PR just for any feedback and discussion.


brookslogan:
Weird thing is probably because
```
tibble:::`[.tbl_df`
```
looks like it copies the class and other attributes using `tibble:::vectbl_restore` (calling [`tibble_restore_impl`](https://github.com/tidyverse/tibble/blob/eac1191b62b27663091deb1a5edc016c6a5b0c7c/src/attributes.c)),
but
```
tsibble:::`[.tbl_ts`
```
does not look like it does the same.

R/slide.R
```diff
@@ -180,3 +194,62 @@ epi_slide_one_grp = function(.data_group, index_fun, f, ..., before_num,
...
+    tibble::add_column(
+              results.for.issue = . %>>%
+                dplyr::pull(issue) %>>%
+                purrr::map(function(issue) {
+                  max_time_value = issue_to_max_time_value(issue)
+                  min_time_value = max_time_value - time_step(n-1L)
+                  x$epi_df_as_of(issue) %>>%
+                    ## XXX we may want to `complete` the `time_value`s here
+                    dplyr::filter(min_time_value <= time_value & time_value <= max_time_value) %>>%
+                    dplyr::group_by(geo_value, !!!x$naming_info()$other.key.colnames) %>>%
```
brookslogan, line 246: `epi_archive` is lacking a `group_by` + `epi_slide` combination to match `epi_df`.  It was written assuming that `epi_slide` would have always been called on things grouped by the key and is written to do that implicitly, but that's not actually the case.  Maybe it'd be as simple as implementing a `group_by` method to check and store group column names and then using them here, not sure.

ryantibs: Thanks! My new implementation of `slide()` allows explicit specification of grouping variables, and defaults to key variables in the `epi_archive`.

ryantibs:
@lcbrooks I'm going to go ahead and merge this PR. I ended up doing a lot of work on this branch that I think will be valuable beyond archive that I just want to get this in the main branch, and take a break from the archive work while I pursue other things. Summary outside of the `epi_archive` work:

- Cleaned up a lot of `epi_df` and `epi_slide()` behavior so that they're simpler and more complete
- Generalized `epi_cor()` so that it takes two "by" arguments: `shift_by` and `cor_by`
- Lots of documentation improvements and updates

This archive work is now in a `wip` folder, and currently removed from the main package (so no users try to user it at this point). We should revisit it in another PR to finish it off. It is close to being finished. Summary of what's left:

- I think we need a `join()` method for `epi_archive` so it knows how to join on another data table onto its own one, and perform LOCF appropriately, update its metadata, etc. I want to do this at the data table level (rather than data frame or tibble level) because it should be faster
- Still need to test out the `slide()` method for `epi_archive` (which I reimplemented)
- Still need to finish off the archive vignette, probably the best idea just to repeat the case forecasting example from the slide vignette, but with doctor's visits as a covariate

PR #38, opened 2022-02-07
Qpmnguyen time agg vignette

ryantibs:
This is me working off of Quang's #37.


ryantibs:
@qpmnguyen great job. I passed over and made edits here and there. Apart from condensing and rearranging text/code blocks, a brief summary:

- Switched to Massachusetts and Vermont to demonstrate that if you use county name as the key, it'll fail, since they have overlapping county names
- Induced missingness by censoring according to small values, to make the point that zero-filling can be important for downstream stuff like trailing averages

Issue #41, opened 2022-02-13
Write something intelligent at the bottom of growth rates vignette


ryantibs:
I just recently implemented a `growth_rate()` function, which now merges/replaces any of the functionality for estimating relative changes and derivatives and the like, in the `epiprocess` package. The new [growth rate vignette](https://cmu-delphi.github.io/epiprocess/articles/growth_rate.html) likewise replaces the old relative change and derivatives vignettes.

At the [bottom of that vignette](https://cmu-delphi.github.io/epiprocess/articles/growth_rate.html#log-scale-estimation), I demonstrate how all the functionality works on the log scale. Something interesting happens: the relative change estimates look like they become less volatile and match the linear regression estimates much more closely. 

I want to write something intelligent here but I ran out of energy and intelligence. @jacobbien @dajmcdon You guys are intelligent guys, did you want to take a ponder and give it a shot?  

Also I would like to leave some intelligent general comment about when we might do things on the log scale versus not, based on our working assumptions about the noise model.

jacobbien:
Here's my interpretation for the first part of this issue:

> Something interesting happens: the relative change estimates look like they become less volatile and match the linear regression estimates much more closely.

Let's write `R = (A - B) / B` for the relative change (A = "after"; B = "before").

The question is why first differences on the log scale look less volatile.  To relate these two approaches, note that

`log(B) - log(A) = log(B/A) = log(1 + R)`.

We are doing averaging over a window, so the observation is that

`E[log(1+R)]` looks less volatile than `E[R]`.

There are two relevant inequalities for relating these, both based on the concavity of log:

`E[log(1+R)] <= log(1+E[R]) <= E[R]`.

The first inequality is Jensen's; the second inequality is because the tangent line of a concave function lies above it.

Here is the plot for GA, where 

`Red = E[log(1+R)]`,
`Green = log(1+E[R])`, and
`Blue = E[R]`.

![Screen Shot 2022-02-14 at 9 24 29 AM](https://user-images.githubusercontent.com/2008608/153914667-18ae678f-53da-416a-bfde-67dfec7c97a5.png)

It shows that both inequalities are relevant here, but that Jensen's accounts for more of the difference.

jacobbien:
I modified the code from the vignette to make the above plot.  Here's the code if you'd like to reproduce:

```
library(covidcast)
library(epiprocess)
library(dplyr)
library(tidyr)

x <- covidcast_signal("jhu-csse", 
[...]

PR #44, opened 2022-02-15
Add explanation for why rel_change_log appears less volatile than rel_change

jacobbien:
This is addressing one of the two points raised in issue #41.


ryantibs (approved):
Awesome, thank you Jacob!

Issue #26, opened 2021-11-11
validation for correct grouping structure


elray1:
As discussed in comments on PR #14, we are asking users to group their data according to any relevant `key`/identifier variables before calling functions like `slide` or `detect_outliers`, so that those functions know to operate on one group of observations at a time. It would be good to check at some point that users have done their grouping correctly, so that e.g. we're sliding over a series representing a signal for a single `geo_value` or a single combination of `geo_value` and `age_group`. The main way I can think of to check this is to verify that there are no duplicated values of `time_value` within the same group. I could imagine two ways this could go:

1. We could write a helper function for this validation that gets called from any epitools functions that require the correct grouping structure to make sense (like `slide` and `detect_outliers`).  For example, maybe something like this:

```{r}
#' Validate that there are not duplicate `time_values` within the same group of an `epi_tibble`
#'
#' @parameter x an epi_tibble object
#'
#' @return invisible logical, TRUE if grouping structure is OK. Otherwise, an error is thrown.
#'
#' @importFrom dplyr count pull
validate_epi_tibble_grouping = function(x) {
  # extract counts of the number of times each time_value appears within each group of `x`
  time_value_counts = x %>% count(time_value) %>% pull(n) %>% unique()

  # throw an error if any count other than (i.e., larger than) 1 appears within the same group
  if (!identical(time_value_counts, 1L)) {
    stop("x must not have duplicate values of `time_value` within grouping ",
         "levels. Do you need to add more grouping structure?")
  }

  return(invisible(TRUE))
}
```

Then, within `slide`, `detect_outliers`, `correct_outliers`, and any other relevant functions, we'd call `validate_epi_tibble_grouping(x)` before doing anything else.

Example output/start of test case:
```{r}
> x <- covidcast_signal(data_source = "jhu-csse",
+                       signal = "confirmed_incidence_num",
+                       start_day = "2020-06-01",
+                       end_day = "2021-05-31",
+                       geo_type = "state",
+                       geo_values = c("fl", "nj"),
+                       as_of = "2021-10-28") %>%
+   select(geo_value, time_value, cases = value) %>%
+   as.epi_tibble()
Fetched day 2020-06-01 to 2021-05-31: num_entries = 730
> head(x)
# A tibble: 6 × 3
  geo_value time_value cases
  <chr>     <date>     <dbl>
1 fl        2020-06-01   667
2 nj        2020-06-01   486
3 fl        2020-06-02   617
4 nj        2020-06-02   658
[...]

PR #50, opened 2022-02-25
minor bug fixes to outlier detection

elray1:
- add validation that `x` does not contain duplicate values to `detect_outlr`. This resolves the part of Issue #26 that relates to this function.
- prevent error when calling `detect_outlr` without having explicitly loaded the `tibble` package by using `tibble::...` when constructing a default argument value
- take more care when standardizing names of components returned by `feasts::STL`; hopefully will be more robust across versions of feasts (name of the seasonal column changed between the time I wrote this and now).
- prevent error in `dplyr::case_when` with disagreeing data types
- update outliers vignette so that the seasonal STL method uses seasonality; it got lost at some point
- add some (very) minimal unit tests for expected errors from `detect_outlr`


ryantibs:
Looks good, thanks! I notice that you still are working in your old `epitools` repo clone. I will go ahead and merge. 

Git is repointing this to the newly renamed `epiprocess` repo, but I think it would be safer (fewer potential problems downstream) if you just nuked and re-cloned `epiprocess`, assuming you don't have un-pushed work.

Issue #52, opened 2022-02-28
Redefine example forecasters to use control lists


ryantibs:
In the slide vignette, archive vignette, and advanced slide vignette, we should rewrite the example forecaster `prob_ar()` or `prob_arx()` so that all of its arguments outside of **data** (the variables used to produce the forecast) are passed through a control list. 

That is how we're currently thinking about setting things up in `epiprocess` so it'll make sense to set things up here for unification purposes.

PR #53, opened 2022-03-03
Fix var shifting in advanced.Rmd: by geo group, handling time gaps

brookslogan:
Fix problem with lags not being calculated by geo group.  Also edit to handle any time gaps within a geo group, or omissions at beginning or end of data set due to lag=0 not being included or covered by the range of shifts.

Current approach uses a lot of `full_join`s (potentially slow?) and makes `prob_arx` involve the separate `:=` functionality of both `rlang` and `data.table`, which could be confusing.  An alternative (below) would be using pivoting to perform the join, shown below.  But this would be slightly more fragile to derive other forecasters from, as it requires the variables to have the same type, and if a variable is mapped to 0 rows, it will be silently omitted rather than filled in with `NA`s. 

```
  # Build features and response for the AR model, and then fit it
  dat =
    # Make each (covariate,lag) request into a row of a tibble...
    tibble::tibble(i = seq_len(ncol(x)), lag=lags) %>%
    tidyr::unchop(lag) %>%
    # ... assign names we know will won't conflict...
    dplyr::mutate(name = paste0("x",seq_len(nrow(.)))) %>%
    # ... turn each of these requests into a separate tibble with indexing columns...
    dplyr::mutate(tbl_for_var_shift = purrr::pmap(., function(i, lag, name) {
      tibble::tibble(
                geo_value = geo_value,
                time_value = time_value + lag, # shifting values back by `lag` = shifting times forward by `lag`
                value = x[,i]
              )
    })) %>%
    dplyr::select(name, tbl_for_var_shift) %>%
    # ... as well as one for the desired shift of y...
    dplyr::bind_rows(tibble::tibble(
                               name = "y",
                               tbl_for_var_shift = list(tibble(
                                 geo_value = geo_value,
                                 time_value = time_value - ahead, # shifting variable forward by `ahead` = subtracting `lead` from time_values
                                 value = y
                               ))
                             )) %>%
    #  ... and combine them together.
    tidyr::unnest(tbl_for_var_shift) %>%
    tidyr::pivot_wider(id_cols=c(geo_value, time_value), names_from=name)
```


brookslogan:
The approach to handling time gaps sort of breaks the test time selection due to creating some extra `time_value`s that weren't in the original.  Due to the use of `nafill`, it should turn out the same, but still it seems like it's asking for trouble to leave it how it is.

brookslogan:
This still has time indexing issues.  The problem is that if we are making a forecast at `fc_time_value`, we are still only getting data through, at the latest, `fc_time_value-1L` (could also be even earlier depending on the data sources involved).  So the `test_time_value` does not match the `fc_time_value`, which is probably what we would have wanted.

Also, separately, in some debug examples, I'm seeing what looks like 14 days of latency in the CLI data.  I need to double-check whether I'm reading this correctly and whether it's another bug.

ryantibs:
So I just made a few small changes to the forecasters (third to last commit message gives details). Turns out turning off the intercept makes them look better overall. Will merge to main. Thanks very much @lcbrooks for the help (and for catching the bugs!!).

This closes #52 (although I only did it in the advanced vignette).

PR #59, opened 2022-03-21
Refactor vignettes with delphi-epidata-r

### LINKED ISSUE #56: Use `delphi.epidata` in vignettes ready
When the `delphi.epidata` package is "ready for primetime", we should use it instead of `covidcast` in the vignette examples. (Because it should be faster, and we want to encourage people to to start using that package anyway.)

dshemetov:
`delphi-epidata-r` works decently well as a drop-in replacement for these vignettes. Fixes #56 .


ryantibs:
Excellent, thank you. Can you: 

- update the DESCRIPTION to include yourself as a contributor 
- update the DESCRIPTION to include the dependency on `delphi.epidata` (@lcbrooks can help with he syntax ... at one point he put it in there, then we removed it)
- update the DESCRIPTION to remove the dependency on `covidcast` (we don't need it anymore right?)
- rebuild the vignettes/docs? this is achieved by running `pkgdown::build_site()` in an R session, where you set your working directory to be the base directory of this repo.

Now, a question: should we merge this PR while the `delphi.epidata` package is still pretty sparse with documentation & vignettes itself? Or should we leave it here until we've done sufficient work there? (The risk of leaving it here is that it'll get progressively out of sync with main and may be harder to merge eventually...?)

vignettes/epiprocess.Rmd
```diff
@@ -50,21 +50,23 @@ A data frame or tibble that has `geo_value` and `time_value` columns can be
 converted into an `epi_df` object, using the function `as_epi_df()`. As an
 example, we'll work with daily cumulative COVID-19 cases from four U.S. states:
 CA, FL, NY, and TX, over time span from mid 2020 to early 2022, and we'll use
-the [`covidcast`](https://cmu-delphi.github.io/covidcast/covidcastR/) package to
+the [`delphi-epidata-r`](https://github.com/cmu-delphi/delphi-epidata-r) package to
```
ryantibs, line 53: Please make sure this is <= 80 characters (I think you can just move "to" onto the next line).

vignettes/slide.Rmd
```diff
@@ -26,23 +26,26 @@ in this vignette.
 
 As in getting started guide, we'll fetch daily reported COVID-19 cases from CA,
 FL, NY, and TX (note: here we're using new, not cumulative cases) using the
-[`covidcast`](https://cmu-delphi.github.io/covidcast/covidcastR/) package, and
+[`delphi-epidata-r`](https://github.com/cmu-delphi/delphi-epidata-r) package, and
```
ryantibs, line 29: Similar (80 characters).

vignettes/aggregation.Rmd
```diff
@@ -14,24 +14,28 @@ COVID-19 cases in MA and VT.
...
-                      end_day = "2021-12-31",
-                      geo_type = "county", 
-                      geo_values = y$geo_value) %>% 
+x <- covidcast(
+  data_source = "jhu-csse",
+  signals = "confirmed_incidence_num",
+  geo_type = "county",
+  time_type = "day",
+  time_values = epirange(20200601, 20211231),
+  geo_values = paste(y$geo_value, collapse = ",")
```
ryantibs, line 33: Very minor, but the order you specify these four arguments seems to be a bit inconsistent across different calls to `covidcast()` in different vignettes.

I would just choose one preset order for these four arguments (all common arguments that get called in `covidcast()` across different vignettes). 

I happen to like this order: `geo_type`, `time_type`, `geo_values`, `time_values`, since it kind of fits with the column ordering in an `epi_df` (and its metadata).

vignettes/advanced.Rmd
```diff
@@ -192,32 +192,45 @@ to demonstrate the last point in a more realistic setting. First, we fetch the
...
-         percent_cli = value) %>%
+y1 <- covidcast(
+  data_source = "doctor-visits",
+  signals = "smoothed_adj_cli",
+  geo_type = "state",
+  time_type = "day",
+  time_value = epirange(20200601, 20211201),
+  issues = epirange(20200601, 20211201),
+  geo_values = "ca,fl"
+) %>% fetch_tbl()
```
ryantibs, line 208: Insert new line below this line.

ryantibs (commented):
Looks quite good, just left some minor comments.

dshemetov:
So the only dependency I couldn't remove was `covidcast::county_census` to get all the Massachusetts and Vermont counties. Eventually, we can probably port it over though.

vignettes/aggregation.Rmd
dshemetov: There is also the possibility of using the current default order in the package 
```r
delphi.epidata::covidcast(data_source, signals, time_type, geo_type, time_values, geo_values, as_of = NULL, issues = NULL, lag = NULL)
```

No real preference as to the order, but agree with consistency.

dshemetov:
If we merge this in, one benefit is getting more usage of the new R client and thus more testing. One drawback is that we may need to rewrite parts, if the client changes. I would vote to merge, since we will need to more testing and usage eventually and earlier is probably better (given the already-decent state of the new client).

vignettes/aggregation.Rmd
ryantibs: Sorry I'm so late to this. I didn't get any notifications that you were still working on this PR. 

Ok with either ordering as long as it's consistent. I see you've already done in a [7d95ad9](https://github.com/cmu-delphi/epiprocess/pull/59/commits/7d95ad939232c84ae9a849488b37ceec8f0fdf86) below. Thanks!

ryantibs (approved):
Sorry I'm so late on this, somehow I didn't get email notifications that you were still working on the PR.

Everything looks good; for some reason most of the colors in the graphics in the vignettes appears differently (they appear dark, less transparency?) to before; maybe somehow your graphics driver is different than mine (since I was the last person to rebuild the site). Any it's quite minor, just curious why that happened. I'll go ahead and merge.

Issue #42, opened 2022-02-13
Implement testing across the package


ryantibs:
Now that the package is getting farther along, I think it is about time we implement testing across the package. Especially because I believe (and hope) that more others will be contributing to it in the near future.

@capnrefsmmat You are our resident stickler-for-the-rules-kind-of-guy, did you want to take the lead on this? You know I'm kidding 😛, and I would be grateful if you took this on.  I think @qpmnguyen @rafaelcatoia could be willing to help as well, so if you have some ideas/scaffolding then perhaps they would be willing to implement it with you.

qpmnguyen:
Hello! Happy to help write unit tests!

ryantibs:
@qpmnguyen @rafaelcatoia Did you guys want to start taking a look at this? You can see the way that testing is implemented in the covidcast R package (for example) and also read up on testing in R in general if you haven't had exposure to this already.

In side conversation with @capnrefsmmat, he said he is busy for a little bit with other responsibilities, but can start taking a look perhaps mid March, so in any case I think it makes sense for Quant & Rafael to just start learning/putting things together in the meantime, if you're still up for it.  Thanks!

p.s. in the `advanced.Rmd` vignette, I have some code chunks with `include = FALSE` that were basically just some hand-tests I was running to make sure the results were what I expected with sliding. In building tests, you could remove these chunks from that vignette (they're not needed there anymore, they're just in there so I didn't forget them) and incorporate them into the tests you're writing.

dajmcdon:
@dshemetov @qpmnguyen @rafaelcatoia Is anyone actively working on this? I'm going to re-assign if no.

dshemetov:
I'm not.

kenmawer:
I submitted a pull request for a test, `correlation`.

kenmawer:
I just realized that the following do not have tests:
`archive.R`
`growth_rate.R`
`methods-epi_archive.R`
`slide.R` (may not be necessary as `epi_slide.R` is already tested)

brookslogan:
Tests for the above have now been addressed in PRs (e.g., #159, #197); closing as completed.  We may want to push on #204 to catch anything else hiding untested.

PR #85, opened 2022-05-25
Clean up a few more issues

### LINKED ISSUE #65: Improve `epi_slide` error messaging on bad `ref_time_values`
When the user inputs `ref_time_values` before or after the `time_value`s in the `epi_df` (even if there'd be data in the window), it gives a hard-to-understand error message:
```
Error: Can't subset columns that don't exist.
✖ Column `slide_value` doesn't exist.
Run `rlang::last_error()` to see where the error occurred.
```
We should instead detect this case and output an understandable error message.

Examples:
```
edf = bind_rows(
  tibble(geo_value = "ak", time_value = as.Date("2020-01-01") + 1:200, value=1:200),
  tibble(geo_value = "al", time_value=as.Date("2020-01-01") + 1:5, value=-(1:5))
) %>%
  as_epi_df()

f = function(x, ...) tibble(value=mean(x$value), count=length(x$value))

## --- These cases generate the error: ---

## `ref_time_values` + `align` that result in no slide data, generate the error:
edf %>% group_by(geo_value) %>% epi_slide(f, n=3L, ref_time_values=as.Date("2020-01-01")) # before the first, no data in the slide windows
edf %>% group_by(geo_value) %>% epi_slide(f, n=3L, ref_time_values=as.Date("2020-01-01")+207L) # beyond the last, no data in window

## `ref_time_values` + `align` that should? have slide data, but generate the error:
edf %>% group_by(geo_value) %>% epi_slide(f, n=3L, ref_time_values=as.Date("2020-01-01"), align="left") # before the first, but we'd expect there to be data in the window
edf %>% group_by(geo_value) %>% epi_slide(f, n=3L, ref_time_values=as.Date("2020-01-01")+201L) # beyond the last, but still with data in window

## --- This case doesn't generate the error: ---

## (this doesn't produce an error; the error appears only if the ref time values are out of the range for every group:)
edf %>% group_by(geo_value) %>% epi_slide(f, n=3L, ref_time_values=as.Date("2020-01-01")+200L)
```

### LINKED ISSUE #66: Improve documentation for `epi_slide`'s `f` parameter when it's a function with named arguments
```
edf = bind_rows(
  tibble(geo_value = "ak", time_value = as.Date("2020-01-01") + 1:10, x1=1:10, y=1:10 + rnorm(10L))
) %>%
  as_epi_df()

edf %>%
  group_by(geo_value) %>%
  epi_slide(function(x, method="qr", ...) tibble(model=list(lm(y ~ x1, x, method=method))), n=7L)
```
produces
```
An `epi_df` object, with metadata:
* geo_type  = state
* time_type = day
* as_of     = 2022-04-05 12:05:18

# A tibble: 10 × 5
# Groups:   geo_value [1]
   geo_value time_value    x1      y slide_value_model
 * <chr>     <date>     <int>  <dbl> <list>           
 1 ak        2020-01-02     1 -0.128 <lm>             
 2 ak        2020-01-03     2  2.35  <lm>             
 3 ak        2020-01-04     3  2.65  <lm>             
 4 ak        2020-01-05     4  4.94  <lm>             
 5 ak        2020-01-06     5  4.00  <lm>             
 6 ak        2020-01-07     6  6.72  <lm>             
 7 ak        2020-01-08     7  6.33  <lm>             
 8 ak        2020-01-09     8  6.89  <lm>             
 9 ak        2020-01-10     9  8.66  <lm>             
10 ak        2020-01-11    10 10.0   <lm>             
Warning messages:
1: In lm(y ~ x1, x, method = method) :
  method = 'ak' is not supported. Using 'qr'
2: In lm(y ~ x1, x, method = method) :
  method = 'ak' is not supported. Using 'qr'
3: In lm(y ~ x1, x, method = method) :
  method = 'ak' is not supported. Using 'qr'
4: In lm(y ~ x1, x, method = method) :
  method = 'ak' is not supported. Using 'qr'
5: In lm(y ~ x1, x, method = method) :
  method = 'ak' is not supported. Using 'qr'
6: In lm(y ~ x1, x, method = method) :
  method = 'ak' is not supported. Using 'qr'
7: In lm(y ~ x1, x, method = method) :
  method = 'ak' is not supported. Using 'qr'
8: In lm(y ~ x1, x, method = method) :
  method = 'ak' is not supported. Using 'qr'
9: In lm(y ~ x1, x, method = method) :
  method = 'ak' is not supported. Using 'qr'
10: In lm(y ~ x1, x, method = method) :
  method = 'ak' is not supported. Using 'qr'
```
because the unnamed grouping-variables tibble (here, it'd be `tibble(geo_value="ak")` for our single group) is being passed in as the `method` argument to `f`, overriding the default `qr` (plus some "interesting" behavior of `==` and `sprintf` on tibbles going on inside `lm` when it checks it).

### LINKED ISSUE #68: Add final newline to epi_archive print statement
Currently, epi_archive doesn't print a final newline and leaves the REPL prompt or future output hanging in a weird position.  Add a trailing newline to fix.

dajmcdon:
* Closes #65 
* Closes #66 
* Closes #68

### DIFF
--- a/R/archive.R
+++ b/R/archive.R
@@ -16,7 +16,7 @@
 #' @details An `epi_archive` is an R6 class which contains a data table `DT`, of
 #'   class `data.table` from the `data.table` package, with (at least) the
 #'   following columns:
-#'
+#' 
 #' * `geo_value`: the geographic value associated with each row of measurements.
 #' * `time_value`: the time value associated with each row of measurements.
 #' * `version`: the time value specifying the version for each row of
@@ -31,7 +31,7 @@
 #'   on `DT` directly). There can only be a single row per unique combination of
 #'   key variables, and thus the key variables are critical for figuring out how
 #'   to generate a snapshot of data from the archive, as of a given version.
-#'
+#' 
 #' In general, last observation carried forward (LOCF) is used to data in
 #'   between recorded versions. Currently, deletions must be represented as
 #'   revising a row to a special state (e.g., making the entries `NA` or
@@ -43,7 +43,7 @@
 #'   reference semantics. A primary consequence of this is that objects are not
 #'   copied when modified. You can read more about this in Hadley Wickham's
 #'   [Advanced R](https://adv-r.hadley.nz/r6.html#r6-semantics) book.
-#'
+#' 
 #' @section Metadata:
 #' The following pieces of metadata are included as fields in an `epi_archive`
 #'   object:
@@ -75,7 +75,7 @@
 #'   sliding computation at any given reference time point t is performed on
 #'   **data that would have been available as of t**. More details on `slide()`
 #'   are documented in the wrapper function `epix_slide()`.
-#'
+#' 
 #' @importFrom R6 R6Class
 #' @export
 epi_archive =
@@ -89,7 +89,7 @@ epi_archive =
           additional_metadata = NULL,
 #' @description Creates a new `epi_archive` object.
 #' @param x A data frame, data table, or tibble, with columns `geo_value`,
-#'   `time_value`, `version`, and then any additional number of columns.
+#'   `time_value`, `version`, and then any additional number of columns. 
 #' @param geo_type Type for the geo values. If missing, then the function will
 #'   attempt to infer it from the geo values present; if this fails, then it
 #'   will be set to "custom".
@@ -105,12 +105,12 @@ epi_archive =
 #' @return An `epi_archive` object.
 #' @importFrom data.table as.data.table key setkeyv
           initialize = function(x, geo_type, time_type, other_keys,
-                                additional_metadata) {
+                                additional_metadata) {  
             # Check that we have a data frame
             if (!is.data.frame(x)) {
               Abort("`x` must be a data frame.")
             }
-
+                  
             # Check that we have geo_value, time_value, version columns
             if (!("geo_value" %in% names(x))) {
               Abort("`x` must contain a `geo_value` column.")
@@ -121,7 +121,7 @@ epi_archive =
             if (!("version" %in% names(x))) {
               Abort("`x` must contain a `version` column.")
             }
-
+              
             # If geo type is missing, then try to guess it
             if (missing(geo_type)) {
               geo_type = guess_geo_type(x$geo_value)
@@ -131,7 +131,7 @@ epi_archive =
             if (missing(time_type)) {
               time_type = guess_time_type(x$time_value)
             }
-
+            
             # Finish off with small checks on keys variables and metadata
             if (missing(other_keys)) other_keys = NULL
             if (missing(additional_metadata)) additional_metadata = list()
@@ -145,7 +145,7 @@ epi_archive =
                     c("geo_type", "time_type"))) {
               Warn("`additional_metadata` names overlap with existing metadata fields \"geo_type\", \"time_type\".")
             }
-
+            
             # Create the data table; if x was an un-keyed data.table itself,
             # then the call to as.data.table() will fail to set keys, so we
             # need to check this, then do it manually if needed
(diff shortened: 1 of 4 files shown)


dajmcdon:
Note: there's some weird formatting. I'm going to clean that up after (to avoid continued maintenance on 2 forks).

brookslogan:
- Confirming that the improved error messaging works to resolve #65.  But beyond #65 we should probably be more stringent on `ref_time_values` and require them to all be in `unique(x$time_value)` rather than silently filtering to such values + disallow passing in 0 `ref_time_values`.  That should also give users an easier time debugging as they don't have to track down what `starts` and `stops` are.
  - ---> This can be filed in a new Issue & addressed later.
- Re. #66:
  - The main `@param f` documentation is still wrong and doesn't reference the clarifying `@details` in this section: "If a function, `f` must take `x`, a data frame with the same column names as the original object; followed by any number of named arguments; and ending with `...`". I think the proper thing may be that `f` should take `x`, `g`/`k`/`key`/think-of-a-better-name, and named args (dots don't seem necessary; I think originally they were just there to absorb the unlisted `g`).
  - In `@details`, the following text too specific to the particular example that follows: "if a tibble with an unnamed grouping variable is passed in as the **method** argument to `f`".  This explanatory text may be unnecessary anyway once the main `@param f` documentation is fixed.  The dots in the example may be unnecessary.  The example code could also use a `library(dplyr)`, and these interspersed details&code could be turned into `@examples` comments&code so users can more easily run them and so package checks ensure that these examples run.  (Separately, as part of #42, maybe these examples should be turned into tests as well?)
- Re. #68: `archive_cases_dv_subset` needs to be regenerated or reworked to form an `epi_archive` from more basic parts on demand; it currently appears to load with the old class definition, missing the extra "\n", plus some bad srcrefs that may make user debugging of tweaked examples harder:
```
archive_cases_dv_subset$print
<srcref: file "C:/Users/chloe/Desktop/Summer2022/epiprocess/R/archive.R" chars 162:19 to 187:11>
<environment: 0x55c5b02d6358>
Warning message:
In getSrcLines(srcfile, x[7L], x[8L]) :
  restarting interrupted promise evaluation
> body(archive_cases_dv_subset$print)
{
    cat("An `epi_archive` object, with metadata:\n")
    cat(sprintf("* %-9s = %s\n", "geo_type", self$geo_type))
    cat(sprintf("* %-9s = %s\n", "time_type", self$time_type))
    if (!is.null(self$additional_metadata)) {
        sapply(self$additional_metadata, function(m) {
            cat(sprintf("* %-9s = %s\n", names(m), m))
        })
    }
    cat("----------\n")
    cat(sprintf("* %-14s = %s\n", "min time value", min(self$DT$time_value)))
    cat(sprintf("* %-14s = %s\n", "max time value", max(self$DT$time_value)))
    cat(sprintf("* %-14s = %s\n", "min version", min(self$DT$version)))
    cat(sprintf("* %-14s = %s\n", "max version", max(self$DT$version)))
    cat("----------\n")
    cat(sprintf("Data archive (stored in DT field): %i x %i\n", 
        nrow(self$DT), ncol(self$DT)))
    cat("----------\n")
    cat(sprintf("Public methods: %s", paste(names(epi_archive$public_methods), 
        collapse = ", ")))
}
``` 
The quick fix is to just regenerate the object + file an Issue to later figure out how to generate the example epi_archive on demand or otherwise ensure that it is current with the implementation of epi_archive.  The less-quick fix is to try to immediately address this issue.

brookslogan:
Re. #68 and regenerating / syncing the example archive object: the simple way is just to re-run to regenerate the archive object.  The more complex way is to store its DT and reconstruct the archive on demand; I've implemented this as part of 8f36d44c2379a847da27a642e335eee7a4b17c48, but based on an older version/name of the archive object, and involving the  compactify parameter under development in #101, so it will need a little tweaking.

PR #92, opened 2022-06-03
Km/growth rate

kenmawer:
Pull request to update growth_rate.


ChloeYou (changes requested):
I believe `data/ca_daily_cases.rda` isn't actually used? Then it can be deleted in this PR.

R/growth_rate.R
```diff
@@ -100,6 +100,16 @@
 #'   user.
 #' 
 #' @export
+#' @examples
```
ChloeYou, line 103: The internal dataset name is now called `jhu_csse_daily_subset`. The following example can be rewritten as the following:

```
# COVID cases growth rate by state using default method relative change
jhu_csse_daily_subset %>% 
  group_by(geo_value) %>% 
  mutate(cases_gr = growth_rate(x = time_value,  y = cases))

# Log scale, degree 4 polynomial and 6-fold cross validation
jhu_csse_daily_subset %>% 
  group_by(geo_value) %>% 
  mutate(gr_poly = growth_rate( x = time_value, y = cases, log_scale = TRUE,  ord = 4,  k = 6))
```
Afterwards call `devtools::document()` to update `man/growth_rate.Rd`, I think then the R-CMD-checks should pass.

ChloeYou (changes requested):
Update example in growth_rate.R to adjust for changes in the main branch.

R/growth_rate.R
kenmawer: I updated them. Sorry I didn't see this until now; the comment was on a weekend when I didn't check.

kenmawer:
I updated them based on Chloe's comment. Sorry I didn't see this until now; the comment was on a weekend when I didn't check.

ChloeYou (approved):

PR #101, opened 2022-06-14
Km compactify rectify

kenmawer:
I rectified compactify by overhauling it.


R/archive.R
```diff
@@ -145,14 +150,64 @@ epi_archive =
...
+            }
+            
+            # Warns about redundant rows
+            if (is.null(compactify) && nrow(elim) > 0) {
+              warn <- paste("\nLOCF rows found. To remove warning, set",
+              "compactify to TRUE or fix these rows: \n")
+              # call elim with for loop, up to 6
+              for (i in 1:min(6,nrow(elim))) {
+                warn <- paste0(warn,
+                  paste(elim[[i,1]],elim[[i,2]],elim[[i,3]],elim[[i,4]],"\n")
```
brookslogan, line 202: There might be more than four columns; you can might try `paste(collapse="\n",capture.output(print(elim)))` or something along those lines to print however many columns there are.

R/archive.R
```diff
@@ -145,14 +150,64 @@ epi_archive =
...
+                )
+              }
+              if (nrow(elim) > 6) {
+                warn <- paste0(warn,"And so on...")
+              }
+              warning(warn)
```
brookslogan, line 208: We want to use rlang conditions for errors, warnings, and messages in this package.  Our helper function `Warn` in its current form might mess up the display of `elim` because it does line (re-)wrapping; I'm working on augmenting it to be more convenient in #102.  But for now, maybe just use `rlang::warn` instead of `warning`, and rename the `warn` to something like `warning_message` to prevent confusion.

tests/testthat/test-compactify.R
```diff
@@ -0,0 +1,62 @@
+library(epiprocess)
+library(data.table)
+library(dplyr)
+
+dt <- archive_cases_dv$DT
```
brookslogan, line 5: Expecting an uncompactified `archive_cases_dv$DT` here will probably break the tests if `archive_cases_dv` is re-built down the line with the default `compactify` setting.  Maybe update the `archive_cases_dv` generation code to use `compactify=FALSE`, or use `TRUE` and make the code below insert some LOCF rows or something?

tests/testthat/test-compactify.R
```diff
@@ -0,0 +1,62 @@
...
+test_that("LOCF values are ignored with compactify=FALSE", {
+  expect_identical(nrow(dt),nrow(dt_false))
+})
+
+test_that("LOCF values are taken out with compactify=TRUE", {
+  dt_test <- as_tibble(as_epi_archive(dt[-c(21,22,40),],compactify=FALSE)$DT)
+  
+  expect_identical(dt_true,dt_null)
+  expect_identical(dt_null,dt_test)
+})
```
brookslogan, line 62: We probably a test here that the `$as_of` operations remain the same for various queried `version`s, whether compactified or not.

R/archive.R
```diff
@@ -145,14 +150,64 @@ epi_archive =
...
+              DT = rm_locf(DT)
+            } else {
+              # Create empty data frame for nrow(elim) to be 0
+              elim = tibble()
+            }
+            
+            # Warns about redundant rows
+            if (is.null(compactify) && nrow(elim) > 0) {
+              warn <- paste("\nLOCF rows found. To remove warning, set",
+              "compactify to TRUE or fix these rows: \n")
```
brookslogan, line 198: New users are likely to see this the first time they try to make an epi archive; this message may not be explanatory enough for them.  Maybe re-use some of the `@param` documentation.  "Fix" might be misleading; I think the most common thing is just that many recording mechanisms would naturally have some or a lot of redundant rows, and if those were never generated or read it might make the entire pipeline more efficient; or, maybe they can't be removed upstream and they can be removed here for space efficiency; or, probably less likely, these redundant rows are faulty records and there's some bug upstream the data pipeline.

vignettes/archive.Rmd
```diff
@@ -117,6 +147,9 @@ object:
 * `geo_type`: the type for the geo values.
 * `time_type`: the type for the time values.
 * `additional_metadata`: list of additional metadata for the data archive.
+* `compactify`: TRUE eliminates redundant entries using LOCF for efficiency;
+no entry also brings up a message to indicate which columns should be removed
+for efficiency; and FALSE does not eliminate redundant entries.
```
brookslogan, line 152: I don't think `compactify` is actually part of the metadata:
```
# `dt` here is taken from the tests
as_epi_archive(dt,compactify=TRUE)$geo_type # "state"
as_epi_archive(dt,compactify=TRUE)$compactify # NULL
```

Two approaches:
- Just remove this line.
- Try to make most of this line true.  Just adding `compactify` as a field would be a bit confusing; I'd expect `my_archive$compactify` to be a function that would perform compactification (--- that might be an idea to implement later).  So instead the metadata field would probably need to be called something like `compactified`.

R/archive.R
```diff
@@ -145,14 +150,64 @@ epi_archive =
...
             # then the call to as.data.table() will fail to set keys, so we
             # need to check this, then do it manually if needed
             key_vars = c("geo_value", "time_value", other_keys, "version")
             DT = as.data.table(x, key = key_vars)
             if (!identical(key_vars, key(DT))) setkeyv(DT, cols = key_vars)
-
+            
+            # Checks to see if a value in a vector is LOCF
+            is_locf <- function(vec) {
+              ifelse(!is.na(vec) & !is.na(lag(vec)),
```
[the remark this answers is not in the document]
kenmawer: Yes. I fixed it.

I also fixed this.

Are you saying that I should hard-code this as to never eliminate the first row under any circumstances? The first row is not eliminated as long as at least one value is not NA. For example, if both `geo_value` and `time_value` are not NA, but the other values are, the row will still be there.

R/archive.R
kenmawer: I fixed this and made the layout better. Thank you for telling me about `capture.output.` I wasn't aware of it until now.

R/archive.R
kenmawer: I replaced the word "fix." Perhaps a good idea would be to put on the vignette about what may be inherently wrong with having LOCF rows.

tests/testthat/test-compactify.R
kenmawer: I noticed that the dataset stored in R as an rda in data is different from what happens if I were to run the R code. It works for now; however, I will need to update it when it gets fixed to allow for LOCF to be used.

vignettes/archive.Rmd
kenmawer: I updated this on the documentation.

tests/testthat/test-compactify.R
kenmawer: Added.

R/archive.R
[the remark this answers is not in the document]
kenmawer: I actually found another problem with the code altogether when the first row is comprised entirely of NA's:

image

It's as if we should assume `time_value` and `version` should be non-NA.

tests/testthat/test-compactify.R
brookslogan: Sorry, I didn't see that you were working on this and set something up for this.  It rebuilds the archive from a DT whenever the user first requests it, and uses compactify=FALSE to keep the current tests running.

R/archive.R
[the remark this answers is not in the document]
brookslogan: Yes, I think the hard-code to never eliminate the first row is the "fix".  This is basically saying that lagged first row has a different geo_value etc. from the first row, just how it works on the first row of later geo_value etc. combinations.

Nice catch on the time_value and version NAs.  We can keep it as the separate issue you filed.

R/archive.R
brookslogan: I'm tweaking this message and some of the documentation.  Then I will try to make that version of the documentation true by addressing #109.  Then I think this will be good to go!

R/archive.R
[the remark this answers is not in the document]
dajmcdon: Moved to #106

brookslogan:
Merged back in some changes that were deleted due to object renames + merge conflict resolution.  Closes #125 but doesn't address the double memory usage part.  Added `clobberable_versions_start` and `observed_versions_end` params to let `as_of` results be the same when using `compactify=TRUE` and `compactify=FALSE` in hopefully the last edge case.  But `epix_merge` needs to be edited to take these fields into account, and probably have some better defaults since we are basing `as_of` / all `epi_archive` operations on last-version-carried-forward.  And accounting for these the `observable_versions_end` parameter is actually a bit of a mess because the use may want to do various things, and it will matter, e.g., when making production forecasts.

brookslogan:
Migrating note from discussion in meeting: sysdata.rda contains internal / non-exported data of the package. This is working as intended.

brookslogan:
I just pushed some updates to `epix_merge` to make it work with the field additions (clobberable versions start, observed versions end) used to make compactify satisfy the invariants we wanted.  It still needs some more tests and the vignettes will need updated.  + debug of the test failures and merge conflicts (these are just Rd updates, we can probably just accept the new versions).

brookslogan:
Okay, I think I've finished up my "minor tweak" to make compactify more compatible with other operations + assorted bugfixes and clarifications along the way.  @kenmawer please review my changes; once your feedback is addressed I think this will be ready to merge.

brookslogan:
Cleaned up some bad tidyselect usage in the tests.  I was a bit dubious when they appeared to work; should have investigated further.

brookslogan:
Closes #111.  Closes #136.

kenmawer (commented):
Looks good to me. I'd say that on a separate PR, we can specify examples, etc. on vignettes about what things such as clobberable versions mean.

brookslogan:
Okay, discussing with Ryan, we might need to roll back some of my changes, at least.

kenmawer:
The file is filled with random changes, which makes it hard to track down what had actually been changed within this pull request rather than outside of it.

brookslogan:
The `Files changed` helps with this, or using `commit1..commit2` or `commit1...commit2` with git commands.  Of course, it probably would have been better if for some of these incidental bugfixes if I just made checks for the edge cases they addressed and complained and made the fixes in separate PRs.  Then we'd be able to merge more quickly and have less changes from merge conflict resolution to deal with.

To try to prevent this from continuing to get out of hand, I'll try to do the minimal immediate edits from Ryan's feedback and file others in issues.

kenmawer:
It seems that by fixing the conflicts, the tests fail? Did I wrongly use the bottom part rather than the top part of the conflict?

brookslogan:
I'm looking at `git log --remerge-diff` and some minor point did pop up.  I'll push changes for what I notice, then you can double-check.  After that, maybe I'll just file the remaining issues Ryan brought up in Issues and merge so we don't have to keep merging main in here.

brookslogan (approved):

PR #102, opened 2022-06-14
Augment `Abort` & co.; add `rlang` condition tutorial, helpers

brookslogan:
This is a draft.  The additional `Abort`&co. features still need to have `testthat tests`, and to be used throughout the package.

Motivation: we "use" `rlang` conditions, but
- don't use any of its formatting, error class name, and error object field-setting features.
- this breaks the `options(error=recover)` debugging workflow in at least some R versions
- many users may be unaware of some of the features of `rlang` conditions from just following the prompts to `rlang::last_error()` and `rlang::last_trace()`

We want to help users get accustomed to `rlang` conditions, and take advantage of more of its features; building on `Abort` and `Warn` seems like a good way.

Changes so far:
- Add `Inform`, similar to `Abort` and `Warn`
- Add `class_suffix`, `display_subfields`, `more_subfields`, and `call` to
  `Abort`, `Warn`, `Inform`
- On first raised condition, provide a tutorial on `rlang` conditions to help
  users with compatibility issues and potentially-unknown features
- Improve condition traces via `call` default
- Improve condition message line wrapping, compatibility with messages with
  bullet points, etc.

To complement the tutorial, some convenience functions for working with global calling handlers, as well as some handler functions, are provided.


brookslogan:
Roxygen templates provide a different way to execute dynamic code with `<%= ...... %>`.  Might provide a way around the line wrap fighting.

man-roxygen/handler_fn.R
```diff
@@ -0,0 +1,32 @@
+# See notes on the use of `r ...` dynamic R code in roxygen2 comments
+# `utils_conditions.R`. We have to use a bit different code to get this to work
+# inside this template tag file, leading to long `r ...` chunks; don't
+# wrap/fill/format these, or else current roxygen2 at time of writing will not
+# properly process them. Keep the blank lines around them to help prevent
+# accidental auto-formatting.
```
brookslogan, line 6: Roxygen templates provide a different way to execute dynamic code with <%= ...... %>. Might provide a way around the line wrap fighting.  (Here and in other template file, and maybe also in other non-template places where dynamic R code is used that could be turned into templates.)

R/utils_conditions.R
```diff
@@ -0,0 +1,735 @@
...
+#'   `imports_hit_result`, `exports_hit_result`, `namespace_hit_result`}
+#'
+#' @details The tests to detect a package imports and exports environments are
+#'   mostly name-based and thus may be a little imprecise, but also may be more
+#'   robust to any package-related environment restructuring that may happen (or
+#'   may have happened already) in other versions of R.
+#'
+#' @importFrom rlang caller_env is_reference global_env base_env empty_env
+#'   is_function
+#' @importFrom tibble lst
```
brookslogan, line 119: Might need `@noRd` here and in other internal helper functions.  Not the handler ones though, as they will be referenced by the tutorial.

brookslogan:
For migration from existing `Abort` etc., should allow `class` parameter as an alternative to `class_prefix` parameter.

R/utils_conditions.R
brookslogan: vs. `@keywords internal`?

Issue #93, opened 2022-06-05
Improved docs for `as_epi_df()`


dajmcdon:
It would be helpful to add some documentation illustrating how to create an `epi_df` with additional keys.

1. Convert a `tsibble` that has an extra key. Show the meta data
2. Use a data frame that has an additional key and maybe misnamed `geo_value` or `time_value` (or both).
3. The same with something added later. Say, a `covidcast` -> `epi_df` then add a column that adds a key.

Related, are there other types of `additional_metadata` then keys?

This could be done in the examples for `as_epi_df()` or in the main `epiprocess` vignette (or both).

And just as a flag, [note that in the `roxygen`](https://roxygen2.r-lib.org/articles/rd.html#including-external--rmd-md-files) one can use:
```r
#' @includeRmd man/rmd/<some-name>.Rmd details
```

The trailing `details` means to include the Rmd file in the `@details` section (this is the default). You could also put it in any other section (e.g. `@format`).

ChloeYou:
Hi Daniel! @dajmcdon 

I'd like to get some clarifications:

Additional keys are referring to `additonal_metadata` right? Based on the description of `additional_metadata` in the `as_epi_df` function:
`List of additional metadata to attach to the epi_df object. The metadata will have geo_type, time_type, and as_of fields; named entries from the passed list or will be included as well.
`
So the metadata can contain (at least) geo_type, time_type, and as_of fields. Are there other types that are commonly used? 

1. Show the meta data meaning, printing the first couple rows of data in the documentation for the readers to see?
2. What behaviours are we illustrating for the misnamed `geo_value` and `time_value`? Are the misnamed columns additional to an already existing and correctly named `geo_value` and `time_value`?
3. We're pulling some data from `covidcast` and then use `as_epi_df` to turn into an `epi_df` object, then add new columns?

Thank you!

dajmcdon:
So, the metadata is simpler than you're imagining. Here's an example:

```r
attr(jhu_csse_county_level_subset, "metadata")
$geo_type
[1] "county"

$time_type
[1] "day"

$as_of
[1] "2022-05-23 14:35:45 PDT"
```

Here's some example data from BC that sort of illustrates 1 and 2 (extra keys, misnamed columns)

```r
remotes::install_github("mountainmath/CanCovidData")
bc <- CanCovidData::get_british_columbia_case_data()
bc_dated <- bc %>% 
[...]

Issue #89, opened 2022-06-02
Consider validating sensible key values in `epi_archive`, valid ops for & performance improvements from nonunique keys


brookslogan:
Currently, we do not check for distinct key values in `epi_archive`.
- User experience-wise: this is convenient but also gives them the opportunity to form unexpected or invalid archives.
- Semantics: we assume there aren't duplicate key values.  `$as_of` might give the right result if the duplicate keys have duplicate values and we don't want duplicate rows as output, but gives the wrong thing if instead user is trying to take advantage of a particular update-reporting structure which might enable some performance improvements, which we could take further advantage of by being more flexible with the key, described next.
- Performance: we require a version-search for *every key value* across the *single* huge archive DT.
  -  If all (geo_value, time_value, otherkey1, ..., otherkeyn) are re-reported in every `version` --- we are working off of full snapshots in DT --- then we only need to look up the version once, and can key by `version` alone.  But we can't use the `unique` lookup for `as_of` here; maybe a rolling join would work & generalize to the next case.
  - If all (geo_value, otherkey1, ..., otherkeyn) are re-reported in every `version` for time_values `version - 1:windowlength`, then we could key just by (time_value, version).
  - If we have patch-based reporting and no special guarantees, then we need to have (geo_value, otherkey1, ..., otherkeyn, version) as the key, and there should be no duplicates.
  - (Stratifying the DT into multiple DTs, say, one per geo, might increase the number of lookups required but make them faster.)

Might interact with #87.

brookslogan:
Performance part is duplicate of part of #76.

brookslogan:
```r
as_epi_archive(tibble(geo_value=1,time_value=1,version=1,value=c(1,1)))
#> Error in `new_epi_archive()` at epiprocess/R/archive.R:516:3:
#> ! `x` must have one row per unique combination of the key variables. If you have additional key
#>   variables other than `geo_value`, `time_value`, and `version`, such as an age group column, please
#>   specify them in `other_keys`. Otherwise, check for duplicate rows and/or conflicting values for the
#>   same measurement.
```

PR #103, opened 2022-06-14
improve epi df doc

ChloeYou:
Closes https://github.com/cmu-delphi/epiprocess/issues/93.

One example file was created and stored in `man/rmd/epi-df-example.Rmd`. The file was referred to in the main epiprocess vignette and also in the `as_epi_df()` example section.

### DIFF
--- a/R/epi_df.R
+++ b/R/epi_df.R
@@ -111,6 +111,7 @@ NULL
 #' @return An `epi_df` object.
 #'
 #' @export
+#' @includeRmd man/rmd/epi_df_example.Rmd examples
 as_epi_df = function(x, ...) {
   UseMethod("as_epi_df")
 }
--- a/vignettes/epiprocess.Rmd
+++ b/vignettes/epiprocess.Rmd
@@ -198,3 +198,8 @@ ggplot(x, aes(x = time_value, y = cases)) +
   scale_x_date(minor_breaks = "month", date_labels = "%b %y") +
   labs(x = "Date", y = "Confirmed cases of Ebola in Sierra Leone") 
 ```
+
+## Examples on Additional Keys in epi_df
+
+```{r child = 'man/rmd/epi_df_example.Rmd'}
+```
--- a/man/as_epi_df.Rd
+++ b/man/as_epi_df.Rd
@@ -69,3 +69,57 @@ the \code{tbl_ts} class is dropped, and any key variables (other than
 \code{other_keys} field.
 }}
 
+\examples{
+# create an epi_df with additional keys
+
+ ## convert a tsibble that has county code as an extra key
+ ex1 <- tidyr::tibble(
+   geo_value = rep(c("ca", "fl", "pa"), each = 3),
+   county_code = c(06059,06061,06067,
+                   12111,12113,12117,
+                   42101, 42103,42105),
+   time_value = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
+                        by = "day"), length.out = length(geo_value)),
+   value = 1:length(geo_value) + 0.01 * rnorm(length(geo_value)),
+ ) \%>\% 
+   tsibble::as_tsibble(index = time_value, key = c(geo_value, county_code))
+ 
+ ex1 <- as_epi_df(x = ex1, geo_type = "state", time_type = "day", as_of = "2020-06-03")
+ 
+ attr(ex1,"metadata")
+ 
+ ## misnamed columns such as geo_value or time_value
+ ## renaming the columns to match epi_df format
+ ex2 <- tidyr::tibble(
+   state = rep(c("ca", "fl", "pa"), each = 3), # misnamed
+   pol = rep(c("blue", "swing", "swing"), each = 3), # extra key
+   reported_date = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
+                           by = "day"), length.out = length(state)), # misnamed
+   value = 1:length(state) + 0.01 * rnorm(length(state)),
+ ) \%>\% data.frame()
+ 
+ head(ex2) 
+ 
+ ex2 <- ex2 \%>\% rename(geo_value = state, time_value = reported_date) \%>\%
+   as_epi_df(geo_type = "state", as_of = "2020-06-03", 
+             additional_metadata = c(other_keys = "pol"))
+ 
+ attr(ex2,"metadata")
+ 
+ ## add a key to an epi_df object
+ 
+ ### dataset from covidcast
+ ex3 <- jhu_csse_county_level_subset \%>\%
+   filter(time_value > "2021-12-01", state_name == "Massachusetts") \%>\%
+   dplyr::slice_tail(n = 6) 
+ 
+ attr(ex3,"metadata") # geo_type is county currently
+ 
+ ### add state (MA) as a new column and a key to the metadata
+ ex3 <- ex3 \%>\% 
+   tsibble::as_tsibble() \%>\% # needed to add the additional metadata
+   mutate(state = rep("MA",6)) \%>\%
+   as_epi_df(additional_metadata = c(other_keys = "state"))
+ 
+ attr(ex3,"metadata")
+}
--- a/man/as_epi_df.Rd
+++ b/man/as_epi_df.Rd
@@ -69,57 +69,124 @@ the \code{tbl_ts} class is dropped, and any key variables (other than
 \code{other_keys} field.
 }}
 
-\examples{
-# create an epi_df with additional keys
+\section{Convert a \code{tsibble} that has county code as an extra key}{
+\if{html}{\out{<div class="sourceCode r">}}\preformatted{ex1 <- tibble(
+  geo_value = rep(c("ca", "fl", "pa"), each = 3),
+  county_code = c(06059,06061,06067,
+             12111,12113,12117,
+             42101, 42103,42105),
+  time_value = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
+                       by = "day"), length.out = length(geo_value)),
+  value = 1:length(geo_value) + 0.01 * rnorm(length(geo_value))
+  ) \%>\% 
+  as_tsibble(index = time_value, key = c(geo_value, county_code))
 
- ## convert a tsibble that has county code as an extra key
- ex1 <- tidyr::tibble(
-   geo_value = rep(c("ca", "fl", "pa"), each = 3),
-   county_code = c(06059,06061,06067,
-                   12111,12113,12117,
-                   42101, 42103,42105),
-   time_value = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
-                        by = "day"), length.out = length(geo_value)),
-   value = 1:length(geo_value) + 0.01 * rnorm(length(geo_value)),
- ) \%>\% 
-   tsibble::as_tsibble(index = time_value, key = c(geo_value, county_code))
- 
- ex1 <- as_epi_df(x = ex1, geo_type = "state", time_type = "day", as_of = "2020-06-03")
- 
- attr(ex1,"metadata")
- 
- ## misnamed columns such as geo_value or time_value
- ## renaming the columns to match epi_df format
- ex2 <- tidyr::tibble(
-   state = rep(c("ca", "fl", "pa"), each = 3), # misnamed
-   pol = rep(c("blue", "swing", "swing"), each = 3), # extra key
-   reported_date = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
-                           by = "day"), length.out = length(state)), # misnamed
-   value = 1:length(state) + 0.01 * rnorm(length(state)),
- ) \%>\% data.frame()
- 
- head(ex2) 
- 
- ex2 <- ex2 \%>\% rename(geo_value = state, time_value = reported_date) \%>\%
-   as_epi_df(geo_type = "state", as_of = "2020-06-03", 
-             additional_metadata = c(other_keys = "pol"))
- 
- attr(ex2,"metadata")
- 
- ## add a key to an epi_df object
- 
- ### dataset from covidcast
- ex3 <- jhu_csse_county_level_subset \%>\%
-   filter(time_value > "2021-12-01", state_name == "Massachusetts") \%>\%
-   dplyr::slice_tail(n = 6) 
- 
- attr(ex3,"metadata") # geo_type is county currently
- 
- ### add state (MA) as a new column and a key to the metadata
- ex3 <- ex3 \%>\% 
-   tsibble::as_tsibble() \%>\% # needed to add the additional metadata
-   mutate(state = rep("MA",6)) \%>\%
-   as_epi_df(additional_metadata = c(other_keys = "state"))
- 
- attr(ex3,"metadata")
+ex1 <- as_epi_df(x = ex1, geo_type = "state", time_type = "day", as_of = "2020-06-03")
+}\if{html}{\out{</div>}}
+
+The metadata now includes \code{county_code} as an extra key.\if{html}{\out{<div class="sourceCode r">}}\preformatted{attr(ex1,"metadata")
+}\if{html}{\out{</div>}}\preformatted{## $geo_type
+## [1] "state"
+## 
+## $time_type
+## [1] "day"
+## 
+## $as_of
+## [1] "2020-06-03"
+## 
+## $other_keys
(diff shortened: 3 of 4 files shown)


R/epi_df.R
```diff
@@ -111,6 +111,51 @@ NULL
 #' @return An `epi_df` object.
 #'
 #' @export
+#' @examples 
+#' # Convert a `tsibble` that has county code as an extra key
+#' ex1 <- tsibble::tibble(
+#'   geo_value = rep(c("ca", "fl", "pa"), each = 3),
+#'   county_code = c(06059,06061,06067,
+#'                   12111,12113,12117,
+#'                   42101, 42103,42105),
```
brookslogan, line 120: County codes should likely be strings, in order to preserve any leading zeros without extra formatting.  E.g., that would help prevent issues conforming with [the COVID-19 Forecat Hub format](https://github.com/reichlab/covid19-forecast-hub/blob/master/data-processed/README.md#location).

R/epi_df.R
```diff
@@ -111,6 +111,51 @@ NULL
 #' @return An `epi_df` object.
 #'
 #' @export
+#' @examples 
+#' # Convert a `tsibble` that has county code as an extra key
+#' ex1 <- tsibble::tibble(
```
brookslogan, line 116: Here and below, `tsibble::tibble` -> `tibble::tibble`.  Too easy to read `tsibble::` and think this is `tsibble::tsibble`.

R/epi_df.R
```diff
@@ -111,6 +111,51 @@ NULL
...
+#'                   12111,12113,12117,
+#'                   42101, 42103,42105),
+#'   time_value = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
+#'                        by = "day"), length.out = length(geo_value)),
+#'   value = 1:length(geo_value) + 0.01 * rnorm(length(geo_value))
+#' ) %>% 
+#'   tsibble::as_tsibble(index = time_value, key = c(geo_value, county_code))
+#' 
+#' ex1 <- as_epi_df(x = ex1, geo_type = "state", time_type = "day", as_of = "2020-06-03")
+#' attr(ex1,"metadata")
```
brookslogan, line 128: Maybe add a comment here showing the reader how `other_keys` has been set, rather than making them run the example.

R/epi_df.R
```diff
@@ -111,6 +111,51 @@ NULL
...
+#' # Dealing with misspecified column names
+#' ex2 <- tsibble::tibble(
+#'   state = rep(c("ca", "fl", "pa"), each = 3), # misnamed
+#'   pol = rep(c("blue", "swing", "swing"), each = 3), # extra key
+#'   reported_date = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
+#'                           by = "day"), length.out = length(state)), # misnamed
+#'   value = 1:length(state) + 0.01 * rnorm(length(state))
+#' ) %>% data.frame()
+#' 
+#' head(ex2)
```
brookslogan, line 139: Consider renaming `ex2` above to `ex2_input` and something similar for `ex1` and `ex3`, so when the user is running example chunks the output of stuff on `ex{1,2,3}` won't vary based on what chunks they have most recently evaluated or if they accidentally didn't finish evaluating all the way through to the conversion to an `epi_df`.

R/epi_df.R
```diff
@@ -111,6 +111,51 @@ NULL
...
+#' ex2 <- ex2 %>% dplyr::rename(geo_value = state, time_value = reported_date) %>%
+#'   as_epi_df(geo_type = "state", as_of = "2020-06-03", 
+#'             additional_metadata = c(other_keys = "pol"))
+#' 
+#' attr(ex2,"metadata")
+#' 
+#' # Adding additional keys to an `epi_df` object
+#' 
+#' ex3 <- jhu_csse_county_level_subset %>%
+#'   filter(time_value > "2021-12-01", state_name == "Massachusetts") %>%
```
brookslogan, line 150: `filter` -> `dplyr::filter`.  If a user is trying to copy-paste this and convert it to use without attaching `epiprocess` (by using `epiprocess::` everywhere), this is probably going to break because it will be using `stats::filter` rather `epiprocess::filter` / `dplyr::filter`.

R/epi_df.R
```diff
@@ -111,6 +111,51 @@ NULL
...
+#'   dplyr::slice_tail(n = 6) 
+#'
```
brookslogan, line 152: --- TODO: pick up reviewing from here ---

R/epi_df.R
```diff
@@ -111,6 +111,51 @@ NULL
...
+#' attr(ex1,"metadata")
+#' 
+#' # Dealing with misspecified column names
+#' ex2 <- tsibble::tibble(
+#'   state = rep(c("ca", "fl", "pa"), each = 3), # misnamed
+#'   pol = rep(c("blue", "swing", "swing"), each = 3), # extra key
+#'   reported_date = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
+#'                           by = "day"), length.out = length(state)), # misnamed
+#'   value = 1:length(state) + 0.01 * rnorm(length(state))
+#' ) %>% data.frame()
```
brookslogan, line 137: Is it important that this is a non-`tibble`?  Otherwise, maybe leave it as a tibble, `head(ex2)` -> `ex2`, the user will also get some quick type info if just running the example, and avoids inconsistencies between R <4 and >=4 and between user option settings regarding `stringsAsFactors`.

R/epi_df.R
```diff
@@ -111,6 +111,51 @@ NULL
...
+#'   time_value = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
+#'                        by = "day"), length.out = length(geo_value)),
+#'   value = 1:length(geo_value) + 0.01 * rnorm(length(geo_value))
+#' ) %>% 
+#'   tsibble::as_tsibble(index = time_value, key = c(geo_value, county_code))
+#' 
+#' ex1 <- as_epi_df(x = ex1, geo_type = "state", time_type = "day", as_of = "2020-06-03")
+#' attr(ex1,"metadata")
+#' 
+#' # Dealing with misspecified column names
```
brookslogan, line 130: Maybe explain a bit more?  E.g., something like
```
#' # Geographical and temporal information must be provided in columns named
#' # `geo_value` and `time_value`; if we start from a data frame with a
#' # different format, it must be converted to use `geo_value` and `time_value`
#' # before calling `as_epi_df`
```
?

vignettes/epiprocess.Rmd
```diff
@@ -198,3 +198,89 @@ ggplot(x, aes(x = time_value, y = cases)) +
...
+data.frame(
+  state = rep(c("ca", "fl", "pa"), each = 3), # misnamed
+  pol = rep(c("blue", "swing", "swing"), each = 3), # extra key
+  reported_date = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
+                       by = "day"), length.out = length(geo_value)), # misnamed
+  value = 1:length(geo_value) + 0.01 * rnorm(length(geo_value))
+) %>% as_epi_df() 
+```
+
+The columns can be renamed to match `epi_df` format. In the example below, notice there is also an additional key `pol`.
```
brookslogan, line 240: This line of text and the one above seem great for explaining the corresponding `@examples` example above.

Many of the other comments / requested changes above also apply to this vignette.

I wonder if there's a way to reduce duplication here via `@example` files or `system.file` + some sort of magic on the vignette side?  We file that away in a separate issue, though.

vignettes/epiprocess.Rmd
```diff
@@ -198,3 +198,89 @@ ggplot(x, aes(x = time_value, y = cases)) +
   scale_x_date(minor_breaks = "month", date_labels = "%b %y") +
   labs(x = "Date", y = "Confirmed cases of Ebola in Sierra Leone") 
 ```
+
+## Examples on Additional Keys in epi_df
```
brookslogan, line 202: -> "## Using additional key columns in `epi_df`"?

And maybe move this to come immediately after the metadata section, since it's also metadata?

vignettes/epiprocess.Rmd
```diff
@@ -198,3 +198,89 @@ ggplot(x, aes(x = time_value, y = cases)) +
...
+In the following examples we will show how to create an `epi_df` with additional keys.
+
+### Convert a `tsibble` that has county code as an extra key
```
brookslogan, line 205: "Convert" -> "Converting"

vignettes/epiprocess.Rmd
```diff
@@ -198,3 +198,89 @@ ggplot(x, aes(x = time_value, y = cases)) +
...
+
+### Adding additional keys to an `epi_df` object
+
+In the above examples, all the keys are added to objects that are not `epi_df` objects. We illustrate how to add keys to an `epi_df` object. 
+
+We use a subset dataset from the the `covidcast` library. 
+
+```{r}
+ex3 <- jhu_csse_county_level_subset %>%
+  filter(time_value > "2021-12-01", state_name == "Massachusetts") %>%
```
brookslogan, line 268: Explicitly say we are filtering to a single state for simplicity here, but it'd work if we actually were working with multi-state data, or make it a multi-state example (e.g., `left_join`ing with some state name information, or something like `dplyr::mutate(state=datasets::state.abb[match(.data$state_name, datasets::state.name)])`.)  Maybe also use `tolower` / "ma" because I think that's what we would expect if this were a `geo_value` column --- just to make readers get used to seeing&using lower case.

vignettes/epiprocess.Rmd
```diff
@@ -198,3 +198,89 @@ ggplot(x, aes(x = time_value, y = cases)) +
...
+  slice_tail(n = 6) 
+  
+attr(ex3,"metadata") # geo_type is county currently
+```
+
+Now we add state (MA) as a new column and a key to the metadata. 
+```{r}
+
+ex3 <- ex3 %>% 
+  as_tsibble() %>% # needed to add the additional metadata
```
brookslogan, line 278: `as_tibble()` would be a lighter-weight (package-wise and cognitively, for those who are not tsibble experts) dependency.

Also, it's a bug that if we exclude this line, we silently ignore the `additional_metadata` arg.  This could be filed in another issue.

vignettes/epiprocess.Rmd
```diff
@@ -198,3 +198,89 @@ ggplot(x, aes(x = time_value, y = cases)) +
...
+  mutate(state = rep("MA",6)) %>%
+  as_epi_df(additional_metadata = c(other_keys = "state"))
+
+attr(ex3,"metadata")
+```
+
+
+
```
brookslogan, line 286: If it's not already done, explain how setting these other keys is important / changes anything?  (E.g., default `epi_slide` behavior.)

vignettes/epiprocess.Rmd
```diff
@@ -198,3 +198,89 @@ ggplot(x, aes(x = time_value, y = cases)) +
...
+### Convert a `tsibble` that has county code as an extra key
+```{r}
+ex1 <- tibble(
+  geo_value = rep(c("ca", "fl", "pa"), each = 3),
+  county_code = c(06059,06061,06067,
+             12111,12113,12117,
+             42101, 42103,42105),
+  time_value = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
+                       by = "day"), length.out = length(geo_value)),
+  value = 1:length(geo_value) + 0.01 * rnorm(length(geo_value))
```
brookslogan, line 214: Remove the randomness or use `withr::with_rng_version` and `withr::with_seed` so that we don't generate git diffs whenever rebuilding vignettes.

brookslogan (changes requested):
Nice examples!  We may want to borrow these when constructing any tests for behavior regarding `other_keys`.  I have a some minor comments/requests + we should check on / copy over the appropriate data attribution to this vignette + a couple things that should be filed as separate issues.

R/epi_df.R
brookslogan: (I think there was some vignette text that might work well here.)

vignettes/epiprocess.Rmd
ChloeYou: I can create a issue for this. Thank you!

ChloeYou: https://github.com/cmu-delphi/epiprocess/issues/133

vignettes/epiprocess.Rmd
ChloeYou: Hi @brookslogan! 

I'm actually unsure about how the other keys will affect `epi_slide()`? Would you be able to elaborate a bit more on this?

For example, would we expect the default bahaviour of `epi_slide` to group by geo_value and the other keys? Currently, it doesn't seem to be doing that. 

``` r
suppressPackageStartupMessages(library(epiprocess))
suppressPackageStartupMessages(library(dplyr))
suppressPackageStartupMessages(library(tsibble))

ex1 <- tibble(
  geo_value = rep(c("ca", "fl", "pa"), each = 3),
  county_code = c("06059","06061","06067",
                  "12111","12113","12117",
                  "42101","42103","42105"),
  time_value = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
                       by = "day"), length.out = length(geo_value)),
  value = 1:length(geo_value) + 0.01 * withr::with_rng_version("3.0.0", withr::with_seed(42, length(geo_value)))
) %>% 
  as_tsibble(index = time_value, key = c(geo_value, county_code))

ex1 <- as_epi_df(x = ex1, geo_type = "state", time_type = "day", as_of = "2020-06-03")

# do we expect the following two lines to be the same?
ex1  %>% group_by(geo_value, county_code) %>% epi_slide(avg = mean(value), n = 2)
#> An `epi_df` object, with metadata:
#> * geo_type  = state
#> * time_type = day
#> * as_of     = 2020-06-03
#> 
#> # A tibble: 9 x 5
#> # Groups:   geo_value, county_code [9]
#>   geo_value county_code time_value value   avg
#> * <chr>     <chr>       <date>     <dbl> <dbl>
#> 1 ca        06059       2020-06-01  1.09  1.09
#> 2 ca        06061       2020-06-02  2.09  2.09
#> 3 ca        06067       2020-06-03  3.09  3.09
#> 4 fl        12111       2020-06-01  4.09  4.09
#> 5 fl        12113       2020-06-02  5.09  5.09
#> 6 fl        12117       2020-06-03  6.09  6.09
#> 7 pa        42101       2020-06-01  7.09  7.09
#> 8 pa        42103       2020-06-02  8.09  8.09
#> 9 pa        42105       2020-06-03  9.09  9.09
ex1  %>% epi_slide(avg = mean(value), n = 2)
#> An `epi_df` object, with metadata:
#> * geo_type  = state
#> * time_type = day
#> * as_of     = 2020-06-03
#> 
#> # A tibble: 9 x 5
#>   geo_value time_value county_code value   avg
#> * <chr>     <date>     <chr>       <dbl> <dbl>
#> 1 ca        2020-06-01 06059        1.09  4.09
#> 2 fl        2020-06-01 12111        4.09  4.09
#> 3 pa        2020-06-01 42101        7.09  4.09
#> 4 ca        2020-06-02 06061        2.09  4.59
#> 5 fl        2020-06-02 12113        5.09  4.59
#> 6 pa        2020-06-02 42103        8.09  4.59
#> 7 ca        2020-06-03 06067        3.09  5.59
#> 8 fl        2020-06-03 12117        6.09  5.59
#> 9 pa        2020-06-03 42105        9.09  5.59
```
<sup>Created on 2022-07-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

brookslogan: I wasn't sure either.  Looking over, I also don't see it impacting `epi_slide`.  And from a quick skim, it looks like the `"other_keys"` metadata in `epi_df` does not impact anything and is purely informational at this point, unlike in `as_epi_archive`, where the `other_keys` parameter affects how the update data is interpreted.  This mismatch may be surprising, so (a) we should document it, even if it's saying "we don't use this in one case", and (b) we may want to revisit it (see https://github.com/cmu-delphi/epiprocess/issues/89).

ChloeYou: Sure thing, thanks Logan. I'll add a one-line comment in the vignette to point out that currently `other_keys` metadata in `epi_df` doesn't impact `epi_slide`, contrary to `other_keys` in `as_epi_archive` which affects how the update data is interpreted.

ChloeYou:
@brookslogan I think I've addressed all the comments. Let me know if anything else should be changed, otherwise we can probably merge this? Thanks!

Issue #94, opened 2022-06-05
Create an `epi_df` constructor


dajmcdon:
it would be useful to have a constructor for an `epi_df` (possibly an `epi_archive` as well). It should probably just build an empty tibble with the correct metadata. Probably with some default `geo_type` and `time_type` and `as_of`. 

This would be helpful in some downstream tasks and likely also in tests.

Name should be `new_epi_df()`.

PR #104, opened 2022-06-16
Created an `epi_df` constructor

rachlobay:
As per [issue 94](https://github.com/cmu-delphi/epiprocess/issues/94), created a constructor for `epi_df` called `new_epi_df()`, which was added to the epi_df R script. By default, it builds an empty tibble with the correct metadata (based on `as_epi_df` function). Note that this was basically shifting some the code from `as_epi_df()` to this new constructor function... I am not sure if that is ok as now the two functions have a lot of overlap, so let me know what you think. Also, there are some tests for `new_epi_df()` in test-epi_df.R.

Regarding `epi_archive`, `epi_archive$new` is the public method that creates a new `epi_archive` object as $new is the constructor for R6 objects. So, I don't think that there is much to be done there, but maybe there is a modification you want me to make? In that case, let me know what that is on a separate issue.


dajmcdon (approved):
Looks good to me.

Issue #96, opened 2022-06-05
`head()` and `tail()` drop the class


dajmcdon:
``` r
library(epiprocess)
#> 
#> Attaching package: 'epiprocess'
#> The following object is masked from 'package:stats':
#> 
#>     filter

head(jhu_csse_daily_subset)
#> # A tibble: 6 × 6
#>   geo_value time_value case_rate_7d_av death_rate_7d_av cases cases_7d_av
#>   <chr>     <date>               <dbl>            <dbl> <dbl>       <dbl>
#> 1 ca        2020-03-01         0.00327         0            6        1.29
#> 2 ca        2020-03-02         0.00435         0            4        1.71
#> 3 ca        2020-03-03         0.00617         0            6        2.43
#> 4 ca        2020-03-04         0.00980         0.000363    11        3.86
#> 5 ca        2020-03-05         0.0134          0.000363    10        5.29
#> 6 ca        2020-03-06         0.0200          0.000363    18        7.86
tail(jhu_csse_daily_subset)
#> # A tibble: 6 × 6
#>   geo_value time_value case_rate_7d_av death_rate_7d_av cases cases_7d_av
#>   <chr>     <date>               <dbl>            <dbl> <dbl>       <dbl>
#> 1 tx        2021-12-26            41.5            0.119 12559      12177.
#> 2 tx        2021-12-27            44.7            0.188 17331      13137 
#> 3 tx        2021-12-28            46.9            0.197 14795      13770.
#> 4 tx        2021-12-29            48.9            0.195 14286      14365.
#> 5 tx        2021-12-30            50.4            0.205 15600      14811.
#> 6 tx        2021-12-31            51.4            0.205 15762      15096.
```

<sup>Created on 2022-06-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

These methods both convert to `tibble` then dispatch to the `data.frame` methods. They should either use `epiprocess:::reclass()` afterword, or, likely better, use `[.epi_df`. 

Possibly see [here](https://adv-r.hadley.nz/s3.html#s3-inheritance)

PR #105, opened 2022-06-16
Keep the `epi_df` class when head and tail are used

rachlobay:
As per [issue 96](https://github.com/cmu-delphi/epiprocess/issues/96) `head.epi_df` and `tail.epi_df` to keep the `epi_df` class when they are used. Note the tests for this in test-methods-epi_df.R


tests/testthat/test-methods-epi_df.R
```diff
@@ -0,0 +1,21 @@
...
+      length.out = 5
+    ), times = 2),
+    geo_value = rep(c("ca", "hi"), each = 5)
+  ) %>% epiprocess::as_epi_df()
+  
+  expect_true(is_epi_df(head(tib)))
+  expect_true(is_epi_df(tail(tib)))
+  expect_identical(attributes(head(tib))$metadata$geo_type, attributes(tib)$metadata$geo_type)
+  expect_identical(attributes(head(tib))$metadata$time_type, attributes(tib)$metadata$time_type)
+  expect_identical(attributes(head(tib))$metadata$as_of, attributes(tib)$metadata$as_of)
```
brookslogan, line 17: Is `additional_metadata` also supposed to be preserved?  The tests for head and tail here should probably ensure that the desired behavior (retaining, dropping, etc.) is achieved.

R/methods-epi_df.R
```diff
@@ -67,15 +67,15 @@ summary.epi_df = function(object, ...) {
 #' @export
 #' @noRd
 head.epi_df = function(x, ...) {
-  head(tibble::as_tibble(x), ...)
+  as_epi_df(NextMethod("head"))
```
brookslogan, line 70: Minor "style"-like thing: might want to exclude `"head"` if it's not necessary to match how `NextMethod` is called elsewhere in the file, or explain why it's necessary in a comment if it is necessary.

brookslogan:
The comments above are very minor.  This PR could be merged to address #96 without any changes.

tests/testthat/test-methods-epi_df.R
```diff
@@ -0,0 +1,21 @@
+test_that("head and tail do not drop the epi_df class", {
+  tib <- tibble::tibble(
```
brookslogan, line 2: `tib` sounds like it should be a tibble, but this is an `epi_df`; maybe rename to something else.

rachlobay: Done.

R/methods-epi_df.R
rachlobay: Added comment as to why "head" was specified in NextMethod() in file (essentially, the first arg. should be the name of the generic fun. that want to offload the work to).

tests/testthat/test-methods-epi_df.R
rachlobay: Sure I could add a column for that... I haven't seen it done elsewhere (ex. in the other tests). So, let me know what a reasonable example of `additional_metadata` would be and I can put that in.

dajmcdon:
@rachlobay I spent a bit of time looking into this, and I think we're doing this the wrong way (my fault). `head()` actually calls `utils:::head.matrix()` eventually, which uses `[.tbl_df`. So rather than implementing `head.epi_df()` we should actually be implementing `[.epi_df`. Those changes will then propagate to a call to `head()` or `tail()` as well as keeping the class if someone does `some_epi_df[1:5, ]`.

I think we want it to *drop* the class and the metadata if they do column selection: `some_epi_df[ ,1:2]` should just go to a `tbl_df`.

Good examples are in [tibble](https://github.com/tidyverse/tibble/blob/main/R/subsetting.R) or probably even better [tsibble](https://github.com/tidyverts/tsibble/blob/main/R/subset.R). I suspect that we essentially need most of the code in the tSibble link.

dajmcdon (changes requested):
I think there are a few additional tests that we need here, otherwise, the rest is linting.

R/methods-epi_df.R
```diff
@@ -62,20 +62,56 @@ summary.epi_df = function(object, ...) {
...
+  not_epi_df <- !("time_value" %in% cn) || !("geo_value" %in% cn) || vctrs::vec_size(res) > nr || any(i > nr)
+  if (not_epi_df) return(tibble::as_tibble(res))
+  
+  # Case when i is numeric and there are duplicate values in it
+  if (is.numeric(i) && vctrs::vec_duplicate_any(i) > 0) return(tibble::as_tibble(res))
+  
+  # Column subsetting only, then return res as tibble
+  if (rlang::is_null(i) && !rlang::is_null(j)) return(tibble::as_tibble(res))
+  
+  att_x = attributes(x)
```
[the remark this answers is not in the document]
rachlobay: Done.

tests/testthat/test-methods-epi_df.R
```diff
@@ -0,0 +1,21 @@
...
+    x = 1:10,
+    y = 1:10,
+    time_value = rep(seq(
+      as.Date("2020-01-01"),
+      by = 1,
+      length.out = 5
+    ), times = 2),
+    geo_value = rep(c("ca", "hi"), each = 5)
+  ) %>% epiprocess::as_epi_df()
+
```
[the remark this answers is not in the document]
rachlobay: Added various tests

dajmcdon (approved):

brookslogan:
Some notes:
- Above test failures are related to the `additional_metadata`; at least some are due to a missing `$additional_metadata` in the tests.
- (Optional) The `additional_metadata` can be set directly in `as_epi_df`; see `pkgload::dev_help("as_epi_df")`.
- Either `additional_key_name` and `"additional_key_type"` should be changed to `other_keys` and `"<some column that acts as part of the key in addition to geo and time values>"` or turned into something without "key" in the name.  E.g., `source_url` -> `"https://example.com/data.csv"` or something like that.

tests/testthat/test-methods-epi_df.R
brookslogan: See `other_keys` / `source_url` example below.  Marking this resolved so just that comment remains to look at.

Issue #62, opened 2022-03-29
Add option to compactify issue data in `epi_archive`


brookslogan:
`epi_archive`s can be formed based on a conglomeration of full snapshots, issue data with duplicate re-reporting, and/or minimal patch-like issues.  Some space (unsure about time) can be saved by removing rows that match LOCF of previous issues.  Space can be essential if we are attempting in-memory analysis.

Proposal: introduce a constructor argument `compactify`:
- `TRUE`: remove unnecessary rows to give same LOCF results.  Make sure to maintain the same `max_issue` value as the original data.
- `FALSE`: leave data as-is
- default value (say, `NULL`): same as `TRUE` except message the user if this actually changed the data, and telling them how to silence the message

Use cases:
- User inputs full snapshot data, to prevent using space quadratic in the number of snapshots.  (A further enhancement would be to directly work off of a directory of snapshot files or something similar.)
- We are working off of a covidcast data source that historically did not use diff-based issues and/or has many full re-issues.  (E.g., repeating the analysis [here](https://github.com/cmu-delphi/covidcast-indicators/issues/1362) gives covidcast jhu-csse state-level case issue data at 79% duplicates despite the shift to having routine issues being diff-based.  This is still just reducing what `object.size` says is 40MB--50MB down to ~10MB, but at the county level it _might_ matter more.)

ryantibs:
@lcbrooks Thanks for the idea! I think it sounds like a good idea. 

Not sure if `compactify` would be the name I would have come up with. Is that a somewhat commonly-used verb? Or something you made up?

brookslogan:
Regarding the naming: I just didn't want to call it `compress` (this is not applying a general-purpose compression algorithm but something archive-specific) and couldn't think of anything better.  Not sure where I picked up the word, but it's in some dictionaries and appears to be used in math and physics theory in a limited number of contexts.

kenmawer:
I can help with this.

brookslogan:
Is this in a PR already?  If so, could you please point me to it; otherwise, could you open a PR?

PR #107, opened 2022-06-17
km-compactify_rectify2

kenmawer:
Note that the first row may be erroneously removed if all fields (except `version`) are NA's. However, the handling of NA's is beyond the scope of this branch and rectification of compactify.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -51,7 +51,6 @@ Config/testthat/edition: 3
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.1.2
+RoxygenNote: 7.2.0
 Depends: 
     R (>= 2.10)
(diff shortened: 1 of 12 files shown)


brookslogan:
Testing how project board automation reacts: this PR addresses #62.

brookslogan:
Resolves #62.

brookslogan:
Not sure if this should be a separate PR or commits pushed onto #101.  The former does might allow for a bit easier use of "/compare" GitHub endpoints to compare associated branches since the branches names are at the top of the commit (although this can still be accomplished in the latter setup by copying the desired commit hashes from the PR conversation.  The latter probably allows for easier tracking of which requested changes/comments are being addressed due to the "out of date" mechanism (but harder to see what they were originally about).  For now, I'm going to go ahead and try to push these changes back onto #101's branch and close this one.

Issue #109, opened 2022-06-18
Allow or help handle `epix_as_of` last recorded version even when this version has no update


brookslogan:
When preparing prospective production forecasts, we may want to be able to mirror `epix_slide` or `epix_as_of` operations that we use to prepare pseudoprospective analyses.  However, this may involve getting a snapshot `epix_as_of` a version with no update data (e.g., due to a nondaily update cadence, a holiday or some other occurrence causing the data source to be stale, or running the forecasts earlier than the data source has been updated and not being able to distinguish this from the data source responding that there has been no change).  Currently, unless we have some "redundant" `DT` rows duplicating previous values with the latest no-change version, this is going to raise an error (`max_version` > `self_max`).  (This also means that compactification (#101) could change the error behavior here.)  We should either
- take another arg to the archive constructor that acts as an override for `self_max`, and would allow this operation to go through with a warning rather than an error, or
- provide some helper functions, parameters, and messages to deal with this situation: if we are dealing with a stale archive (`max(DT$version)` < `forecast_ref_time`), do we interpret this as if the data source reported no change (unless it looks really stale) and allow getting the snapshot, or do we interpret it as an issue fetching the data and, up to some point, allow ourselves to back up to `epix_as_of` the max DT version, providing this as a parameter to the slide function to adjust things?  These are actually different options when you think about fetching backfill-aware training sets and forecast aheads relative to the snapshot version.

brookslogan:
To make this more concrete, using HHS COVID-19 hospitalization data; suppose every Monday we prepare a forecast of COVID-19 admissions based on recent COVID-19 admissions.
```



library(tidyverse)
library(data.table)
library(delphi.epidata)

covid.admissions.tbl = delphi.epidata::covidcast(
  data_source = "hhs",
  signals = "confirmed_admissions_covid_1d",
  time_type = "day",
  geo_type = "state",
  time_values = epirange(12340601, 20221201),
  geo_values = "ca,fl",
  issues = epirange(12340601, 20221201)
) %>% delphi.epidata::fetch_tbl()

some.dates.any.wday = seq(min(covid.admissions.tbl$issue), max(covid.admissions.tbl$issue), by="day")
some.mondays = some.dates.any.wday[as.POSIXlt(some.dates.any.wday)$wday==1L]

( # (parens for prettier data.table chain)
  data.table(issue=unique(covid.admissions.tbl$issue), key="issue")
[...]

Issue #106, opened 2022-06-16
Improving printing of epi_archives with `NA`'s


kenmawer:
When calling an `epi_archive`, if there's an `NA` value in the `time_value` or `version`, the information given about the minimal and maximal values shows up as an `NA` rather than ignoring those values. The `NA`'s should be treated as though they did not even exist, as these values are not applicable regardless.

brookslogan:
We probably shouldn't be able to construct an `epi_archive` containing any `NA` `version` to begin with, or it should have a special, documented, tested meaning.  Currently, if `NA` versions may cause some unintended behavior for LOCF operations.  As for `time_value` (and `geo_value`), I guess `NA`s might be all right... maybe they want to use the versioning mechanism for something constant in time, and this is a workaround the geo&time value requirement to get a more general version-tracking mechanism.

brookslogan:
NA versions are being disallowed in another PR, so I think it's just the time value printing to implement here.

brookslogan:
The PR I thought handled the version validation and printing part of this doesn't seem to have it.  Reopening as a reminder to check on this.

Issue #90, opened 2022-06-02
Don't filter by `time_value` in `as_of`


brookslogan:
The current `as_of` includes a time filter relative to the version value.  This prevents us from using `as_of` when storing forecasts in an `epi_archive`.

brookslogan:
Fixing this will likely break arx_forecaster due to a bug there.  I'm writing up some comments and then will open an issue on that other bug.  The other bug will likely need to be resolved first and will block this one.

rachlobay:
@brookslogan To make sure that I understand, is the section that you are referring to lines 216-218  of archive.R?

Issue #49, opened 2022-02-24
Should `slide()` for `epi_archive` be given access to less than the most up-to-date snapshots?


ryantibs:
Currently the `slide()` method for the `epi_archive` class only uses the most up-to-date values of each signal as of the reference time for the sliding computation.  To do this, it calls the `as_of()` method for the `epi_archive` class.

Some computations, e.g., backcasting or nowcasting, will want to access more than just the most up-to-date signal values. Should we extend `slide()` and `as_of()` to accommodate this? We could do it with an extra argument `all_versions = TRUE`. The default should be `FALSE` so the methods continue to behave as-is.

Or should these be separate functions entirely and keep `slide()` and `as_of()` so that they just consider/deliver the most recent snapshots?

@lcbrooks @elray1 @dajmcdon @jacobbien Do you guys have opinions?

brookslogan:
My thoughts below.  It doesn't look like there's not a dominant approach.  I guess I'd lean toward separate functions.

Pro for using parameter: doesn’t expand namespace

Pro for using separate functions: keeps the types involved a little more consistent:
- as_of : fn (epi_archive, version_type) -> epi_df
- archive_as_of : fn (epi_archive, version_type) -> epi_archive
- slide: fn (epi_archive, formula/tidymutation of epi_df) -> epi_df
- slide_by_archive: fn (epi_archive, formula/tidymutation of epi_archive) -> epi_df
- slide_by_archive_different_semantics:  fn (epi_archive, fn (epi_archive) -> epi_df) -> epi_archive
Rather than expecting/producing different types based on the value of a string argument. 

The interface is the most important, but code-wise we might also think that the parameter approach might provide more immediate input or avoid repeated validation of the same inputs, while the separate-functions approach would have simpler control flow (no branching on the string argument; instead, we’d probably have as_of call archive_as_of internally, and slide call slide_by_archive internally).

ryantibs:
Thanks! Reading what you wrote, and thinking about it myself, I'm actually think about implementing it as a parameter `all_versions`. I think that would be very simple to do.

@dajmcdon @jacobbien @elray1 Opinions?

brookslogan:
The `all_versions` approach already seems to have been implemented.  Closing.

ryantibs:
Just curious, which branch/PR?

brookslogan:
Sorry, good catch.  Got mixed up with `all_rows`.

nmdefries:
[...]

PR #114, opened 2022-06-21
Port Jingjing's backcasting preprocessing utils

### LINKED ISSUE #88: Add functions that will enable backcasters to be built on top
Jingjing has already prepared these outside of epiprocess.  They will need to be copied over and integrated with `epi_archive`.  Depends on or interacts with #49.

brookslogan:
Resolves #88.  Might involve #49, #90, #106, #109.

Progress so far:
- Ported over some of Jingjing's preprocessing functions and tests.
- Got the tests running in a package environment.

Some TODOs:
- [ ] Get these working based on `epiprocess` classes (taking care that current functions work on a single geo/epigroup at a time)
- [x] Work either on implied lag `version - time_value` or an explicit lag/lag-like column.
- [ ] ~Check that, or change to: either always, or optionally, add NAs getting/filling lag training data if it looks like we missed recording a version (it's okay if there is no update for an observation in a target version, but a problem if there are no updates for any observations in a target version)~ unless there are bugs from assuming that versions are evenly spaced, this is a separate convenience function/arg to think about independently
- [ ] Change `fill_rows` and `fill_missing_updates` to use a mix of last-version-carried-forward and NA/0/customizable fill-in determined by archive's [check out $fill_through_version --- may or may not be useful]
- [ ] Add examples
- [x] Use `Abort`, etc., rather than `stop`
- [ ] Think about interaction with `epix_slide`.  Straightforward is probably combining with https://github.com/cmu-delphi/epiprocess/issues/49.  But we might also think about turning the time&version-lag covariates dfs into a custom type of epi_archive (base epi_archive would be pretty inefficient though; lazy merge might be one way to improve). ------ Just do the straightforward way to begin with, then profile.
- [ ] [Use alternative to zoo::rollmeanr.  Maybe data.table::frollmean?]
- [ ] [Check compatibility of "shift" terminology with tidymodels, which is used by epipredict.  (Might be incompatible with some of our production COVID-19 hospitalization forecasters, but that might be fine.)]


nmdefries:
Closing because this PR is stale. Underlying code on both `epiprocess` and backcasting sides have changed so it would be easier to redo this if we're still interested in the future. Most of the related issues have been addressed elsewhere; only https://github.com/cmu-delphi/epiprocess/issues/109 and https://github.com/cmu-delphi/epiprocess/issues/106 are still open.

PR #118, opened 2022-06-22
Created a separate vignette to describe `compactify`, as well as its indirect ability to save time

kenmawer:
The vignette describes `compactify`, taken out from the archive vignette originally. It also describes how the use of `compactify` can save time for operations such as `as_of`. Note that the runtimes are not hard-coded, meaning that the time it takes may differ from one run of the code to another.

Note that once complete, the `archive` vignette should have a link to `compactify` in it so a user can find the information they need by clicking "See more" or something of the like.


brookslogan:
Were all the changes that were in this PR incorporated into #101?

dajmcdon:
I think that's right. @ken and I went through these together to decide which to keep.

PR #130, opened 2022-06-28
Updated `epi_slide` examples to ensure new col is printed

rachlobay:
As per our discussion @brookslogan, updated the `epi_slide` examples to remove a nonessential var in the output so that the new col is ensured to be printed when the usual setting of `getOption("width")` of 80 is used in R or RStudio. Should be good to go now.

### DIFF
--- a/R/slide.R
+++ b/R/slide.R
@@ -95,7 +95,7 @@
 #'   epi_slide(cases_7dav = mean(cases), n = 7, 
 #'             align = "right") %>% 
 #'   # rmv a nonessential var. to ensure new col is printed
-#'   dplyr::select(-c(death_rate_7d_av))
+#'   dplyr::select(-death_rate_7d_av) 
 #'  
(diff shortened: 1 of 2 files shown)


R/slide.R
```diff
@@ -93,13 +93,17 @@
 #'   jhu_csse_daily_subset %>%
 #'   group_by(geo_value) %>%
 #'   epi_slide(cases_7dav = mean(cases), n = 7, 
-#'             align = "right")
+#'             align = "right") %>% 
+#'   # rmv a nonessential var. to ensure new col is printed
+#'   dplyr::select(-c(death_rate_7d_av))
```
dajmcdon, line 98: Here and below, you can just use:

```r
dplyr::select(-death_rate_7d_av) # no need to wrap with c()
```

dajmcdon (approved):
Just a minor change, feel free to merge when done.

Issue #126, opened 2022-06-28
`guess_time_type`'s "week" argument should be 7, not -7 on line 82


kenmawer:
This is a very simple bug to fix.

PR #131, opened 2022-06-28
Updated and fixed utils, included adding tests

### LINKED ISSUE #123: guess_geo_type not checking strictly enough for `county` and `nation`
I'm writing tests for `utils` and I noticed that strings comprised of numbers more than 5 numbers long are grouped as `"county"`, and strings comprised of numbers more than 2 letters long are grouped as `"nation"`, when both should be grouped as `"custom"`.

For example:

`  guess_geo_type(c("Vancouver Coastal","Interior","Fraser", "Northern","Vancouver Island"))` gives `"nation"` even though these are British Columbia health regions.

kenmawer:
This incorporates changes from `km-test-utils` as well as those from `km-fix-utils`. The former added tests to utils, while the latter fixed buggy function on utils while commenting out tests that had initially failed on test-utils. I incorporated two different branches into one as to avoid conflicts should these had been done individually.

This resolves #123 and #126.


dajmcdon (approved):
Looks good to me.

Issue #98, opened 2022-06-07
Improve documentation experience for users constructing first epi_archive


brookslogan:
- [x] `epi_archive$new` doesn't provide hints about relevant parameter names or autocompletion of those names.  Point users to `as_epi_archive` in the intro and in the `$initialize`/`$new` documentation.
- [x] add an example in `epi_archive` `@examples` of constructing an epi archive from a small amount of fake data

PR #134, opened 2022-07-05
Improved documentation experience for users constructing first `epi_archive`

rachlobay:
Improved documentation experience for users constructing first `epi_archive` by addressing two issues noted in [Issue 98](https://github.com/cmu-delphi/epiprocess/issues/98)
1. Pointed users to `as_epi_archive` in the intro and in the `$initialize`/`$new` documentation. 
2. Added an example in `epi_archive` `@examples` of constructing an epi archive from a small amount of fake data


R/archive.R
```diff
@@ -104,6 +119,10 @@ epi_archive =
 #'   fields; named entries from the passed list or will be included as well.
 #' @return An `epi_archive` object.
 #' @importFrom data.table as.data.table key setkeyv
+#' 
+#' @details 
+#' Refer to the documentation for `as_epi_archive()` for more information
```
ChloeYou, line 124: Minor improvement but I found online that we can actually link the function help page in roxygen using `\code{\link{as_epi_archive}}`. Might be helpful here?

https://stackoverflow.com/questions/25489042/linking-to-other-packages-in-documentation-in-roxygen2-in-r

rachlobay: Thanks, Chloe! Based on a quick look into this, I think that one of the easiest ways to link to another function in the package we're working on is to use something like `[as_epi_archive()]` in the code, which translates to `\code{\link[=as_epi_archive]{as_epi_archive()}}` in the Rd file. I've now done that where I stated to look to the `as_epi_archive()` function in the `epi_archive` documentation.

dajmcdon (approved):

Issue #143, opened 2022-07-13
Clean up `epi_slide` `f` parameter documentation 


rachlobay:
As discussed with Logan, update the `@param` documentation for the `f` parameter in `epi_slide`. The following are the points to address:

- Instead of “If a function, `f` must take `x`, a data frame with the same column names as the original object; followed by any number of named arguments; and ending with `...`“, the proper thing may be that `f` should take `x`, `g`/`k`/`key`/`think-of-a-better-name`, and named args (dots don’t seem necessary; I think originally they were just there to absorb the unlisted `g`).
- Correct "same column names as the original object" --- we might only get the non-grouping columns here.  [This raises a separate issue.  We are feeding the user an `epi_df` via `f` that might not have the geo_value column as one might expect.  Hopefully we can get away with it, but maybe this needs to be a proper `epi_df` or just a `tibble` instead.]
- If you think it clarifies things, change “named arguments” -> “named arguments (which will be taken from `...`)“.
- Make sure that the description of `g` above matches/resembles how `?group_modify` describes such a parameter, if it makes sense.
- In the docs for the `f` being a formula case, describe `.y` in the same manner as `g`, if `.y` can indeed be used [I'm guessing that it is usable].

Originally mentioned in PR #85.

brookslogan:
Still see remnants of the old docs somehow.

PR #144, opened 2022-07-13
Update f documentation in epi_slide

rachlobay:
Addressed Issue #143 as per discussions with Logan. See the documentation for the `f` parameter in `epi_slide` and let me know if any re-wording would be helpful here.


rachlobay:
@dajmcdon. This pull request has the two changes to `[.epi_df` that we discussed earlier today (1) removal of the two-ish lines of code in which an `epi_df` is converted to a tibble if column subsetting is performed and 2) Removal of the `%>% group_by(geo_value)` in the last two tests for `epi_slide` because `as_tibble()` in `[.epi_df` drops the grouping). Please review when you can (I think this should fix the R CMD Check errors for Maggie and Ken).

dajmcdon (approved):

dajmcdon (approved):
One minor issue. Can you remove the .DS_Store file? Just add `*.DS_Store` to .gitignore

Issue #137, opened 2022-07-07
Fix `epix_slide` example referring to nonexistent column


brookslogan:
```
options(warn=1L)
devtools::run_examples()
```
produces (along with some other output and warnings)
```
> epix_slide(x = archive_cases_dv_subset,
+            f = ~ mean(.x$case_rate),
+            n = 3,
+            group_by = geo_value,
+            ref_time_values = time_values,
+            new_col_name = 'case_rate_3d_av')
Warning: Unknown or uninitialised column: `case_rate`.
Warning in mean.default(.x$case_rate) :
  argument is not numeric or logical: returning NA
Warning: Unknown or uninitialised column: `case_rate`.
Warning in mean.default(.x$case_rate) :
  argument is not numeric or logical: returning NA
Warning: Unknown or uninitialised column: `case_rate`.
Warning in mean.default(.x$case_rate) :
  argument is not numeric or logical: returning NA
[....................]
```
It looks like the column has been renamed or removed (maybe while addressing #83):
```
> archive_cases_dv_subset[["DT"]] %>% names()
[1] "geo_value"       "time_value"      "version"         "percent_cli"    
[5] "case_rate_7d_av"
```
So we should either change the column used for this computation, or reintroduce the old column and address #83 via selecting the interesting columns somewhere within those examples.

rachlobay:
I could be wrong, but I think this relates back to changes made a couple months ago to the internal data (for ex. https://github.com/cmu-delphi/epiprocess/commit/d8cf12485622e95038ec8c56197af6eba3004658) and so looking at those changes may or may not help here.

mgyliu:
Yeah I traced it down to the same source. Thanks for commenting @rachlobay!

[...]

PR #149, opened 2022-07-18
[Issue 137] Fix epix slide example

mgyliu:
This PR addresses Issue #137. It changes an example in `epix_slide` to fix a warning. 

Context: previously, the column being used in the example was nonexistent in the dataset. The issue arose because the dataset column was renamed, but calls to the old column name were not updated.


dajmcdon:
Closes #137 .

PR #156, opened 2022-07-21
Ignore `NA`s when printing an archive's min/max `time_value`

nmdefries:
### Description
Toggle on the `na.rm` option (via the custom `Min` and `Max` functions) so even if an `epi_archive`'s `time_value` field contains some missing values (`NA`) printed min/max values are still useful. Prints `NA` if _all_ `time_value`s are missing and "no non-missing arguments" warning is raised, or if an un-caught warning occurs.

### Changes
- `R/archive.R` - added `na.rm` setting and try-catch to handle all-missing case.
- `man/as_epi_df.Rd` - not my changes, showed up when I re-`document()`ed

### Fixes
Closes https://github.com/cmu-delphi/epiprocess/issues/106

### DIFF
--- a/R/archive.R
+++ b/R/archive.R
@@ -188,28 +188,16 @@ epi_archive =
               })
             }
             cat("----------\n")
+            if (length(self$DT$time_value) == 0 || all(is.na(self$DT$time_value)) || all(is.nan(self$DT$time_value))) {
+              min_time = max_time = NA
+            } else {
+              min_time = Min(self$DT$time_value)
+              max_time = Max(self$DT$time_value)
+            }
             cat(sprintf("* %-14s = %s\n", "min time value",
-                        tryCatch({
-                          Min(self$DT$time_value)
-                        },
-                        warning = function(w) {
-                          if (w$message != "no non-missing arguments to min; returning Inf") {
-                            warning(w)
-                          }
-                          NA
-                        }
-                        )))
+                          min_time))
             cat(sprintf("* %-14s = %s\n", "max time value",
-                        tryCatch({
-                          Max(self$DT$time_value)
-                        },
-                        warning = function(w) {
-                          if (w$message != "no non-missing arguments to max; returning -Inf") {
-                            warning(w)
-                          }
-                          NA
-                        }
-                        )))
+                          max_time))
             cat(sprintf("* %-14s = %s\n", "min version",
                         min(self$DT$version)))
             cat(sprintf("* %-14s = %s\n", "max version",
--- a/R/archive.R
+++ b/R/archive.R
@@ -189,9 +189,27 @@ epi_archive =
             }
             cat("----------\n")
             cat(sprintf("* %-14s = %s\n", "min time value",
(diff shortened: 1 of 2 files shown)


brookslogan:
Thanks, I like the described behavior.  Implementation-wise, I'm not sure if the warning message could vary between different versions of R.  (+ pattern-wise we might want to keep a habit of using `Warn` with a `parent` arg rather than `warning(w)` to rethrow... not sure it's better here but it would be in more complicated cases where we don't want to lose stack traces that we could enable.)  Maybe the tryCatch could be replaced with a check on whether the vector is length-0 or all NA (or NaN)?

brookslogan:
Not sure if I was too afraid of "brittleness" above.  Slight tweak to the current logic to get it closer to how the warning is triggered:

`|| all(is.na(self$DT$time_value)) || all(is.nan(self$DT$time_value))` looks like it wouldn't catch a mix of NAs and NaNs, so it'd seem like we'd want `|| all(is.na(self$DT$time_value) | is.nan(self$DT$time_value))` to detect a mix of NA and NaNs.  However, it turns out that using `is.nan` isn't even necessary: `?is.na` states that:
> The default method for 'is.na' applied to an atomic vector returns a logical vector of the same length as its argument 'x', containing 'TRUE' for those elements marked 'NA' or, for numeric or complex vectors, 'NaN', and 'FALSE' otherwise. (A complex value is regarded as 'NA' if either its real or imaginary part is 'NA' or 'NaN'.)  'dim', 'dimnames' and 'names' attributes are copied to the result.
So we can just change the above to `|| all(is.na(self$DT$time_value))`.  I did that in the commit above.

Side note from testing: we can't directly construct an epi_archive with all-NaN time_values, although I see no use cases for this.
```r
as_epi_archive(tibble::tibble(geo_value=1,time_value=NaN,version=1,value=1))
#> Error in if (is.numeric(time_value) && all(time_value == as.integer(time_value)) &&  (from utils.R#93) : 
#>  missing value where TRUE/FALSE needed
```
I do see a use case for all-NA time values in an epi_archive, to provide version tracking for non-time-series data (or maybe time series in a `nest`ed format for some reason?) using epi_archive, which insists on a time_value column.

brookslogan:
To clarify in case we revisit this thread: we _are_ currently able to directly construct an epi_archive with all NA `time_value`s, just not all `NaN` ones.  This PR properly `print`s the `time_value` range of an `epi_archive` given any mixture of these that we can directly construct or sneak in via mutating the DT contents or reseating the DT field.

brookslogan (approved):

nmdefries:
Thanks for the fix.

> Side note from testing: we can't directly construct an epi_archive with all-NaN time_values

In my testing, it seems that we can. The problem with construction is that the `time_type` auto-detection doesn't work -- I'm getting a

```
Error in if (is.numeric(time_value) && all(time_value == as.integer(time_value)) &&  : 
  missing value where TRUE/FALSE needed
```
but if we set `time_type` manually, the archive is successfully created.

Issue #146, opened 2022-07-17
Clarify naming and fix defaults: `epix_slide` is over versions


brookslogan:
- [x] ~Make `ref_time_values` default to something involving versions, not all time values; `full_seq` up to `observed_versions_end` with period determined using `tsibble::guess_frequency` or an alternative to avoid potential licensing complications~ the default value setting is now https://github.com/cmu-delphi/epiprocess/issues/153; the renaming will be part of a different discussion/issue
- [x] ~Rename `ref_time_values` to something involving "version"~ part of separate discussion
- [x] ~Rename `n` to something that doesn't imply it's counting something~ replace with `before` (+add issue about dot prefixes)
- [x] Remove any `n` default value (e.g., 7 doesn't make sense; we won't get 7 time values in typical data sources)
- [x] ~Rename `time_value` column in output to `version`~ part of separate discussion
- [ ] ~Consider making output an `epi_archive`... but we might need to invent geo and time columns...~ part of larger discussion
- [x] Fix up documentation, examples, vignettes, etc. to make sense and work with the above

brookslogan:
We should try to avoid using tsibble if possible.  Another issue with the `full_seq` suggestion is that it could be altered by compactify; we'd at least want to document this, by making sure `compactify` descriptions don't promise full compatibility and noting specifically for the `ref_time_values`/new-name documentation.  This might also lead us to think of a different post-`compactify` format that records other versions that had updates reported.  (This also hints at making `epi_archive` provide a different interpretation: the current setup is as if we continuously and perfectly monitored some data source for updates within some time range; an alternative setup is to say that we only observed a certain set of versions.  It even more hassle for a user to provide this version set explicitly, but we could default to something like the default I was describing.  This alternative interpretation would mean that as_of should not allow us to ask for versions that weren't measured or should return NAs.  It's a bit more complex, but also more precise/realistic.)

ryantibs:
Thanks for raising these! I'd like to talk some of these through before you guys get too far along to understand what the problems are.

I tried to design `epix_slide()` to be as close as possible to `epi_slide()`, thinking this was a feature, rather than a bug. Thus the way I see it is that the name of the function itself, and all its argument names, being close is a good thing.

Hence I like items 1 and 4 in your todo list, but I'm not sure I like the others which are suggesting we rename things (items 2, 3, 5). 

[...]

PR #157, opened 2022-07-21
Possibly fixed 1-5 and 7 of issue #146

kenmawer:
This is based on my own interpretations, before the meeting


brookslogan:
Sorry to make some of this work have to be split out or shelved.  A summary of the meeting is [here](https://github.com/cmu-delphi/epiprocess/issues/146#issuecomment-1192785302).  I know we called the agreed-upon changes the "current issue" (#146), but I feel like so many things were removed that we should just treat it as separate issues: #160 and #161.  Some of the work here could be split out / recreated to address those only.

PR #158, opened 2022-07-21
Added epi_archive method tests, excluding for epix_slide

kenmawer:
The tests for `epix_slide` and edits would be in separate branches. This PR would introduce no coupling.

### DIFF
--- a/tests/testthat/test-methods-epi_archive.R
+++ b/tests/testthat/test-methods-epi_archive.R
@@ -1,0 +1,67 @@
+library(dplyr)
+
+ea <- archive_cases_dv_subset$clone()
+
+# epix_as_of tests
+test_that("epix_as_of behaves identically to as_of method",{
+  expect_identical(epix_as_of(ea,max_version = min(ea$DT$version)),
+                   ea$as_of(max_version = min(ea$DT$version)))
+})
+
+test_that("Errors are thrown due to bad as_of inputs",{
+  # max_version cannot be of string class rather than date class
+  expect_error(ea$as_of("2020-01-01"))
+  # max_version cannot be later than latest version
+  expect_error(ea$as_of(as.Date("2025-01-01")))
+  # max_version cannot be a vector
+  expect_error(ea$as_of(c(as.Date("2020-01-01"), as.Date("2020-01-02"))))
+})
+
+test_that("Warning against max_version being same as edf's max version",{
+  expect_warning(ea$as_of(max_version = max(ea$DT$version)))
+  expect_warning(ea$as_of(max_version = min(ea$DT$version)),NA)
+})
+
+test_that("as_of properly grabs the data",{
+  df_as_of <- ea %>%
+    epix_as_of(max_version = as.Date("2020-07-01")) %>%
+    na.omit() %>%
+    as.data.frame()
+    
+  df_filter <- ea$DT %>%
+    filter(version == as.Date("2020-07-01")) %>%
+    na.omit() %>%
+    select(-version) %>%
+    as.data.frame()
+  
+  expect_equal(df_as_of[1:4],df_filter)
+})
+
+# epix_merge tests
+test_that("epix_merge requires second argument to be a data.table or
+          epi_archive",{
+  expect_error(epix_merge(ea,data.frame(x=1)))
+})
+
+test_that("data.table merging is utilized if second argument is a data.table",{
+  dt1 <- select(ea$DT , -case_rate_7d_av)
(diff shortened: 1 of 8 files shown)


tests/testthat/test-methods-epi_archive.R
```diff
@@ -0,0 +1,67 @@
...
+
+test_that("data.table merging works as intended",{
+  ea <- archive_cases_dv_subset$clone()
+  dt1 <- select(ea$DT , -case_rate_7d_av)
+  ea1 <- as_epi_archive(dt1)
+  dt2 <- select(ea$DT , -percent_cli)
+  
+  expect_identical(
+    as_epi_archive(ea$DT),
+    as_epi_archive(merge(dt1,dt2))
```
brookslogan, line 65: `merge` should always be using `all=TRUE`.  I have a change to the interface that will enforce this on #101, but we should incorporate `all=TRUE` in this example just in case we don't use those & copy from this example later.

tests/testthat/test-methods-epi_archive.R
```diff
@@ -0,0 +1,67 @@
...
+    as.data.frame()
+    
+  df_filter <- ea$DT %>%
+    filter(version == as.Date("2020-07-01")) %>%
+    na.omit() %>%
+    select(-version) %>%
+    as.data.frame()
+  
+  expect_equal(df_as_of[1:4],df_filter)
+})
```
brookslogan, line 38: Please replace this with a hand-crafted example that tests both that `as_of` incorporates data from `max_version` as well as uses LVCF.  E.g., have an observation that's revised from 1 to 2 to 3 to 4, and as for it `as_of` between the 3 and 4 versions; it should output 3 (and not give any extra junk related to 1 or 2, nor somehow see 4).

brookslogan (changes requested):

brookslogan:
Heads-up:
- `clone` doesn't do a full deep copy, even if `deep=TRUE`, because it doesn't specially handle the DT.  I don't think it actually ends up mattering here, because current `epix_merge` and `$merge` reseat the DT reference rather than modifying the DT contents, but it's just something to be aware of if you weren't already.  See #164.
- In #101 I have a change to the `epix_merge` and `$merge` interface and behavior that will remove the undocumented merge-with-data.table behavior + other behaviors + bugfixes.  The test here can be kept, but merging with #101 will require removing this change (+ also reverting the `all=TRUE` setting I've suggested above).

tests/testthat/test-methods-epi_archive.R
kenmawer: I just did so.

brookslogan:
Has this already been merged in via a different PR?

kenmawer:
> Has this already been merged in via a different PR?

Yes, although I don't know how that even got there. I just put in a put request, and it somehow got merged.

tests/testthat/test-methods-epi_archive.R
brookslogan: My comment here is a little imprecise, will explain below.  Changes here are good though.

brookslogan:
Correction to my comment above: while we do want all=TRUE all the time when merging two archives, we don't necessarily want it when merging an archive and a data.table; in this particular case, since the data.table is basically an archive, we would want all=TRUE, though.

The first part of the `ea2` construction could have been simplified using the handy `rep` function and vector operations:
```r
  ea2 <- tibble(
    geo_value = "ca",
    time_value = d + rep(0:3, each=4L),
    version = d + rep(0:3, times=4L)
  ) %>%
    filter(version >= time_value)
```
However, since this undergoes additional filtering and such, I think it's clearer to just directly construct the final form of ea2 from its entries:
```r
  ea2 = tibble::tribble(
    ~geo_value, ~time_value,      ~version, ~cases,
          "ca", "2020-06-01", "2020-06-01",      1,
          "ca", "2020-06-01", "2020-06-02",      2,
    #
          "ca", "2020-06-02", "2020-06-02",      0,
          "ca", "2020-06-02", "2020-06-03",      1,
          "ca", "2020-06-02", "2020-06-04",      2,
    #
          "ca", "2020-06-03", "2020-06-03",      1,
    #
          "ca", "2020-06-04", "2020-06-04",      4,
    ) %>%
    dplyr::mutate(dplyr::across(c(time_value, version), as.Date)) %>%
    as_epi_archive()
  ```
(I got this from printing ea2$DT and doing some text editing; not sure if there's some easier way.)  I pushed a change to use this way of constructing ea2.

PR #159, opened 2022-07-21
Updated testing for instatiation of an epi_archive

kenmawer:


tests/testthat/test-archive.R
```diff
@@ -0,0 +1,24 @@
+library(dplyr)
+
+test_that("first input must be a data.frame",{
+  expect_error(as_epi_archive(c(1,2,3)))
```
brookslogan, line 4: For the `expect_error`s here and below, we want to make sure that we're actually getting stopped by the checks that we expect to fail.  `expect_error` has at least a couple optional arguments to help here: `regexp` and `class`.  At least for the test above, only `regexp` is helpful, because we haven't assigned a custom condition class to the error message:
```r
Abort("`x` must be a data frame.")
```
However, in places where we pass a `class` argument to `Abort` in the checks, we can pass a `class` argument to `expect_error` to only match that class of errors.

tests/testthat/test-archive.R
```diff
@@ -0,0 +1,24 @@
...
+})
+
+dt <- archive_cases_dv_subset$DT
+
+test_that("data.frame must contain geo_value, time_value and version columns",{
+  expect_error(as_epi_archive(select(dt,-geo_value)))
+  expect_error(as_epi_archive(select(dt,-time_value)))
+  expect_error(as_epi_archive(select(dt,-version)))
+})
```
brookslogan, line 13: Please add a test for this warning:
```
Warn("`additional_metadata` names overlap with existing metadata fields \"geo_type\", \"time_type\".")
```

tests/testthat/test-archive.R
```diff
@@ -0,0 +1,24 @@
...
+test_that("other_keys can only contain names of the data.frame columns",{
+  expect_error(as_epi_archive(dt,other_keys = "xyz"))
+  expect_error(as_epi_archive(dt,other_keys = "percent_cli"),NA)
+})
+
+test_that("other_keys cannot contain names geo_value, time_value or version",{
+  expect_error(as_epi_archive(dt,other_keys = "geo_value"))
+  expect_error(as_epi_archive(dt,other_keys = "time_value"))
+  expect_error(as_epi_archive(dt,other_keys = "version"))
+})
```
brookslogan, line 24: Please also add a test that the key of the archive's $DT is set properly, regardless of what kind of (valid) `x` arg it is given: a vanilla data frame, a tibble, a keyed data.table, an unkeyed data.table, and maybe even an epi_df.  See [here](https://github.com/dajmcdon/epiprocess/blob/74f989748a5f63d400d161608898a36f3eb6c7ce/R/archive.R#L168) how setting the key has some gotchas.  Probably good to test to make sure we don't break this again because I don't think we'd notice just from reading the code.

brookslogan (changes requested):

tests/testthat/test-archive.R
brookslogan: "keyed data.table" isn't precise enough.  We need to test constructing an archive from an `x` arg that's a data.table with a key that's not the same as the one we want the $DT field of the archive to have.

kenmawer: I did one where the DT's keys are different from what is passed on to the argument for `as_epi_archive`.

brookslogan: Thanks for the extensive tests!  I think we might be missing one more case: an epi_df with nontrivial other_keys metadata.
- If we provide `other_keys` to `as_epi_archive`, it should override the `epi_df`'s `other_keys` metadata
- If we don't, we should use the `epi_df`'s `other_keys` metadata (so if the epi df had "code" as another key, we the archive by default should have key cols geo_value, time_value, code, version)

brookslogan:
Hopefully just one more test to add (above).  Please also address the warnings emitted by check() when running the tests:
https://github.com/cmu-delphi/epiprocess/pull/159/checks#step:5:245

We should either capture/muffle these warnings, or pass/set things up so they are never generated (e.g., by passing `compactify=TRUE` if these are all just compactification warnings)

brookslogan (changes requested):
Triggering GitHub to mark my review as complete, so that you can re-request review once again when you think these final couple of items are complete.  (Trying to keep on top of reviews using https://github.com/pulls/review-requested)

brookslogan:
Also, I realize that the as_epi_archive behavior here on a keyed DT when we don't pass other_keys seems a bit fishy.  If they've keyed the thing appropriately with geo_type, time_type, <other keys>, version already, I believe this means that we drop all their <other keys> despite the data table key looking like a valid epi_archive DT key.  Probably we want some type of warning here at least.  Could be a separate issue.

kenmawer:
I agree with the most recent comment. However, I believe I have already seen this as an issue.

kenmawer:
Note that I may be mixing up a different issue with what should be addressed with the `key` being dropped.

brookslogan:
I didn't check to see if this was already filed as an Issue or not.  Could you please confirm whether or not it is already filed?

brookslogan:
Also I realize I never "unresolved" the conversation with the "one more test to add (above)" so it was visible again --- please take a look.

tests/testthat/test-archive.R
kenmawer: I just found that `additional_metadata` isn't preserved.

brookslogan: So it looks like we have a few unwanted behaviors:
- keyed data.table + no other_keys arg --> we ignore the current key & replace with default geo time version
- epi_df with other_keys + no other_keys arg --> same deal
- epi_df with additional metadata --> no $additional_metadata

Feel like tackling those here, or filing them as issues + commenting in the tests that this is /unwanted/ behavior that should change?

brookslogan: As for the desired behavior regarding key-setting with no other_keys arg, I've described different things for the keyed data.table vs. the epi_df, but we might want to do roughly the same thing (this is different from the things I described before):
- if input's key includes geo_value and time_value, use `c(input_key[input_key != "version"], "version")` as the epi_archive key --- if they already had "version" as part of the key, this reorders it to be at the end of the key (some epi_archive operations rely on this ordering), and if "version" wasn't already part of the key, it adds it at the end.

brookslogan: For the additional_metadata, I think could just
- use the `additional_metadata` arg if provided
- use the stuff from the epi_df's metadata that makes sense if they don't provide an `additional_metadata` arg

kenmawer:
I didn't see the issue elsewhere about keys not being retained, so I filed it as #201.

tests/testthat/test-archive.R
brookslogan: Okay, I think we have this filed away in issues.  Not sure if the additional test here was added, but that could be part of those issues.

brookslogan (approved):

Issue #160, opened 2022-07-22
Remove default `n` value in `epi_slide`, `epix_slide`


brookslogan:
Make sure tests, examples, and vignettes still run.

kenmawer:
I put in a pull request that would address this issue. Issue #161 will only begin to be addressed on this pull request is approved, as it's more complicated and builds on this issue.

brookslogan:
Closed by https://github.com/cmu-delphi/epiprocess/pull/167.

PR #167, opened 2022-07-26
Remove default `n` values in `epi_slide` and `epix_slide`

kenmawer:
Addresses issue #160.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -56,7 +56,7 @@ Config/testthat/edition: 3
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.0
+RoxygenNote: 7.2.1
 Depends: 
     R (>= 2.10)
 URL: https://cmu-delphi.github.io/epiprocess/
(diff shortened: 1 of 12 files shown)


brookslogan:
Code changes look good to go.  Doc changes look good, but see question below.  Interestingly, as far as I can tell, the vignettes are all already compatible with this change!

Question: is it worth clarifying in the @param n docs (for epi_slide, epix_slide, $slide) that `n` must be passed by name?

brookslogan:
I don't think GitHub recognizes "addresses", but does recognize "closes" and "resolves".  Closes #160.

kenmawer:
Note that I still haven't done `epix_slide` yet.

kenmawer:
Note that this is for `epi_slide` only and not `epix_slide`.

brookslogan:
In GitHub's diffs, I see changes in epi_slide, epix_slide, and epi_archive's $slide.  Are some of them not supposed to be there?

kenmawer:
These should be here, because we are removing the default `n` values. Any changes in the documentation, etc. is due to taking out the now outdated info about the default `n` being 7 where it's specified.

brookslogan:
Sorry, having trouble loading in context.

> Note that this is for `epi_slide` only and not `epix_slide`.

Could you please explain this a bit further?

kenmawer:
My point was that I did not make changes relating to removing the default `n` value in `epix_slide`. However, I did these changes for `epi_slide`.

brookslogan:
Note from call: removing the `n` default did apply to both epi slide and epix slide.  The change that didn't apply to both was in another PR.

Issue #182, opened 2022-08-02
Add checks & test for `additional_metadata` format in `epi_df`


mgyliu:
# Action items

- [x] Update examples in [`as_epi-df`](https://cmu-delphi.github.io/epiprocess/reference/as_epi_df.html) so that `additional_metadata` is always a `list()` type.
- [x] Add example in `as_epi_df` with multiple `other_keys`
- [ ] (Maybe) Add a test in `epipredict` to fit a model on an `epi_df` with multiple `other_keys`. 
   - TODO (mgyliu): create & link an `epipredict` issue for this

# The problem

`other_keys` gets stored differently in `epi_df` if you initialize it as a vector vs. a list. The `epi_df` constructor expects `additional_metadata` to be a `list` but passing a `vector` still "works" (i.e., no error). Examples are shown below for each case. The examples are taken from the [`as_epi-df` reference](https://cmu-delphi.github.io/epiprocess/reference/as_epi_df.html). 

```
library(dplyr)
library(recipes)
library(epiprocess)
```

### Example 1
```
ex1_input <- tibble::tibble(
  geo_value = rep(c("ca", "fl", "pa"), each = 2),
  county_code = c("06059","06061","06067",
                  "12111","12113","12117"),
  another_key = 1:6, # <- I added this additional key 
  time_value = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
                       by = "day"), length.out = length(geo_value)),
  value = 1:length(geo_value) + 0.01 * rnorm(length(geo_value))) %>% 
  tsibble::as_tsibble(
    index = time_value, 
    key = c(geo_value, county_code, another_key))

# The `other_keys` metadata (`"county_code"` in this case) is automatically
# inferred from the `tsibble`'s `key`:
ex1 <- as_epi_df(x = ex1_input, geo_type = "state", time_type = "day", as_of = "2020-06-03")
attr(ex1,"metadata")
```
Output:
```
$geo_type
[1] "state"

$time_type
[1] "day"

$as_of
[1] "2020-06-03"

$other_keys
[1] "county_code" "another_key"
```
### Example 2 (but ex3 from the as_epi_df documentation so we'll keep the names)

```
ex3_input <- jhu_csse_county_level_subset %>%
  dplyr::filter(time_value > "2021-12-01", state_name == "Massachusetts") %>%
  dplyr::slice_tail(n = 6) %>% 
  tsibble::as_tsibble() %>% # needed to add the additional metadata
[...]

Issue #166, opened 2022-07-26
Document the differences for column re-ordering in `new_epi_df` and `as_epi_df`


rachlobay:
It probably would be good to document the differences for column re-ordering in`new_epi_df` and `as_epi_df`. For ex., `new_epi_df` always has column re-ordering so that the columns are `geo_value, time_value ...` . In contrast, `as_epi_df` does not appear to enforce this column re-ordering when we're already feeding an epi_df to it (it simply returns the `epi_df` object unchanged).  But it does enforce the column re-ordering if we are converting a `tbl_df`, `data frame`, or `tbl_ts` to an `epi_df` (as it ultimately calls `new_epi_df`).

PR #183, opened 2022-08-03
[Issue 182] Update `epi_df` examples, check `additional_metadata` type at construction

mgyliu:
Addresses #182 

Changes:
- Update `new_epi_df` examples so that (i) `additional_metadata` argument is type `list` instead of `vector` and (ii) there is an example with multiple `other_keys`
- Add a check in `new_epi_df` to ensure `additional_metadata` is a `list`
- Write a test for the previous change

I checked in `epipredict` for calls to `as_epi_df` and all `additional_metadata` arguments there are `lists` so this change will not break `epipredict`. Afaik there aren't any dependencies on `epiprocess` other than `epipredict`.

### DIFF
--- a/R/epi_df.R
+++ b/R/epi_df.R
@@ -106,8 +106,8 @@ NULL
 #' @param additional_metadata List of additional metadata to attach to the
 #'   `epi_df` object. The metadata will have `geo_type`, `time_type`, and
 #'   `as_of` fields; named entries from the passed list will be included as
-#'   well. If your tibble has additional keys, be sure to specify them 
-#'   in the `other_keys` component of `additional_metadata`. 
+#'   well. If your tibble has additional keys, be sure to specify them as a
+#'   character vector in the `other_keys` component of `additional_metadata`. 
 #' @param ... Additional arguments passed to methods.
 #' @return An `epi_df` object.
 #' 
@@ -190,8 +190,8 @@ new_epi_df = function(x = tibble::tibble(), geo_type, time_type, as_of,
 #' @param additional_metadata List of additional metadata to attach to the
 #'   `epi_df` object. The metadata will have `geo_type`, `time_type`, and
 #'   `as_of` fields; named entries from the passed list will be included as
-#'   well. If your tibble has additional keys, be sure to specify them 
-#'   in the `other_keys` component of `additional_metadata`.
+#'   well. If your tibble has additional keys, be sure to specify them as a
+#'   character vector in the `other_keys` component of `additional_metadata`.
 #' @param ... Additional arguments passed to methods.
 #' @return An `epi_df` object.
 #'
--- a/R/epi_df.R
+++ b/R/epi_df.R
@@ -117,7 +117,11 @@ new_epi_df = function(x = tibble::tibble(), geo_type, time_type, as_of,
   if (!is.data.frame(x)) {
     Abort("`x` must be a data frame.")
   }
-  
+    
+  if (!is.list(additional_metadata)) {
+    Abort("`additional_metadata` must be a list type.")
+  }
+
   # If geo type is missing, then try to guess it
   if (missing(geo_type)) {
     geo_type = guess_geo_type(x$geo_value)
@@ -230,7 +234,7 @@ new_epi_df = function(x = tibble::tibble(), geo_type, time_type, as_of,
 #' 
 #' ex2 <- ex2_input %>% dplyr::rename(geo_value = state, time_value = reported_date) %>%
 #'   as_epi_df(geo_type = "state", as_of = "2020-06-03", 
-#'             additional_metadata = c(other_keys = "pol"))
+#'             additional_metadata = list(other_keys = "pol")) 
 #' 
 #' attr(ex2,"metadata")
 #' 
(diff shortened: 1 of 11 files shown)


dajmcdon (changes requested):
Everything here looks good.

As I read this though, I think the `epi_df` and `as_epi_df` documentation should be more specific. A `tbl_ts` is converted so that additional keys go as `other_keys = xxx`. I think we should mention this somewhere.

If someone wants to make an `epi_df` to use with `epipredict` that correctly uses those keys, they MUST be called `other_keys` or they'll be ignored.

R/epi_df.R
```diff
@@ -244,8 +248,10 @@ new_epi_df = function(x = tibble::tibble(), geo_type, time_type, as_of,
 #' 
 #' ex3 <- ex3_input %>% 
 #'   tsibble::as_tsibble() %>% # needed to add the additional metadata
-#'   dplyr::mutate(state = rep("MA",6)) %>%
-#'   as_epi_df(additional_metadata = c(other_keys = "state"))
+#'   dplyr::mutate(
+#'     state = rep("MA",6),
+#'     pol = rep(c("blue", "swing", "swing"), each = 2)) %>% # 2 extra keys
+#'   as_epi_df(additional_metadata = list(other_keys = "state", "pol"))
```
brookslogan, line 254: This should be `list(other_keys = c("state", "pol"))`.

dajmcdon: Good catch!

brookslogan:
If `other_keys` settings are causing issues, we should consider printing them in `print.epi_df` rather than having the user have to separately print the metadata to see.

R/epi_df.R
mgyliu: Thanks for catching this! Just pushed up a fix

mgyliu:
@dajmcdon Good suggestions - I'm thinking of adding a blurb in the `additional_metadata` section under ["Arguments"](https://cmu-delphi.github.io/epiprocess/reference/as_epi_df.html#arguments) that talks about `other_keys`. The `as_epi_df` documentation does  already state those things about `tbl_ts` conversion, under ["Methods by class"](https://cmu-delphi.github.io/epiprocess/reference/as_epi_df.html#methods-by-class-). Perhaps the verbiage isn't clear enough? What do you think?

> If someone wants to make an `epi_df` to use with `epipredict` that correctly uses those keys, they MUST be called `other_keys` or they'll be ignored.

^ To be clear, we would just add some documentation to make this clear to the user? i.e., not "strictly" enforce it in the "throw an error if the specification is wrong" sense?

I do agree with @brookslogan's suggestion of having `other_keys` as its own argument in those functions. It seems important enough to not be considered "metadata" at all? For this PR though, I'll add the extra documentation and open an issue for Logan's suggestions.

dajmcdon:
I agree that `as_epi_df()` would "get it right" when converting from a `tbl_ts`. But the documentation for `as_epi_df` doesn't say it very explicitly.

So, if I call `as_epi_df()` on a regular tibble that has some extra keys, the docs don't tell me what to do with that. It should say something like: "if your tibble has extra keys, be sure to add them as a character vector to the `other_keys` component of the `additional_metadata` list or bad things could happen". The same goes for the docs on `new_epi_df()`.

mgyliu:
@dajmcdon Gotcha. I just pushed up changes to the `as_epi_df` and `new_epi_df` documentation to explain this under the `additional_metadata` argument. Also updated the `epiprocess` vignette with an explanation. Let me know what you think!

dajmcdon (approved):
Looks good to me. @brookslogan?

brookslogan:
Requests/questions:
- I'm afraid `new_epi_df` vs. `as_epi_df` might not be clear from the roxygen docs.  Perhaps the description of one or both could include a compare&contrast or rationale?  (I think the most important thing to note here is the behavior on `epi_df`s --- that `as_epi_df` will return an `epi_df` exactly as is, keeping the current column order and ignoring any metadata arguments, while `new_epi_df` will order `geo_value` and `time_value` first, and reset/overwrite --- most --- metadata.)  E.g., like in `?rlang::new_quosure` / `?rlang::as_quosure`, `tibble::new_tibble`, `dplyr::new_grouped_df`, etc.
- Is `as_of` the only metadata field that `new_epi_df` will preserve when called on an `epi_df`?  Do we want it that way?
- If the docs for all of the common parameters of these two functions are the same, use `@inheritParams <otherfn>` in one of them to not have to copy-paste all the roxygen updates.  (If they're not the exact same, there are more tedious ways to reduce duplication, but maybe that could be a different issue.)

A few lingering thoughts which probably belong in separate issues:
- Unfortunately, I don't think we ever explained what "keys" were in the context of an `epi_df`, and googling "tibble key" is surprisingly unhelpful (currently giving links to lock picks and decoder tools).  Maybe explaining this could just be rolled into the issue for adding `other_keys` as a parameter, though; it'd be easier to do in its own `@param`.
- `ex3` requiring conversion away from an `epi_df` in order not ignore `other_keys` and all the other parameters seems error-prone.  There is precedent with `rlang::{new,as}_quosure`, but still.  Wondering what we should do here; maybe something like calling `Abort` if the user passes in metadata that doesn't match the pre-existing metadata.

brookslogan:
Migrating the notes above to other issues, so this is good to go!

brookslogan:
@mgyliu should we now close #182 as completed?

I've copied/moved some of the notes above from here and in #121 into #166 and #186.

mgyliu:
Thanks @brookslogan! Yes I'll close 182.

Issue #161, opened 2022-07-22
Replace `n` in `epi[x]_slide` with `before`, `after`


brookslogan:
See https://github.com/cmu-delphi/epiprocess/issues/146#issuecomment-1192785302 for details.

brookslogan:
This has been completed (though we plan to revert).

Issue #177, opened 2022-07-30
Periodically bump version, perform GitHub release, maintain news/changelog


brookslogan:
Raised by @dshemetov a while back: we should periodically bump the package version and perform a GitHub release.  Maintaining a news/changelog file would also be helpful.  Version numbers help with lifecycle deprecation messaging.  We should think about how backwards-compatible we want to be, and how much we want to catch with deprecation messages.

dshemetov:
I suggest a simple and common approach to version numbering (h/t @krivard): `A.B.C`, where `A` increments for large backwards-incompatible versions, `B` increments for large feature additions (backwards compatible), `C` increments for minor changes/fixes.

krivard:
Credit: https://semver.org/

brookslogan:
We're "at" version 1.0.0 for a while already but are still cleaning up some interface issues.  With current pull-into-main + semver we will reach ridiculous version numbers quickly.  It feels like we are more in < 1.0.0 by semver standards, so I'm thinking of doing something a little semver-incompatible until we stabilize more --- maybe just call the current version 0.x.y.z or 1.x.y.z with x.y.z semver-like (I think this might be sort of like 1.0.0-alpha-x.y.z, but in an R numeric_version-compatible way).  When we switch to semver proper we should consider moving to pull-into-dev, at least for breaking changes.

dshemetov:
What do you think about [this viewpoint](https://tom.preston-werner.com/2022/05/23/major-version-numbers-are-not-sacred.html)?

dshemetov:
I'm really just in favor of simple solutions. The reason I prefer sticking with the original SemVer is because adding epochal version numbers is 
- treading new ground in terms of the convention
- makes us non-compliant with their spec
- makes it more likely that we have further issues
- sets us up for having to make rules for what decides an epoch

I feel like you haven't really addressed my suggestion of just bumping the version numbers up and letting them get large. The benefits are:
- we stay in SemVer and avoid the issues above
- you can start doing it now, without issue, on the PRs you have now

We COULD also delete the 1.0.0 release, since that was just a test and I don't think any one is using it. And revert back to 0.z.w to get us back to alpha. Maybe that's what you're suggesting in your route 3. I think that would also be fine.

dshemetov:
[...]

PR #188, opened 2022-08-06
Updated epi_slide to use `before` and `after` and added checks

kenmawer:
Note that `epix_slide` has been untouched.

Resolves issue #161.


R/slide.R
```diff
@@ -133,33 +122,30 @@ epi_slide = function(x, f, ..., n, ref_time_values,
...
+  if (before < 0 || after < 0) {
+    Abort("`before` and `after` must be at least 0.")
   }
+  
+  if (floor(before) < ceiling(before) || floor(after) < ceiling(after)) {
+    Abort("`before` and `after` must be integers.")
+  }
+  
+  before_num = before
+  after_num = after
```
brookslogan, line 148: Can we collapse the code in this area to something like the following?
```r
  if (missing(time_step)) {
    before_num = before
    after_num = after
  } else {
    before_num = time_step(before_num)
    after_num = time_step(after_num)
  }
```

tests/testthat/test-epi_slide.R
```diff
@@ -4,31 +4,55 @@ edf = dplyr::bind_rows(
   dplyr::tibble(geo_value = "ak", time_value = as.Date("2020-01-01") + 1:200, value=1:200),
   dplyr::tibble(geo_value = "al", time_value=as.Date("2020-01-01") + 1:5, value=-(1:5))
 ) %>%
-  as_epi_df()
+  as_epi_df() %>%
+  group_by(geo_value)
```
brookslogan, line 8: --> change name of `edf` to `gedf` or `grouped` or something like that?

tests/testthat/test-epi_slide.R
```diff
@@ -4,31 +4,55 @@ edf = dplyr::bind_rows(
...
 
 f = function(x, ...) dplyr::tibble(value=mean(x$value), count=length(x$value))
 
 ## --- These cases generate the error: ---
+test_that("`after` must be defined as a non-zero integer if `before` is missing", {
+  expect_error(epi_slide(edf, f, after = 0L, ref_time_values=as.Date("2020-01-01")),
+               "`before` cannot be missing when `after` is set to 0.")
+})
```
brookslogan, line 16: Also explicitly check the case where the user doesn't specify `after` at all.   (I assume that this triggers the same error as the above test given the current implementation.  But we want to ensure that it remains this way in the future.)  We should also check that the error message is helpful for debugging in this case.

tests/testthat/test-epi_slide.R
```diff
@@ -4,31 +4,55 @@ edf = dplyr::bind_rows(
...
+
+test_that("Both `before` and `after` must be nonnegative integers",{
+  expect_error(epi_slide(edf, f, before = -1L, ref_time_values=as.Date("2020-01-01")+2L),
+               "`before` and `after` must be at least 0.")
+  expect_error(epi_slide(edf, f, before = 2L, after = -1L, ref_time_values=as.Date("2020-01-01")+2L),
+               "`before` and `after` must be at least 0.")
+  expect_error(epi_slide(edf, f, before = 0.5, ref_time_values=as.Date("2020-01-01")+2L),
+               "`before` and `after` must be integers.")
+  expect_error(epi_slide(edf, f, before = 1, after = 0.5, ref_time_values=as.Date("2020-01-01")+2L),
+               "`before` and `after` must be integers.")
```
brookslogan, line 31: (In addition to non-length-1 before/after, as noted above) might want to detect NA before/after

tests/testthat/test-epi_slide.R
```diff
@@ -4,31 +4,55 @@ edf = dplyr::bind_rows(
...
                "starting and/or stopping times for sliding are out of bounds") # beyond the last, no data in window
 })
 
 test_that("`ref_time_values` + `align` that have some slide data, but generate the error due to ref. time being out of time range", {
-  expect_error(edf %>% group_by(geo_value) %>% epi_slide(f, n=3L, ref_time_values=as.Date("2020-01-01"), align="left"), 
+  expect_error(epi_slide(edf, f, before=0L, after=2L, ref_time_values=as.Date("2020-01-01")), 
                "starting and/or stopping times for sliding are out of bounds") # before the first, but we'd expect there to be data in the window
-  expect_error(edf %>% group_by(geo_value) %>% epi_slide(f, n=3L, ref_time_values=as.Date("2020-01-01")+201L), 
+  expect_error(epi_slide(edf, f, before=2L, ref_time_values=as.Date("2020-01-01")+201L), 
                "starting and/or stopping times for sliding are out of bounds") # beyond the last, but still with data in window
```
brookslogan, line 47: I think at some point we want to change this behavior, right?  Do you remember if this is already filed in an issue?

tests/testthat/test-epi_slide.R
```diff
@@ -4,31 +4,55 @@ edf = dplyr::bind_rows(
...
 test_that("`ref_time_values` + `align` that result in no slide data, generate the error", {
-  expect_error(edf %>% group_by(geo_value) %>% epi_slide(f, n=3L, ref_time_values=as.Date("2020-01-01")), 
+  expect_error(epi_slide(edf, f, before=2L, ref_time_values=as.Date("2020-01-01")), 
                "starting and/or stopping times for sliding are out of bounds") # before the first, no data in the slide windows
-  expect_error(edf %>% group_by(geo_value) %>% epi_slide(f, n=3L, ref_time_values=as.Date("2020-01-01")+207L), 
+  expect_error(epi_slide(edf, f, before=2L, ref_time_values=as.Date("2020-01-01")+207L), 
                "starting and/or stopping times for sliding are out of bounds") # beyond the last, no data in window
 })
 
 test_that("`ref_time_values` + `align` that have some slide data, but generate the error due to ref. time being out of time range", {
```
brookslogan, line 43: Could update test name now that `align` isn't a parameter

tests/testthat/test-epi_slide.R
brookslogan: We would want errors if there is no data in the window, but if there is partially data in the window, I assume we'd either want to uniformly allow it or uniformly reject it, but currently we aren't consistent (we allow partial data in window it for a trailing window for ref time values near min(time_value) in the edf, but don't allow it for manually-input ref time values slightly beyond the max(time_value)).  In any case, something for another issue.

vignettes/advanced.Rmd
```diff
@@ -222,15 +222,15 @@ x <- y1 %>%
     version = issue,
     percent_cli = value
   ) %>%
-  as_epi_archive()
+  as_epi_archive(compactify=FALSE)
```
brookslogan, line 225: For vignettes, we should use `compactify=TRUE` if possible.  I used `compactify=FALSE` in the data generation because otherwise some archive-related tests that relied on the uncompactified version of the example archive would fail.

vignettes/advanced.Rmd
```diff
@@ -222,15 +222,15 @@ x <- y1 %>%
...
 
 # mutating merge operation:
 x$merge(y2 %>%
   select(geo_value, time_value,
     version = issue,
     case_rate_7d_av = value
   ) %>%
-  as_epi_archive()
+  as_epi_archive(compactify=FALSE)
```
brookslogan, line 233: ditto

vignettes/aggregation.Rmd
```diff
@@ -177,7 +177,7 @@ Explicit imputation for missingness (zero-filling in our case) can be important
 for protecting against bugs in all sorts of downstream tasks. For example, even
 something as simple as a 7-day trailing average is complicated by missingness.
 The function `epi_slide()` looks for all rows within a window of 7 days anchored
-on the right at the reference time point (when `n = 7` and `align = "right"`).
+on the right at the reference time point (when `after = 6`).
```
brookslogan, line 180: Think this should be `before = 6`.  Anchored on the right at the reference time point == right-aligned == before is n-1 and after is 0

brookslogan: "left" and "right" are starting to get really confusing for me right now.  Maybe we should favor "trailing" and "leading", or "right-aligned/trailing", "left-aligned/leading"

brookslogan (changes requested):
Looks good.  Main requested changes: tweaks to the documentation to try to clarify a bit.  I've also noted some edge cases not caught by the before&after validation, seem pretty low-probability but while it's on our minds probably should add the checks.  + some misc minor things.

Please address + re-request review (and maybe ping me on Slack).  Remaining TODO on my end: double-check that we've stamped out all references to `n` and `align` in code, docs, tests, examples, vignettes.

R/slide.R
```diff
@@ -95,16 +85,14 @@
...
+#'   epi_slide(cases_7dav = mean(cases), before = 6) %>% 
 #'   # rmv a nonessential var. to ensure new col is printed
 #'   dplyr::select(-death_rate_7d_av) 
 #'  
 #'  # slide a left-aligned 7-day average
 #'   jhu_csse_daily_subset %>%
 #'   group_by(geo_value) %>%
-#'   epi_slide(cases_7dav = mean(cases), n = 7, 
-#'             align = "left") %>% 
+#'   epi_slide(cases_7dav = mean(cases), before = 6) %>%
```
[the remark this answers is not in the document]
kenmawer: Why should I mention `n` when it wouldn't exist due to its being replaced with `before` and `after`?

R/slide.R
kenmawer: I just replaced `before_num` with `before` and `after_num` with `after` as I deemed this redundant.

vignettes/aggregation.Rmd
kenmawer: I fixed the first comment.

As for the second comment, I did the latter.

tests/testthat/test-epi_slide.R
kenmawer: `is.numeric` returns `FALSE` for an NA, but I added those tests.

tests/testthat/test-epi_slide.R
kenmawer: What behaviour are you talking about? I don't quite understand what you mean here....

kenmawer:
Note: #203 should be used in favour of this due to the former also incorporating changes to `epix_slide`.

kenmawer:
Add case for centre alignment.

R/slide.R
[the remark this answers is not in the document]
brookslogan: This is if you add an example with center alignment based on an `n` variable, along the lines of what you implemented/moved to the outlier detection function.  If you don't add such an example, then it wouldn't be mentioned.  Such an example isn't necessary, but might be useful.

tests/testthat/test-epi_slide.R
brookslogan: Side note: saw an unrelated(?) pre-existing bug: some instances to `unique(x$time_value)` likely need to be sorted to be robust.

Current behavior:
- check for out-of-range start/stop time values only triggers if they are ALL out of range.
- a user-provided ref_time_values is silently filtered to down to those in unique(x$time_value)`.

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(epiprocess)
#> 
#> 
#>     filter
edf = tibble(geo_value="g1", time_value = 1:50, value=1:50) %>% as_epi_df()
# Silently excludes 51 (which would have partial data in window if we allowed
# it), but includes 1 (which has partial data):
edf %>% epi_slide(~ toString(.x$value), before=2L, ref_time_values=c(1,50,51))
#> An `epi_df` object, 2 x 4 with metadata:
#> * geo_type  = custom
#> * time_type = custom
#> * as_of     = 2022-08-15 15:30:44
#> 
#> # A tibble: 2 × 4
#>   geo_value time_value value slide_value
#> * <chr>          <int> <int> <chr>      
#> 1 g1                 1     1 1          
#> 2 g1                50    50 48, 49, 50
# Triggers error:
edf %>% epi_slide(~ toString(.x$value), before=2L, ref_time_values=51)
#> Error in `Abort()` at epiprocess/R/slide.R:189:4:
#> ! The starting and/or stopping times for sliding are out of bounds with
#> respect to the range of times in your data. Check your settings for
#> ref_time_values and align (and before, if specified).
```

<sup>Created on 2022-08-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I'm hoping for something more consistent.  Could you please make sure that the two things above (sorting the unique times when appropriate, and making the above more consistent) are filed as issues?

R/slide.R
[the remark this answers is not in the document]
kenmawer: I've already added an example with central alignment.

tests/testthat/test-epi_slide.R
kenmawer: I added issues to both.

R/outliers.R
```diff
@@ -179,7 +179,7 @@ detect_outlr_rm = function(x = seq_along(y), y, n = 21,
 
   # Calculate lower and upper thresholds and replacement value
   z = z %>%
-    epi_slide(fitted = median(y), n = n, align = "center") %>%
+    epi_slide(fitted = median(y), before = floor((n-1)/2), after = ceiling((n-1)/2)) %>%
```
dajmcdon, line 182: I think the documentation should explain what will happen with `n` here and who it relates to `before` and `after` in `epi_slide()`. Would it make more sense to add `before`/`after` as args to the function that default to these transformations?

R/slide.R
```diff
@@ -19,24 +20,28 @@
 #' @param ... Additional arguments to pass to the function or formula specified
 #'   via `f`. Alternatively, if `f` is missing, then the current argument is
 #'   interpreted as an expression for tidy evaluation. See details.  
-#' @param n Number of time steps to use in the running window. For example, if
-#'   `n = 7`, one time step is one day, and the alignment is "right", then to
-#'   produce a value on January 7 we apply the given function or formula to data
-#'   in between January 1 and 7.
+#' @param before A nonnegative integer specifying the number of time steps
+#'   before each of the `ref_time_values` to extract data from.
```
dajmcdon, line 24: I don't think "extract data from" is the right phrase. Maybe
> A nonnegative integer specifying the number of time steps before the `ref_time_value` to use in the running window.

R/slide.R
```diff
@@ -19,24 +20,28 @@
...
+#'   This must be a vector of length 1.
+#'   Set to 0 for a right-aligned/trailing sliding window, meaning
+#'   that no
+#'   `time_value` after the slide will be used for the sliding calculation.
+#'   It is mandatory to specify a `before` value, unless `after` is specified
+#'   as a non-zero value. In this case, `before` will be assumed to be 0, as it
+#'   assumes the user wants to do a left-aligned/leading sliding window.
+#'   However, this usage is discouraged and will thus produce a warning.
+#' @param after A nonnegative integer specifying the number of time steps after
+#'   each of the `ref_time_values` to extract data from.
```
dajmcdon, line 34: Same as previous comment.

And in at least one of these we should give an example with concrete dates the way there was previously.

R/slide.R
```diff
@@ -60,27 +65,32 @@
 #'   according to the `new_col_name` argument. 
 #' 
 #' @details To "slide" means to apply a function or formula over a running
-#'   window of `n` time steps, where the unit (the meaning of one time step) is
+#'   window of `before` time steps before and `after` time steps after,
```
dajmcdon, line 68: Clearer is:

> ...over a running window of time steps where the window is entered at a reference time and left and right endpoints are given by the `before` and `after` arguments.

R/slide.R
```diff
@@ -60,27 +65,32 @@
...
+#'   where the unit (the meaning of one time step) is
 #'   implicitly defined by the way the `time_value` column treats addition and
 #'   subtraction; for example, if the time values are coded as `Date` objects,
 #'   then one time step is one day, since `as.Date("2022-01-01") + 1` equals
 #'   `as.Date("2022-01-02")`. Alternatively, the time step can be set explicitly
 #'   using the `time_step` argument (which if specified would override the
-#'   default choice based on `time_value` column). If less than `n` time steps
-#'   are available at any given reference time value, then `epi_slide()` still
+#'   default choice based on `time_value` column). If certain time steps
```
dajmcdon, line 75: better:

> If there are not sufficient time steps available to complete the  window at any given reference time, then...

R/slide.R
```diff
@@ -95,29 +105,42 @@
...
-#'  group_by(geo_value) %>%
-#'  epi_slide(a = data.frame(cases_2dav = mean(cases), 
+#'   jhu_csse_daily_subset %>% 
+#'   group_by(geo_value) %>%
+#'   epi_slide(a = data.frame(cases_2dav = mean(cases), 
 #'                           cases_2dma = mad(cases)),
-#'            n = 2, as_list_col = TRUE)
-epi_slide = function(x, f, ..., n, ref_time_values,
-                     align = c("right", "center", "left"), before, time_step, 
+#'             before = 1, as_list_col = TRUE)
```
dajmcdon, line 138: I feel like `before` isn't special. I think better would make both = 0 by default. Then nothing is ever missing and we don't need the warning.

dajmcdon (changes requested):

R/slide.R
kenmawer: I was bringing this up to Logan, who disagreed and claimed that a person wouldn't want an "identity" map when neither `before` nor `after` is specified, despite its not producing bugs.

R/outliers.R
kenmawer: I asked Logan about this on Slack.

dajmcdon: And what was the answer? To be clear, I'm advocating either of the following:
1. In the documentation for `@param n` above, explain that if, for example, n=5, slide gets used under the hood with before=2 and after=2. Then link to `epi_slide()`.
2. Remove the n parameter and replace it with `before` / `after` directly.
If the decision was not to do 2, then you need to do 1.

dajmcdon: Oh, I see that you already did 1. Thanks!

R/slide.R
dajmcdon: I can't seem to find where this was discussed. @brookslogan if this is your preference, feel free to approve.

R/slide.R
```diff
@@ -6,7 +6,8 @@
 #'
 #' @param x The `epi_df` object under consideration.
 #' @param f Function or formula to slide over variables in `x`. To "slide" means
-#'   to apply a function or formula over a running window of `n` time steps
+#'   to apply a function or formula over a running window of `before`
```
dajmcdon, line 9: I'd rewrite this as:
> ...over a running window of time steps. The window is determined by the `before` and `after` parameters described below. One time step is typically one day or one week; see details for more explanation. ...

R/slide.R
```diff
@@ -19,24 +20,32 @@
...
 #'   via `f`. Alternatively, if `f` is missing, then the current argument is
 #'   interpreted as an expression for tidy evaluation. See details.  
-#' @param n Number of time steps to use in the running window. For example, if
-#'   `n = 7`, one time step is one day, and the alignment is "right", then to
-#'   produce a value on January 7 we apply the given function or formula to data
-#'   in between January 1 and 7.
+#' @param before A nonnegative integer specifying the number of time steps
+#'   before the `ref_time_value` to use in the running window.
+#'   This must be a vector of length 1.
+#'   Set to 0 for a right-aligned/trailing sliding window, meaning that no
```
dajmcdon, line 26: Somehow, this doesn't sound correct to me. If before=0 and after=5, then this is left-aligned and leading. Right?

dajmcdon (commented):
One minor bit of confusion, and one comment to be approved by @brookslogan .

R/slide.R
brookslogan: Issue for this is https://github.com/cmu-delphi/epiprocess/issues/161, but the discussion/decision is actually [here](https://github.com/cmu-delphi/epiprocess/issues/146#issuecomment-1192785302).  The idea was that trailing windows are the most common choice, so passing just `before` should be possible.  (Ken also made it possible to just pass `after`, so `before` actually isn't special; maybe we need to update the defaults that the user sees to make this clear.)  I think it's much more likely that passing no before&after is a bug than the user would want to have a 0-width window.

R/outliers.R
brookslogan: Debating on this.  Options:
- Explicate `@param n` in `detect_outlr_rm` and `@param n_threshold` in `detect_outlr_stl` that these are centered windows (+ maybe some more comments about what happens for even values).
- Change `n` -> `before`, `after` in `detect_outlr_rm`; just explain more about `@param n_threshold` in `detect_outlr_stl`.
- Change `n` and `n_threshold` to things involving `before` and `after`.

The later options add functionality and make naming more consistent with the updated slide, but also would introduce some naming mismatch in the stl one, which also has `n_trend`, `n_seasonal`.

dajmcdon: I'm happy with option 1. It's also the simplest.

elray1: Logan directed me to this conversation in a message on slack. I think 1 (in Dan's list of options above) is an OK answer, but I see two possible arguments for doing 2 instead:

1. This would provide an interface that's more consistent with the other functions in the package that are being rewritten to use before/after
2. It allows for more-consistent behavior at the end of the time series where a centered window is not feasible (and in practice, at the end of the time series slider will truncate the window so that it would effectively have `before=2` and `after=0` in Dan's example with `n=5`). A user might be concerned about that and prefer to specify a window with `before=...` and `after=0` to get more stable results.

R/slide.R
brookslogan: Think you're right, this should be left-aligned/leading, and in `@param after` setting after=0 is right-aligned/trailing.

Separately: explicitly passing 0 isn't even required; maybe this should be "Set to 0 or leave unset".

R/slide.R
brookslogan: +1, but suggest "running" -> "rolling".  (I think "running" would refer to an expanding window with the left side anchored at the beginning of the time series.)

R/slide.R
brookslogan: Sorry, I forgot what we actually settled on; `before` is special, and passing `after` without `before` is the warning case.

R/slide.R
kenmawer: You're right. I got them mixed up.

R/slide.R
brookslogan: This is a little complex and asymmetric, but use cases are also asymmetric.
- before = k, after = 0: very common.  Used in version-unaware forecasting, forecast covariate preparation, growth rate calculations, direction calculations, maybe outlier detection, maybe week-level target definitions (Saturdays as representative dates).
- before = 0, after = k: uncommon or common.  Used in growth rate and direction calculations, maybe week-level target definitions (Sundays as representative date).
- before = k, after = k: common.  Used for more representative time values for calculations (e.g., maybe forecast covariates), outlier detection, maybe week-level target definitions (Wednesdays as representative dates).
- before = 0, after = 0: extremely uncommon.  Can't think of a use case.  If `epi_slide` was more like summarize, maybe it could be used for geo aggregation, but I don't think it could be used that way now.

I think an error/warning on missing before && missing after should stay.  Decisions to be made:
- Do we keep a warning on `before` missing, `after` nonzero?
  - Sketching out the above makes me think not.  Leading windows aren't all that rare relative to centered windows, and this warning makes the interface more asymmetric.
- How do we express the defaults in the function signature?  E.g., assuming we take out the warning above:
  - `before, after`: user might be confused by `before`, `after`, `ref_time_values`, and `time_step` with no defaults. --- I slightly favor this option.
  - `before=0, after=0`: then user might be surprised by the error when they don't pass either, because it looks like they are already taken care of from the function signature.  (Plus manually passing 0s for both to address this would be even more mysterious.)
  - `before, after=0`: doesn't make sense if we have taken out the warning above
- Do we error/warn on `before` missing, user-provided `after=0`?
  - I suggest a warning here.  A user is either (a) new and trying to get a trailing window, and forgetting to set `before` or expecting some default (--> error/warning), or (b) tweaking parameters of an existing trailing window to see if a 0-width "window" is actually better than another value that they were trying (--> warning/nothing).
- Do we error/warn on `after` missing, user-provided `before=0`?
  - I suggest a warning here.

R/slide.R
kenmawer: Fixed.

R/outliers.R
brookslogan: Since Ken is trying to wrap up a bunch of PRs all at once this week, it probably makes sense just to do approach 1 + add an issue to revisit this later.  There is at least one other issue to consider regarding behavior at the end of the time series which I'd guess would be even more important: the fact that when data is trending, say, upward, the data at the end of the time series will be judged against recent data that is below it, VS. post-hoc we could see that it's in the middle of earlier data below it + later data above it.  Perhaps splitting `detect_outlr_rm`'s `n` -> `before`, `after` could be tackled at the same time, or maybe this will motivate some alternative approach that would be its own separate function.

brookslogan: I'm not so sure now that there are other things to do as I claimed; using `detect_outlr_rm` with `before=k` and `after=0` would likely involve tweaking parameters so that they worked for historical trending-upward/downward regions, and so they should also be appropriate for the end of the time series.  Still, since Ken already implemented approach 1, I'll just file this in another issue.

brookslogan:
> Remaining TODO on my end: double-check that we've stamped out all references to `n` and `align` in code, docs, tests, examples, vignettes.
I just did a cursory search (`rg` (ripgrep)-ing for "`n`").  It looks like the [intro paragraph](https://github.com/cmu-delphi/epiprocess/blob/7995dfe1ebebabc7d4bb809dc152c9a980df50d1/vignettes/slide.Rmd#L14) to `slide.Rmd` still needs updated.

R/slide.R
kenmawer: I will update this to take into consideration what you suggest.

R/outliers.R
```diff
@@ -128,6 +128,11 @@ detect_outlr = function(x = seq_along(y), y,
 #'   `y`).
 #' @param y Signal values.
 #' @param n Number of time steps to use in the rolling window. Default is 21.
+#'   This value is centrally aligned. When `n` is an odd number, the
+#'   rolling range goes between `(n-1)/2` `time_value`s before to `(n-1)/2`
+#'   `time_value`s after. When `n` is even, then  the
+#'   rolling range goes between `n/2-1` `time_value`s before to `n/2`
+#'   `time_value`s after.
```
brookslogan, line 135: If this is based on sliding, it's probably not quite accurate; slide is based on a sliding time window/range, which, if the `time_value`s are not evenly spaced, could contain a variable number of `time_value`s.  (Separate issue: I wish we had some better / less easily confused terminology to use, but even in slide docs, it might be a bit easy to confuse the two.)

brookslogan: I think the current way we would describe this is "time steps before" and "time steps after".  Not much better, but good for consistency.

brookslogan: Done.

brookslogan:
- [ ] TODO: check the pkg checks, implement idea from https://github.com/cmu-delphi/epiprocess/issues/177 before merging.

brookslogan:
Merging into `dev` as is.  Remaining items can be addressed separately there before release.

Issue #194, opened 2022-08-08
`Abort` or decay to tibble if `epi_df` column duplication or renaming invalidates key


brookslogan:
If a user duplicates one of the key columns or renames the `geo_value` or `time_value` column, we should either raise an error or decay to a tibble.  Related to, or intersects with, #192.  Taken from discussion on #185.

Marking as P3 as it seems low probability that this will cause confusion.  (It seems extremely unlikely that a user would directly duplicate key columns.  They might be more tempted to try renaming `geo_value` and `time_value`, but I think we also make it clear that we expect these fixed names, so they should probably suspect this would cause issues.  Libraries might try to do such duplication or renaming in the implementation details of some sensible operation, but I can't think of when this might actually happen.)

rachlobay:
Looking at the ex. from our discussion on https://github.com/cmu-delphi/epiprocess/pull/185, if we take `ex1[,c(1,1,2:4)]` and try to coerce that to a tibble using `as_tibble(ex1[,c(1,1,2:4)])`, this results in an error on my end:
```
! Column name `pol` must not be duplicated.
Use .name_repair to specify repair.
Caused by error in `repaired_names()`:
! Names must be unique.
```
Similarly if we try to duplicate other cols in that df. Hence, it may be best to abort for col. duplication.

Issue #193, opened 2022-08-08
When decaying from `epi_df` to `tibble`, remove metadata that doesn't make sense


brookslogan:
From discussion from #185.  Operations such as `[` can remove essential columns from an `epi_df`, so we decay from an `epi_df` to a `tibble`; however, the metadata sticks around, and some of it doesn't make sense.

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(epiprocess)
#> 
#> Attaching package: 'epiprocess'
#> The following object is masked from 'package:stats':
#> 
#>     filter
decayed =
  new_epi_df(tibble(geo_value="ak",time_value=1L,value=1L)) #%>%
  #select(-geo_value)
decayed[-1]
#> # A tibble: 1 × 2
#>   time_value value
#>        <int> <int>
#> 1          1     1
decayed[-1] %>% attr("metadata")
#> $geo_type
#> [1] "state"
#> 
#> $time_type
#> [1] "custom"
#> 
#> $as_of
#> [1] "2022-08-08 09:15:22 PDT"
```

<sup>Created on 2022-08-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Here, `geo_type` doesn't make sense.  We didn't have any `other_keys` to begin with, but if we did, they might not make sense either.  In other cases, different metadata will be invalidated.  We might want to just drop all the metadata in all cases to keep it simple.

Marking as P3 as the presence of metadata that doesn't make sense on a tibble shouldn't affect its operation, and if we convert back to an `epi_df` using `{new,as}_epi_df`, it looks like the only metadata that will be carried over is `as_of`, which might or might not make sense depending on what intermediate operations have occurred.  We should probably make it easier to override this `as_of` setting, though (intersects with https://github.com/cmu-delphi/epiprocess/issues/166).

rachlobay:
Modified the above ex. to explicitly use indexing to make sure goes through `[.epi_df`. As discussed, will go for the simplest soln to drop all metadata.

Issue #192, opened 2022-08-08
Adjust `other_keys` or decay to tibble when selecting, renaming columns


brookslogan:
Taken from discussion on #185.
``` r
## pak::pkg_install("cmu-delphi/epiprocess@9259796dafa159c8139f1f2eef619e1405c778ed")
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(epiprocess)
#> 
#> Attaching package: 'epiprocess'
#> The following object is masked from 'package:stats':
#> 
#>     filter
ex1_input <- tibble::tibble(
  geo_value = rep(c("ca", "fl", "pa"), each = 3),
  county_code = c("06059","06061","06067",
                  "12111","12113","12117",
                  "42101", "42103","42105"),
  time_value = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
                       by = "day"), length.out = length(geo_value)),
  value = 1:length(geo_value) + 0.01 * rnorm(length(geo_value))
) %>% 
  tsibble::as_tsibble(index = time_value, key = c(geo_value, county_code))
ex1 <- as_epi_df(x = ex1_input, geo_type = "state", time_type = "day", as_of = "2020-06-03")
# NOTE: cols are re-ordered so 3rd column is `county_code`
# ex1[-3] # (still an `epi_df`, good(? see unique key discussion))
ex1[-3] %>% attr("metadata") %>% .$other_keys
#> [1] "county_code"
# (bad)
```

<sup>Created on 2022-08-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Unlike `geo_value` and `time_value`, we can potentially drop a key column from `other_keys` and still get a sensible `epi_df`.  This `epi_df` should not have `other_keys` that refer to nonexistent columns, though; any key columns that have been selected out from the table data should also be removed from `other_keys`.  A more difficult question is what to do if removing this key column has taken us from a valid unique key to a nonunique key; we have other checks in `[` designed to decay to tibble in those cases, but are lax about enforcing a unique key elsewhere, including in `epi_df` construction.

Additionally, we should consider adding `names<-` and `dimnames<-` implementations in order to update `other_keys` metadata when `other_keys` are renamed, or decay to tibble when `{geo,time}_value` are renamed.

[...]

Issue #195, opened 2022-08-08
Create our own implementation for `rename()` 


rachlobay:
Want to create our own implementation for `rename()` such that if we rename a column corresponding to other_keys, we update the metadata accordingly (to have the updated name). For example, if we consider ex3 from [here](https://cmu-delphi.github.io/epiprocess/articles/epiprocess.html?q=additional%20keys#adding-additional-keys-to-an-epi_df-object), if the user performs `renamed_ex3 = ex3 %>%  dplyr::rename(Pol = pol)`, then the metadata for `renamed_ex3` should be updated to have Pol under `$other_keys` instead of pol. Also, if the user tries to change the name of either `geo_value` or `time_value`, should probably throw an error as those two are important to leave named as they are.

brookslogan:
Related: #223.  `?dplyr_extending` might provide a way to limit the number of S3 implementations we have to provide.

PR #196, opened 2022-08-08
Addressed issues 192-194

rachlobay:
Since Issues https://github.com/cmu-delphi/epiprocess/issues/192, https://github.com/cmu-delphi/epiprocess/issues/193, and https://github.com/cmu-delphi/epiprocess/issues/194 are closely related, addressed all three here.

After discussion with Logan, made the following changes to `[.epi_df` code for the written issue:
- Issue https://github.com/cmu-delphi/epiprocess/issues/192: Any key columns that have been selected out from an `epi_df` are removed from other_keys in the metadata (refer to the ex in that issue).
- Issue https://github.com/cmu-delphi/epiprocess/issues/193: Dropped all metadata when decay to tibble.
- Issue https://github.com/cmu-delphi/epiprocess/issues/194: Abort if column duplication in an `epi_df` (to be consistent with tibble behaviour).

### DIFF
--- a/R/methods-epi_df.R
+++ b/R/methods-epi_df.R
@@ -81,7 +81,7 @@ summary.epi_df = function(object, ...) {
   
   cn <- names(res)
   
-  # Duplicate key columns, Abort
+  # Duplicate columns, Abort
   dup_col_names = cn[duplicated(cn)]
   if (length(dup_col_names) != 0) {
     Abort(paste0("Column name(s) ", 
--- a/R/methods-epi_df.R
+++ b/R/methods-epi_df.R
@@ -79,7 +79,7 @@ summary.epi_df = function(object, ...) {
   
   cn <- names(res)
   
-  # Duplicate key columns, Abort
+  # Duplicate columns, Abort
   dup_col_names = cn[duplicated(cn)]
   if (length(dup_col_names) != 0) {
     Abort(paste0("Column name(s) ", 
--- a/R/methods-epi_df.R
+++ b/R/methods-epi_df.R
@@ -89,6 +89,7 @@ summary.epi_df = function(object, ...) {
                        collapse = ", "), " must not be duplicated."))
   } 
   
+  # If not an epi_df (missing time_value or geo_value), return a tibble
   not_epi_df <- (!("time_value" %in% cn) || !("geo_value" %in% cn))
   
   if (not_epi_df) {
--- a/R/methods-epi_df.R
+++ b/R/methods-epi_df.R
@@ -106,7 +106,7 @@ summary.epi_df = function(object, ...) {
   only_in_orig = setdiff(unlist(attr(x, "metadata")$other_keys), 
                          cn[!(cn %in% c("geo_value", "time_value"))]) 
   
-  if (length(only_orig) != 0) {
+  if (length(only_in_orig) != 0) {
     # remove all other_keys elements in the original epi_df that do not occur in subset
     attr(res, "metadata")$other_keys = 
       attr(res, "metadata")$other_keys[! attr(res, "metadata")$other_keys %in% only_in_orig]
--- a/R/methods-epi_df.R
+++ b/R/methods-epi_df.R
@@ -80,13 +80,38 @@ summary.epi_df = function(object, ...) {
   } 
   
   cn <- names(res)
-  nr <- vctrs::vec_size(x) 
-  not_epi_df <- (!("time_value" %in% cn) || !("geo_value" %in% cn) || vctrs::vec_size(res) > nr || any(i > nr))
   
-  if (not_epi_df) return(tibble::as_tibble(res))
+  # Duplicate key columns, Abort
+  dup_col_names = cn[duplicated(cn)]
+  if (length(dup_col_names) != 0) {
+    Abort(paste0("Column name(s) ", 
+                 paste(unique(dup_col_names), 
+                       collapse = ", "), " must not be duplicated."))
+  } 
   
-  # Use reclass as safeguard (in case class & metadata are dropped)
+  not_epi_df <- (!("time_value" %in% cn) || !("geo_value" %in% cn))
+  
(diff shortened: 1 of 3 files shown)


R/methods-epi_df.R
```diff
@@ -80,13 +80,39 @@ summary.epi_df = function(object, ...) {
   }
```
brookslogan, line 80: Looks like `i_arg` and `j_arg` aren't used anymore; we can remove them entirely

R/methods-epi_df.R
```diff
@@ -80,13 +80,39 @@ summary.epi_df = function(object, ...) {
...
+  # Duplicate columns, Abort
+  dup_col_names = cn[duplicated(cn)]
+  if (length(dup_col_names) != 0) {
+    Abort(paste0("Column name(s) ", 
+                 paste(unique(dup_col_names), 
+                       collapse = ", "), " must not be duplicated."))
+  } 
+  
+  # If not an epi_df (missing time_value or geo_value), return a tibble
+  not_epi_df <- (!("time_value" %in% cn) || !("geo_value" %in% cn))
```
brookslogan, line 93: Style thing: remove outer parens?  (Inner parens may also be unnecessary, but still might help clarify.)

R/methods-epi_df.R
```diff
@@ -80,13 +80,39 @@ summary.epi_df = function(object, ...) {
...
   
-  # Use reclass as safeguard (in case class & metadata are dropped)
+  if (not_epi_df) {
+    attributes(res)$metadata <- NULL 
+    return(tibble::as_tibble(res))
+  }
+  
+  # Use reclass as safeguard (in case class &/or metadata are dropped)
   reclass(res, attr(x, "metadata"))
```
brookslogan, line 101: I don't think this will actually alter `res`; you probably need `res <- reclass(res, attr(x, "metadata"))`

R/methods-epi_df.R
```diff
@@ -80,13 +80,39 @@ summary.epi_df = function(object, ...) {
...
+  
+  # Extract the other_keys colnames in the original epi_df that do not occur in the subset
+  only_in_orig = setdiff(unlist(attr(x, "metadata")$other_keys), 
+                         cn[!(cn %in% c("geo_value", "time_value"))]) 
+  
+  if (length(only_in_orig) != 0) {
+    # remove all other_keys elements in the original epi_df that do not occur in subset
+    attr(res, "metadata")$other_keys = 
+      attr(res, "metadata")$other_keys[! attr(res, "metadata")$other_keys %in% only_in_orig]
+  }
```
brookslogan, line 113: I think (since we're not messaging about this inside the `if` statement) this section might be able to be written shorter as
```
attr(res, "metadata")$other_keys <- intersect(attr(res, "metadata")$other_keys, cn)
```
and I believe this preserves the original order of the keys that stick around, but `?intersect` doesn't seem to guarantee this.  So if we want to guarantee this behavior, maybe
```
attr(res, "metadata")$other_keys <- attr(res, "metadata")$other_keys[attr(res, "metadata")$other_keys %in% cn]
```
plus a comment about why we're not using `intersect`?

R/methods-epi_df.R
brookslogan: Could you please add some test_that tests for the behaviors above?  I assume at least the duplicate-col error must be missing from the tests.

brookslogan (changes requested):
Couple a minor things + a request to add any relevant tests for the new checks.  Are all the rename-related things already split out into other issues?

brookslogan:
Also, misc thing: it looks like git might be using attaching your computer username to the commits here, which might be a bit confusing looking at git logs.  Could you please try running either:
(1)
```sh
git config user.name "<Your Name>"
git config user.email "<email-addr>@<.....>.edu"
```
in each cloned repo, or
(2)
```sh
git config --global user.name "<Your Name>"
git config --global user.email "<your-email-address>@<.....>.edu"
```
from anywhere?

You can check for success using
```
git config --list
```
in each repo, or
```
git config --global --list
```
to view just the global settings.

R/methods-epi_df.R
rachlobay: Added tests for duplicate-col error and missing `geo_value` col.

rachlobay:
Since the rename issues that were mentioned in https://github.com/cmu-delphi/epiprocess/issues/192 and https://github.com/cmu-delphi/epiprocess/issues/194 were related, I put them into one issue [here](https://github.com/cmu-delphi/epiprocess/issues/195), where I added a little more info based on our last discussion.

R/methods-epi_df.R
rachlobay: After looking into `intersect`, it looks like the order of the resulting elements is the same as the order of the elements in the 1st argument. So, we can use that. Also, `cn = names(res)`... so I think what you meant for the 1st argument in `intersect` is is `attr(x, "metadata")$other_keys` (b/c one of the args should be that)?

dajmcdon (approved):
LGTM once all @brookslogan comments are addressed.

R/methods-epi_df.R
brookslogan: Right, `intersect` in my version of R also appears to keep the first arg's order; however, I can't find anywhere in the docs that guarantees that this will be the case for any version of R.  That's why I was still looking at that `%in%` solution; in case for some reason they decide to change `intersect`'s implementation within the bounds of the docs.

I think I wrote it assuming that `attr(res, "metadata")` was set to `attr(x, "metadata")`; if that's not the case, then yes, sounds like something should be changed to `x` in there.

brookslogan: Pushing an update to address this wild fear + to simplify a little bit (we don't even need the if statement!)

brookslogan (approved):
Addressed a minor check comment.  If checks pass, think this is ready to merge.

PR #197, opened 2022-08-10
Test growth_rate

kenmawer:
Note that I was unable to prompt this warning message:

`Warn("`x` contains duplicate values. (If being run on a column in an `epi_df`, did you group by relevant key variables?)")`


brookslogan:
Think you can trigger that warning + some error with
```r
jhu_csse_daily_subset %>% 
  ## group_by(geo_value) %>% 
  mutate(cases_gr = growth_rate(x = time_value,  y = cases, dup_rm=TRUE))
```
but without `dup_rm=TRUE`, it actually returns a result.... is this a bug?  Do you know what the result would represent in this case?

tests/testthat/test-growth_rate.R
```diff
@@ -0,0 +1,14 @@
+library(dplyr)
+
+x <- jhu_csse_daily_subset %>%
+  select(-case_rate_7d_av,-death_rate_7d_av, -cases_7d_av) %>%
+  group_by(geo_value) %>% 
+  mutate(cases_gr = growth_rate(x = time_value,  y = cases))
```
brookslogan, line 6: Looks like this `x` isn't used, and might cause confusion win the `x` columns below.  Maybe remove this statement, or put it inside an `expect_error(jhu_csse_daily_subset %>% <...>, NA)`?

tests/testthat/test-growth_rate.R
```diff
@@ -0,0 +1,14 @@
...
+
+test_that("Test error throwing",{
+  expect_error(growth_rate(x=1:3,y=1:4),
+               "`x` and `y` must have the same length.")
+  expect_error(growth_rate(x=1:20,y=1:20,x0=21),
+               "`x0` must be a subset of `x`.")
```
brookslogan, line 12: Please place a comment here (between the error-expected and no-error-expected tests) explaining that the statements below actually test that there is no error.  (Or, if you think it's clearer, you don't need `expect_error(<stuff>, NA)` at all; you could just write `<stuff>` and I believe testthat has roughly similar behavior.)

tests/testthat/test-growth_rate.R
```diff
@@ -0,0 +1,14 @@
...
+  expect_error(growth_rate(x=1:20,y=1:20,x0=c(1,3)),NA)
+})
```
brookslogan, line 14: Could you please add another test for the actual values coming out?  E.g., construct a simple example where the growth rate should be obvious, and test that we calculate that obvious value.  At least for some of the `method`s.

Also, please test that each of the `method`s, `log_scale` settings, and `na_rm` settings at least run successfully.  If you can't construct an example where you think you know what the output should be, maybe throw in some `testthat::expect_snapshot`s?

brookslogan: (I assume we'll need to take the `expect_snapshot` approach for some/most/all of the `method`s; I'm not sure, e.g., that there's a simple example where we know the trendfilter output.)

kenmawer: `expect_snapshot` creates a new file for snapshots and does something that doesn't seem inherently helpful. All it does it checks for errors from my experience; however, that doesn't seem to be any better than doing `expect_error`.

tests/testthat/test-growth_rate.R
kenmawer: I've just had a habit of doing `expect_error(<stuff>, NA)` as having `<stuff>` makes it unclear if it's just random code I forgot to delete or actual testing.

R/growth_rate.R
```diff
@@ -44,7 +44,8 @@
 #' * "rel_change": uses (B/A - 1) / h, where B is the average of `y` over the
 #'   second half of a sliding window of bandwidth h centered at the reference
 #'   point `x0`, and A the average over the first half. This can be seen as
-#'   using a first-difference approximation to the derivative.
+#'   using a first-difference approximation to the derivative. This is the
+#'   default method if `method` is not specified.
```
dajmcdon, line 48: let's say: "This is the default if `method` is not specified."

R/growth_rate.R
```diff
@@ -145,7 +146,7 @@ growth_rate = function(x = seq_along(y), y, x0 = x,
   
   # Remove NAs if we need to
   if (na_rm) {
-    o = !(is.na(x) & is.na(y))
+    o = !(is.na(x) | is.na(y))
```
dajmcdon, line 149: Just for clarity:

```r
o = !is.na(x) & !is.na(y)
```

tests/testthat/test-growth_rate.R
dajmcdon: @kenmawer I think the point here is to test that the correct calculations are occurring when arguments are correctly specified, not just that errors happen when the inputs are wrong.

dajmcdon (changes requested):

tests/testthat/test-growth_rate.R
kenmawer: I already added a few tests for calculations. I would suspect that for some of the other methods, it would be too complicated to be able to verify the outputted numbers. It also depends on what you mean by "correct calculations," whether it's merely getting the right number of cells or getting those exact numbers.

brookslogan: Since it's likely too complicated to verify the output numbers, that's why I suggested `expect_snapshot`.  When you first run it, it saves the current output to a _snaps directory, which you would then commit (make sure that the test is constructed so that this file is very small; we don't want to bloat the repo), and then later, when we run the snapshot test again, it will detect the file in the _snaps directory and compare against it.  If we inadvertently change the output behavior, this test will catch it.

tests/testthat/test-growth_rate.R
brookslogan: +1 plus this also (I think) tallies it as its own test, which makes sense.  I still think an explanatory comment or putting the NA on a separate line is needed to more prominently highlight that this is non-error behavior.

nmdefries: I don't see any tests here that are expected to NOT error, so I'm assuming this comment has been resolved/non-error tests have been moved elsewhere.

tests/testthat/test-growth_rate.R
```diff
@@ -0,0 +1,56 @@
...
+methods <- c("rel_change","linear_reg","smooth_spline","trend_filter")
+
+gr <- function(method = "rel_change", h = 3, na_rm = TRUE, ...) {
+  growth_rate(x=X,y=Y,method=method,na_rm = na_rm, h = h,...)
+}
+
+test_that("Test error throwing",{
+  # Error cases
+  expect_error(growth_rate(x=1:3,y=1:4),
+               "`x` and `y` must have the same length.")
```
nmdefries, line 15: suggestion [non-blocking]: Probably beyond scope here, but it would be nice if the errors and warnings in `growth_rate` had classes so we could check for that rather than message wording, which is liable to change.

tests/testthat/test-growth_rate.R
```diff
@@ -0,0 +1,56 @@
...
+
+test_that("Test throwing of warning of duplicates",{
+  # The warning that is prompted is as follows:
+  # "`x` contains duplicate values. (If being run on a column in an `epi_df`,
+  # did you group by relevant key variables?)"
+  # Note that putting it in the regexp of expect_warning doesn't seem to work
+  jhu_csse_daily_subset %>%
+    mutate(cases_gr = growth_rate(x = time_value,  y = cases, dup_rm=TRUE)) %>%
+    expect_warning() %>%
+    expect_error()
```
nmdefries, line 31: issue: It's hard to tell what this is doing. If we want to check for both a warning and an error, run this twice with each `expect_*`. If we're trying to do something else, I don't understand.

tests/testthat/test-growth_rate.R
```diff
@@ -0,0 +1,56 @@
...
+})
+
+test_that("log_scale works",{
+  expect_equal(growth_rate(x=1:20,y=exp(1:20),h=5,
+                           method="linear_reg",log_scale = TRUE),
+               rep(1,20))
+})
+
+test_that("Running different methods with NA removal won't fail",{
+  for (m in methods) expect_false(NA %in% gr(method = m,x0=1:5))
```
nmdefries, line 45: question: `x0` doesn't contain any `NA`s, so are we really testing `NA`-removal? Do we want

```suggestion
  for (m in methods) expect_false(NA %in% gr(method = m,x0=c(1:5, NA))) 
```
?

tests/testthat/test-growth_rate.R
nmdefries: Looks like of the requested

> test that each of the methods, log_scale settings, and na_rm settings at least run successfully

only the method outputs are left -- is that correct?

nmdefries (changes requested):
Some questions about desired behavior and some about the state of previous review comments.

dajmcdon:
@nmdefries I'm not sure whether we shouldn't just close this at this point. Maybe investigate if the possible bug fix is needed, but close without merging?

nmdefries:
@dajmcdon Thanks for the feedback. I was tagged recently so not sure what the context is here.

@brookslogan Thoughts?

Issue #210, opened 2022-08-17
Dependency genlasso was removed from CRAN?


rachlobay:
It looks like the dependency genlasso was [very recently removed from CRAN](https://cran.r-project.org/web/packages/genlasso/index.html). This is currently a problem when the Github R CMD Check runs over in epipredict as I am now getting the following error:
```
 Error: <callr_remote_error: Cannot install packages:
 * deps::.: Can't install dependency cmu-delphi/epiprocess
 * cmu-delphi/epiprocess: Can't install dependency genlasso
 * genlasso: Can't find package called genlasso.>
  in process 8347
```
See [here](https://github.com/cmu-delphi/epipredict/runs/7869012059?check_suite_focus=true) for the context of this error.

PR #211, opened 2022-08-17
Use glmgen/genlasso because package was removed from CRAN

rachlobay:
To resolve Issue #210, switched to the remote repo: https://github.com/glmgen/genlasso because genlasso was removed from CRAN.


XuedaShen:
Had the same problem and switched to remote repo as well.

dajmcdon (approved):

Issue #208, opened 2022-08-16
`epix_slide()` converts to a `tibble` rather than an `epi_df` - incompatible with `arx_forecaster()`


rachlobay:
In the slide function for an `epi_archive` object, there is a call to convert to a tibble when performing the computation by group:
```
if (!missing(f)) {
  if (rlang::is_formula(f)) f = rlang::as_function(f)
  
  x = purrr::map_dfr(ref_time_values, function(t) {
    self$as_of(t, min_time_value = t - before_num) %>%
      tibble::as_tibble() %>%  # problem line
      dplyr::group_by(!!!group_by) %>%
      dplyr::group_modify(comp_one_grp,
                          f = f, ..., 
                          time_value = t,
                          key_vars = key_vars,
                          new_col = new_col,
                          .keep = TRUE) %>%
      dplyr::ungroup()
  })
}
```
So, when a function is used the output from that slide function is a tibble, but the main forecasting function `epi_forecaster()` from epipredict requires an `epi_df` to be inputted else it throws an error. Below is an example where we use the latest version of epipredict on GitHub:
```
library(epipredict)
library(epiprocess)
library(covidcast)
library(data.table)
library(dplyr)
library(tidyr)
library(ggplot2)

y <- covidcast_signals(
  c("doctor-visits", "jhu-csse"),
  c("smoothed_adj_cli", "confirmed_7dav_incidence_prop"),
  start_day = "2020-06-01",
  end_day = "2021-12-01",
  issues = c("2020-06-01", "2021-12-01"),
  geo_type = "state",
  geo_values = c("ca", "fl"))

z <- y[[1]] %>%
  select(geo_value, time_value, version = issue, percent_cli = value) %>%
  as_epi_archive()

z <- epix_merge(
  z, y[[2]] %>%
    select(geo_value, time_value, version = issue, case_rate = value) %>%
    as_epi_archive(), sync = "locf")

fc_time_values <- seq(as.Date("2020-08-01"), as.Date("2021-12-01"),
                      by = "1 month")
ahead = 7

# Old arx_forecaster works fine on the archive
z %>% epix_slide(fc = arx_forecaster(
  percent_cli, case_rate, geo_value, time_value,
  args = arx_args_list(ahead = ahead)),
  n = 120, ref_time_values = fc_time_values)

# New arx does not work on the archive: Error: epi_data must be an `epi_df`.
z %>%
  epix_slide(function(x, ...)
[...]

PR #212, opened 2022-08-17
Removed as_tibble() calls in archive slide fun

rachlobay:
To resolve Issue #208 removed the two `as_tibble()` calls in archive slide function so that `epix_slide()` is compatible with `arx_forecaster()` from epipredict (to be able to slide an ARX forecaster over an archive). Making a separate issue to deal with the metadata.


dajmcdon (approved):
@brookslogan This doesn't seem to cause downstream issues. Ok with you?

rachlobay:
Also checked that these changes don't unexpectedly break epipredict (at least locally).

Issue #213, opened 2022-08-18
Carefully consider the metadata for the `epi_df` resulting from sliding a forecaster over an `epi_archive` 


rachlobay:
As originally stated in Issue #208, after making the change to allow `epix_slide` to `epi_df` so that we can slide a forecaster using `arx_forecaster()` over an `epi_archive`, it is good to consider whether the resulting metadata (namely, `as_of`) is what we want.

In the below example, `as_of` is related to the first `fc_time_values` 2020-08-01… but do we really want that? Does it make the most sense to use here?

```
library(epipredict)
library(epiprocess)
library(covidcast)
library(data.table)
library(dplyr)
library(tidyr)
library(ggplot2)

y <- covidcast_signals(
  c("doctor-visits", "jhu-csse"),
  c("smoothed_adj_cli", "confirmed_7dav_incidence_prop"),
  start_day = "2020-06-01",
  end_day = "2021-12-01",
  issues = c("2020-06-01", "2021-12-01"),
  geo_type = "state",
  geo_values = c("ca", "fl"))

z <- y[[1]] %>%
  select(geo_value, time_value, version = issue, percent_cli = value) %>%
  as_epi_archive()

z <- epix_merge(
  z, y[[2]] %>%
    select(geo_value, time_value, version = issue, case_rate = value) %>%
    as_epi_archive(), sync = "locf")

fc_time_values <- seq(as.Date("2020-08-01"), as.Date("2021-12-01"),
                      by = "1 month")
ahead = 7

# Slide arx_forecaster over archive z
z %>% 
  epix_slide(function(x, ...) 
    arx_forecaster(x, outcome = "case_rate", 
                   predictors = "case_rate",
                   args_list = arx_args_list(ahead = ahead))$predictions %>% 
      select(-c(geo_value, time_value)),
    n = 120, ref_time_values = fc_time_values, new_col_name = "fc") 
```

brookslogan:
`as_of` should probably be set to `z$versions_end`.  (This is the last version we "observed", and basically acts like the archive's "version".  In most cases this is equivalent to `max(z$DT$version)`; when it's different, it will be greater than this max value and express that we observed additional versions that contained no updates.)

brookslogan:
If we are talking about the input to the group computations, then yes, it should be what we already get out of `$as_of`; the corresponding ref time value / fc time value / version.

brookslogan:
[...]

Issue #153, opened 2022-07-21
 `epix_slide` won't grab the latest version


kenmawer:
Related to #146, but the `ref_time_values`, which should actually be `ref_versions`, does not take in the latest version for the slide.

brookslogan:
From call: we should just fix the latest version not being included here, and leave the renaming for a separate discussion.  #146 contained some ideas for setting a default.

brookslogan:
Closed by #311.

PR #216, opened 2022-08-19
Replace `n` with `before` (no `after`) in `epix_slide`

kenmawer:
I somehow managed to break the other branch, so I put this in. This does not include random stuff from `epi_slide`.


R/methods-epi_archive.R
```diff
@@ -422,11 +422,11 @@ epix_merge = function(x, y,
 #' Finally, this is simply a wrapper around the `slide()` method of the
 #'   `epi_archive` class, so if `x` is an `epi_archive` object, then:
 #'   ```
-#'   epix_slide(x, new_var = comp(old_var), n = 120)
+#'   epix_slide(x, new_var = comp(old_var), before = 120)
```
brookslogan, line 425: Should this be `before = 119`?

R/archive.R
```diff
@@ -599,8 +599,8 @@ epi_archive =
             }
               
             # If a custom time step is specified, then redefine units 
-            before_num = n-1
-            if (!missing(time_step)) before_num = time_step(n-1)
+            before_num = before-1
```
brookslogan, line 602: Should this be `before_num = before`?

R/methods-epi_archive.R
brookslogan: (same question regarding all the calls to epix_slide, including other instances of n = 120, instances of n = 3, any others)

tests/testthat/test-epix_slide.R
```diff
@@ -0,0 +1,40 @@
...
+    as_epi_archive()
+  
+  xx1 <- epix_slide(x = xx,
+                    f = ~ sum(.x$binary),
+                    before = 3,
+                    group_by = geo_value,
+                    new_col_name = "sum_binary")
+  
+  xx2 <- tibble(geo_value = rep("x",2),
+                time_value = c(5,6),
```
brookslogan, line 27: Please add a comment here that the exclusion of `time_value` 7 is a bug (#153).

tests/testthat/test-epix_slide.R
```diff
@@ -0,0 +1,40 @@
+library(dplyr)
+library(rlang)
```
brookslogan, line 2: could we remove the `library(rlang)`?

tests/testthat/test-epix_slide.R
```diff
@@ -0,0 +1,40 @@
...
+test_that("epix_slide works as intended",{
+  x <- tibble::tribble(~version, ~time_value,
+                       5, c(1:2,4),
+                       6, c(1:2,4:5),
+                       7, 2:6) %>%
+    tidyr::unnest(time_value)
+  
+  xx <- bind_cols(geo_value = rep("x",12),
+                  arrange(x,time_value,version),
+                  binary = 2^(1:12)) %>%
```
brookslogan, line 17: (This example would be easier to read without running if the binary was assigned before the arrange (and then the arrange would not be needed), so we could say, for ref time 5, we see time values `c(1:2,4)`, with `binary` values `2^(1:3)`, which are filtered down to time values >= 3, so we end up only with the time_value 4 row with binary value 2^3.)

tests/testthat/test-epix_slide.R
```diff
@@ -0,0 +1,40 @@
...
+                    f = ~ sum(.x$binary),
+                    before = 3,
+                    group_by = geo_value,
+                    new_col_name = "sum_binary")
+  
+  xx2 <- tibble(geo_value = rep("x",2),
+                time_value = c(5,6),
+                sum_binary = c(2^7,
+                               2^10+2^8)) %>%
+    as_epi_df(as_of = 1)
```
brookslogan, line 30: Please also note that the slide producing something with an `as_of` of 1 is a bug (#213).

vignettes/compactify.Rmd
```diff
@@ -102,7 +102,7 @@ speeds <- rbind(speeds, speed_test(iterate_as_of,"as_of_1000x"))
 
 # Performance of slide
 slide_median <- function(my_ea) {
-    my_ea$slide(median = median(case_rate_7d_av), n = 7)
+    my_ea$slide(median = median(case_rate_7d_av), before = 7)
```
brookslogan, line 105: Unlike the above, where, after changing `before_num` to be `= before`, we would want to adjust the `before` arguments to be 120->119, 3->2, and 2->1, we probably want to keep before=7 here, because for jhu-csse cases, we typically expect time values up to v-1 in each version v, so before=7 would have 7 time values and be averaging across a week.  (Of course, it doesn't make too much sense to be 7-day-averaging a 7-day-average, but that might be a tweak for a separate issue.)

R/archive.R
kenmawer: I remember discussing this with someone, and that this is not a bug despite my thinking it is. Thus, please confirm it is a bug. For example, if `before = 1` with one-day latency, we get nothing with the slide.

R/methods-epi_archive.R
```diff
@@ -358,10 +358,10 @@ epix_merge = function(x, y,
 #' @param ... Additional arguments to pass to the function or formula specified
 #'   via `f`. Alternatively, if `f` is missing, then the current argument is
 #'   interpreted as an expression for tidy evaluation.
-#' @param n Number of time steps to use in the running window. For example, if
-#'   `n = 7`, and one time step is one day, then to produce a value on January 7
-#'   we apply the given function or formula to data in between January 1 and
-#'   7.
+#' @param before Number of time steps to use in the running window. For example,
```
brookslogan, line 361: TODO edit a little

R/methods-epi_archive.R
```diff
@@ -358,10 +358,10 @@ epix_merge = function(x, y,
...
+#'   if `before = 7`, and one time step is one day, then to produce a value on
```
brookslogan, line 362: also around here

R/methods-epi_archive.R
```diff
@@ -358,10 +358,10 @@ epix_merge = function(x, y,
...
+#'   January 7 we apply the given function or formula to data in between January
+#'   1 and 7.
```
brookslogan, line 364: double check here

R/methods-epi_archive.R
brookslogan: Think this + the below has been changed.

R/archive.R
brookslogan: As far as I know, switching to `before_num = before` is the desired behavior.

brookslogan:
Other TODOs
- [ ] Update `$slide` as well.

brookslogan:
- [ ] Decide whether/how to fix "#'   1. `epix_slide()` uses windows that are **always right-aligned** (in"

R/methods-epi_archive.R
brookslogan: Done.

R/methods-epi_archive.R
brookslogan: Done.

brookslogan:
Remaining TODOs: 
- [ ] Check the pkg checks
- [ ] Implement idea from https://github.com/cmu-delphi/epiprocess/issues/177 before merging.

brookslogan:
Merging into `dev` as is.  Remaining items can be addressed separately before creating a new release there.

PR #222, opened 2022-08-30
Add changelog (NEWS.md), associated pkgdown styling

brookslogan:
Part of #177.

Decisions to be made:
- Is this version breakdown appropriate?
- Is this changelog format appropriate?  What changes do we want to make now?  And moving forward?
- Should `main` continue to be a development branch?  If so, we can just merge this into main.  If not, we should PR the latest release into main so it gets the right version number[, or] just bump the version number in this PR [and release/tag it] (--> 0.5.1).

Dan & Dmitry, your input would be very helpful here.

I've prepared some branches like 0.1.0_prep that we could then tag as 0.1.0 if we're happy with the retro version numbering.  We should probably update the changelog inside of PRs moving forward.

(Note: we are trying to follow semver, but the .9999 thing wouldn't follow.  It's to mark it as a dev version, not a release version.  Seen in other R packages.  The idea is to, at release time, look at the changes/changelog and decide what the semver release version number should be.  Murky on what people actually recommend here.)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -183,7 +183,9 @@
   named parameters that did not receive values from `epi_slide`'s `...`.
 * Added some examples throughout the package.
 * Using example data objects in vignettes also speeds up vignette compilation.
-* Cleanup:
+
+## Cleanup:
+
 * Set up gh-actions CI.
 * Added tests for `epi_df`s.
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,8 +1,8 @@
+# epiprocess 0.5.0.9999 (development version)
+
 Note that `epiprocess` uses the [Semantic Versioning
 ("semver")](https://semver.org/) scheme for all release versions, but not for
 development versions. A ".9999" suffix indicates a development version.
-
-# epiprocess 0.5.0.9999 (development version)
 
 ## Cleanup:
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,8 @@
-# epiprocess 0.5.0.9999 (dev version)
+Note that `epiprocess` uses the [Semantic Versioning
+("semver")](https://semver.org/) scheme for all release versions, but not for
+development versions. A ".9999" suffix indicates a development version.
+
+# epiprocess 0.5.0.9999 (development version)
 
 ## Cleanup:
 
@@ -128,7 +132,8 @@
 
 * Removed dependency of some `epi_archive` tests on an example archive.
   object, and made them more understandable by reading without running.
-* Fixed `epi_df` tests relying on S3 implementation external to `epiprocess`.
+* Fixed `epi_df` tests relying on an S3 method for `epi_df` implemented
+  externally to `epiprocess`.
 * Added tests for `epi_archive` methods and wrapper functions.
 * Removed some dead code.
 * Made `.{Rbuild,git}ignore` files more comprehensive.
@@ -148,10 +153,10 @@
   characters to yield `"custom"`, not US `"nation"`.
 * Fixed `time_type` guessing to actually detect `Date`-class `time_value`s
   regularly spaced 7 days apart as `"week"`-type as intended.
-* Improved printing of `epi_df`s, `epi_archives`s
+* Improved printing of `epi_df`s, `epi_archives`s.
 * Fixed `as_of` to not cut off any (forecast-like) data with `time_value >
   max_version`.
-* Expanded `epi_df` docs include conversion from `tsibble`/`tbl_ts` objects,
+* Expanded `epi_df` docs to include conversion from `tsibble`/`tbl_ts` objects,
   usage of `other_keys`, and pre-processing objects not following the
   `geo_value`, `time_value` naming scheme.
 * Expanded `epi_slide` examples to show how to use an `f` argument with
@@ -213,7 +218,8 @@ Classes:
   * Convenience function: `is_epi_df`
 * `epi_archive`: R6 class for version (patch) data for geotemporal
   epidemiological time series data sets. Comes with S3 methods and regular
-  functions wrappers for those unfamiliar with R6 methods. Associated functions:
+  functions that wrap around this functionality for those unfamiliar with R6
+  methods. Associated functions:
   * `as_epi_archive`: prepares an `epi_archive` object from a data frame
     containing snapshots and/or patch data for every available version of
     the data set.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,0 +1,241 @@
+# epiprocess 0.5.0.9999 (dev version)
+
+## Cleanup:
+
+* Added a `NEWS.md` file to track changes to the package.
+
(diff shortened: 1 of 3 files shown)


brookslogan:
- On 1: do you mean version breakdown is okay, or we should have fewer versions?
- On 2: improvements vs. cleanup: improvements "impact" end-user; cleanup doesn't "impact" end-user.  For a fuzzy definition of "impact"; e.g., I guess dealing with dependencies being kicked off CRAN does impact user installation.  This could be clearer, but I'm not sure exactly how to draw the line/lines.  Any ideas?
- On 3&4: yes, seems that tagging each x.x.(x+1) update as a release and using main==dev and 4th digit for post-release development works.  An alternative might be a dev branch with/without a 4th digit.  There may not be much of a distinction between the two approaches if we are marking the dev branch as default PR destination but it ALSO as a consequence gets marked as the default `install_github` branch.

NEWS.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Improvements:
+
+* Improved `epi_slide` error messaging.
+* Fixed description of the appropriate parameters for an `f` argument to
+  `epi_slide`; previous description would give incorrect behavior if `f` had
+  named parameters that did not receive values from `epi_slide`'s `...`.
+* Added some examples throughout the package.
+* Using example data objects in vignettes also speeds up vignette compilation.
+* Cleanup:
```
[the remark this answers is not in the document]
brookslogan: Thanks for spotting!  Messed up converting this from a pure bulleted list.

brookslogan:
The main distinction is if epiprocess development is pushing directly to whatever branch/version epipredict is relying on.
- Same --> more breakage, faster feedback, trouble for users&pkgdown trying to test out during these times
- Different --> no spontaneous breakage, but need to periodically sync epipredict with later epiprocess versions and address breakage then

For the still-still-pending PRs and issues with breaking changes that we're trying to get through soon in epiprocess, we should at least partially follow approach 2.  But once they're all ready on some non-default branch, we'll be faced with a similar decision:
- Point epipredict to a stable branch/tag, then pull the breaking changes into epiprocess main.  Periodically update epipredict's dependency on epiprocess.
- Create a non-default `dev` branch in epiprocess, pull the breaking changes into epiprocess main.  Periodically update epipredict's dependency on epiprocess.
- Pull the breaking changes into epiprocess default branch (`main` or `dev`).  epipredict will break.  Fix epipredict.  Repeat periodically.

dshemetov:
Looking at the mature Python package [`more_itertools`](https://more-itertools.readthedocs.io/en/stable/versions.html) for inspiration, they use:
- New itertools
- Change to existing itertools
- Other changes
- Bug fixes

So they choose to emphasize when a central piece is either added or updated. The "improvement" distinction is decided on each line when describing the change e.g.
> The implementation for [set_partitions()](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.set_partitions) was improved. (thanks to jferard)

We could do something similar for the core functionality like `epi_df` and `epi_archive`.

dshemetov:
I'm not familiar enough with the details of this package to comment on past version attribution. So I'll only comment on `main` vs `dev` as development branch.

My preference for having a `dev` branch is to have `main` always point to the latest stable release. This way someone can just specify this repo, if they want to install the latest stable version. Unfortunately, this conflicts with making `dev` the default branch that PRs are merged against, which is annoying, but something we could live with.

I'm not entirely sure I understand the .9999 thing, but if it's just a placeholder until we agree on a release version, sounds good.

And if I'm following the epipredict sync thread (which I'm not really sure that I am), I think that not having to pin epipredict to epiprocess versions every time is a benefit of having a separate `main` and `dev` branch -- epipredict can just point to `main` most of the time and if there's something going on in epiprocess that is incompatible, then we can just temporarily pin it to a release.

dshemetov (commented):
See my comment above.

NEWS.md
```diff
@@ -0,0 +1,243 @@
...
+## Improvements:
+
+* Fixed `geo_type` guessing on alphabetical strings with more than 2
+  characters to yield `"custom"`, not US `"nation"`.
+* Fixed `time_type` guessing to actually detect `Date`-class `time_value`s
+  regularly spaced 7 days apart as `"week"`-type as intended.
+* Improved printing of `epi_df`s, `epi_archives`s
+* Fixed `as_of` to not cut off any (forecast-like) data with `time_value >
+  max_version`.
+* Expanded `epi_df` docs include conversion from `tsibble`/`tbl_ts` objects,
```
brookslogan, line 154: Inconsistent tense.  Should just decide.

Here and elsewhere: try to limit to a small number of verbs.  Maybe a recommended verb set, although some of these are trying to accomplish what the categories already do, but we might move from categories -> "importance", and then these would still be useful even though they might look redundant now.

NEWS.md
```diff
@@ -0,0 +1,243 @@
+# epiprocess 0.5.0.9999 (dev version)
```
brookslogan, line 1: Note that this is semver-style, and that 0.x.y we are putting breaking&major changes into x and minor and bugfix into y.

NEWS.md
```diff
@@ -0,0 +1,243 @@
...
+* Ignore `NA`s when printing `time_value` range for an `epi_archive`.
+* Fixed misleading column naming in `epix_slide` example.
+* Trimmed down `epi_slide` examples.
+* Synced out-of-date docs.
+
+## Cleanup:
+
+* Removed dependency of some `epi_archive` tests on an example archive.
+  object, and made them more understandable by reading without running.
+* Fixed `epi_df` tests relying on S3 implementation external to `epiprocess`.
```
brookslogan, line 131: "S3 method", not "implementation".  Spell out that this was epipredict.

pkgdown/extra.scss
```diff
@@ -0,0 +1,69 @@
+/* The news/changelog in pkgdown 2.0.6 is squashed relative to 1.6.1, and
+   secondary headings are too prominent when using ## (but we can't change to
+   ### without impacting side navbar). Just trying a couple of bootswatches
+   didn't seem to help, and nice template packages might be limited for use by
+   particular own groups (e.g., tidytemplate).
```
brookslogan, line 5: "particular own groups"

pkgdown/extra.scss
```diff
@@ -0,0 +1,69 @@
...
+*/
+
+/* Current approach: add some spacing with CSS, and change ## -> ### in the .md
+   to get smaller fonts more easily.
```
brookslogan, line 9: This isn't actually the current approach.

NEWS.md
brookslogan: Partially resolved.  Rest spun out as part of another issue.

brookslogan:
Thanks for the comments!  I'm going ahead and merging this.  Some improvements are deferred to [later](https://github.com/cmu-delphi/epiprocess/issues/233).

Issue #217, opened 2022-08-23
Fix particular outstanding doc and vignette issues


brookslogan:
- [x] Fix `"%s"` in `epix_fill_through_version` docs.
- [x] "[...] typo in the readme file("functions in the epiprocess package that operate on epi_df objects begin with epix", should be "epi_archive" objects)" --- thanks to Tianyue for flagging
- [x] Fix `as_list_col` confusing doc: "Should the new column be stored as a list column?" --- Roughly, "new column" should be "computations"; maybe edit some other words.
- [x] Remove `compactify`-is-not-metadata discussion from vignette.

Copy-pasting some additional items identified by Ryan in an old Slack thread:
- [x] on [this page](https://cmu-delphi.github.io/epiprocess/), we advertise “Get the epipredict R package” but the link takes you back to epiprocess
- [x] same page, we should probably advertise “Get the [epidatr] R package”, and not covidcast, when the former is ready
- [x] on [this vignette](https://cmu-delphi.github.io/epiprocess/articles/correlation.html), now that somebody changed the date range (?) from when I initially wrote it, some things look funny. The blue density for the lag 10 correlation (about halfway down) is weirdly bimodal. Also the argmax lag is now 22 (last plot) and not 18, as the text claims
- [x] In the reference index, we should write the titles using code style, so epi_df instead of epi_df, etc., because this is now possible: https://github.com/r-lib/pkgdown/pull/2113
  - [before/after: check that GitHub Actions pkgdown has this update or otherwise does something reasonable]
- [x] We should go through and make sure that no messages/warnings appear in vignettes that are not important. Like any package loading message or warning, and silly ggplot message or warning , etc. I thought I caught all of them when I wrote vignettes but I either missed some or some new ones got introduced.
- [x] The "Attribution" section at the bottom of each vignette page starts with "This document contains dataset ...". Let's change that to "This document contains a data set ..."

ryantibs:
Thanks @brookslogan for organizing these! I think this should be pretty quick work, can we make this happen sometime soon? (Or you can figure out somebody new-ish to delegate this to? Thank you.)

One more thing to add. The "Attribution" section at the bottom of each vignette page starts with "This document contains **dataset** ...". Let's change that to "This document contains **a data set** ..."

brookslogan:
Regarding the correlation vignette plotting change:
- This change looks like it was from a change in the `geo_values` stemming from:
[...]

PR #236, opened 2022-10-21
Fix various documentation bugs noted in #217

brookslogan:
Still need to finish other tasks in #217 and add changelog entries.


brookslogan:
We'll want to check that the [reference index](https://cmu-delphi.github.io/epiprocess/reference/index.html) renders the code style in the titles properly after merging, as GitHub Actions might be stuck with an older version.

PR #259, opened 2023-01-06
[wip] epix slide over all versions

nmdefries:


dajmcdon:
I don't really have any business reviewing the details of the code changes here. But please add some tests to illustrate/confirm the desired performance.

brookslogan:
Status:
- Nat & I are discussing some design decisions about how things should work and might want to get a wider group's feelings.

In the meantime:
- Nat, you should add yourself as at least as a contributor in the DESCRIPTION file.
- suggestion: epix_truncate_versions code duplication looks avoidable, e.g., by having the grouped_epi_archive method do
```{r}
result = x$clone()
result$ungrouped <- epix_truncate_versions(result$ungrouped, .......)
return (result)
```
and maybe there's even a way to eliminate some awkward cloning by doing the right structure of whether the S3 method calls the R6 method or vice versa, not sure.  But maybe we don't need to be too elegant here because we hope to get rid of the R6 stuff altogether sometime
- Thanks for the `.group_key` fixup; wish we had a more different name for the key of the archive (as well as for the non-version non-time parts of the key --- maybe we could call this the "epikey").  Please share any ideas here.  I think dplyr does call this the group key, unfortunate.

nmdefries:
Thanks for the `epix_truncate_versions_after.grouped_epi_archive` suggestion, that looks a lot better! (Had to use the `ungroup()`/`group_by()` here because of the function context but used the general approach).

nmdefries:
> wish we had a more different name for the key of the archive (as well as for the non-version non-time parts of the key --- maybe we could call this the "epikey")

You're proposing:

- <?> key = version, time value, and epikey
- epikey = geo value, and other user-specified fields (maybe age group, race, etc)

?

`epikey` fits in well with the general naming schema. It's fairly evocative (e.g. a normal epidemiological grouping might be geo value + age group, which we could well imagine in this context), and much much shorter than the current "non-time, non-version keys" phrase that the package docs frequently use. Seems convenient.

brookslogan:
Hi Nat, sorry for the long delay.  I've pushed a couple of changes:
* Added R6 analogues for these methods, to keep things consistent until we can overhaul and remove R6 altogether (and maybe data.table as well).
* Fixed $DT in the computation inputs being a tibble rather than a data.table.

I think we're still missing:
* something in `@examples` with `all_versions=TRUE`; maybe the test I added with `f2` could be a starting point.  A more common/intuitive use case would actually have `f2` be a nowcaster (rather than evaluating a particular lag as a potential nowcaster).  However, one/both of these might be too much code for `@examples`.
* a test for behavior with `group_by(geo_value)` --- I've started on this locally.
* a vignette showing this off --- but let's handle this in another Issue/PR; I want to see if we can combine it with some other vignette ideas.

PR #260, opened 2023-01-20
Avoid `filter(cases==1)`, better describe & `complete` Ebola example

brookslogan:
Avoid `filter(cases==1)`: this should either

* involve `filter(status == "confirmed")` instead, or
* involve no filter, and let `sum` take care of things.

We should generally expect the latter approach to be RAM-friendlier (one new column vs. new filtered vectors for every column), so implement the latter approach.

Reformat the code a bit to try to make it read better.

Clarify the description of the example; make clear we're deriving from line list data.

`complete` the data to make the set of `time_value`s evenly spaced and the same for every `geo_value`; even though it doesn't impact the plots, we're demonstrating how to get things into `epi_df` format, and many functions we'd use on `epi_df`s will expect things in this format.

### DIFF
--- a/vignettes/epiprocess.Rmd
+++ b/vignettes/epiprocess.Rmd
@@ -60,6 +60,7 @@ API](https://cmu-delphi.github.io/delphi-epidata/api/covidcast.html).
 library(epidatr)
 library(epiprocess)
 library(dplyr)
+library(tidyr)
 library(withr)
 
 cases <- pub_covidcast(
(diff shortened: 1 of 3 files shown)


vignettes/epiprocess.Rmd
```diff
@@ -263,24 +264,28 @@ ggplot(x, aes(x = time_value, y = value)) +
...
+      "Eastern",
+    district %in% c("Bombali", "Kambia", "Koinadugu", "Port Loko",
+                    "Tonkolili") ~
+      "Northern",
+    district %in% c("Bo", "Bonthe", "Moyamba", "Pujehun") ~
+      "Sourthern",
+    district %in% c("Western Rural", "Western Urban") ~
+      "Western")) %>%
+  group_by(geo_value = province, time_value = date_of_onset) %>%
+  summarise(cases = sum(if_else(status == "confirmed", 1, 0)),
```
dajmcdon, line 284: Can this just be 
```r
summarise(cases = sum(status == "confirmed), .groups = "drop") %>%
```

dajmcdon (approved):
Feel free to ignore my suggestion. Everything else appears fine. Merge when ready.

vignettes/epiprocess.Rmd
brookslogan: I think it can.  This was an attempt to be beginner-friendly, but the shorter form reads pretty well.

dajmcdon: If you edit this, we should probably propagate to https://github.com/cmu-delphi/delphi-tooling-book/blob/a9a5ad01bad1b5cbd7fb8fe79dc0196b0011c876/epidf.qmd#L238

brookslogan:
Finally getting around to revisiting this.  I thought there was something else to do here, but apparently there was only styling left?  I've held off on converting things to use `case_match()` until we want dplyr 1.1.0 functionality for something more important, as potential encoding breaking changes and `summarize()` complaint spam may have people holding off.

PR #268, opened 2023-02-27
Repair/Update pkgdown GitHub Action

brookslogan:


brookslogan:
The above checks don't tell use whether the pkgdown action is going to succeed (and the checks were already using `@v2` actions).  Going to just go ahead and merge to check if this repairs the site build.  Might be good to explore, e.g., https://github.com/nektos/act, to see if there's a way to test locally in the future.

PR #269, opened 2023-02-27
Update GitHub Actions themselves, not just embedded dependencies

brookslogan:
```{r}
usethis::use_github_action("pkgdown")
usethis::use_github_action_check_release() # and follow instructions regarding badge
```


dajmcdon (approved):

brookslogan:
(Migrating note to GitHub:)  It looks like the updated pkgdown Action will try building the site for PRs and give us a heads up about failures before they are merged into `main`.  The tradeoff is that this likely involves building the vignettes another time (I think current checks build the vignettes twice, so this would be a third time), meaning potentially more time spent waiting on Actions to complete.

PR #283, opened 2023-03-20
Bump version to 0.6.0

brookslogan:
Keeping this in the `dev` branch as pushing to `main` will break `epipredict`, and it might make sense to incorporate repairs for the current breaking changes with those for some other planned breaking changes that are hopefully imminent.


brookslogan:
The bump to 0.6.0.9999 was incomplete; the DESCRIPTION file should also have been updated.

PR #284, opened 2023-03-21
add epix_slide example comparing same computation with all_vers FALSE

nmdefries:
Add example to `epix_slide` documentation comparing the same computation for `all_versions = TRUE` and `all_versions = FALSE`.


brookslogan (approved):
I eliminated a warning from `min(x$DT$version)` and added some more commentary.  (The special strings we have for 0-row inputs don't show up now due to the recycling rules, but will appear in combination with #311.)

(GitHub isn't running checks on the dev branch; running checks locally, I only see irrelevant errors.  I think I have the auto checks enabled for `dev` in #311 along with fixes for some of these irrelevant check errors/warnings.)

Issue #163, opened 2022-07-22
Draft and discuss naming schemes for `epix_slide` parameters, output


brookslogan:
See https://github.com/cmu-delphi/epiprocess/issues/146#issuecomment-1192785302; `time_value` vs `version` bullet point.

brookslogan:
"Discuss" will eventually mean running sketches of use by some potential users / other developers.

brookslogan:
As discussion on #170 and #171 has brought up, another naming option for the relevant output column(s), besides `time_value`, `version`, and both, is `ref_time_value`.

brookslogan:
Some remaining discussion points from #146:

> - Separate discussion: should we rename `ref_time_values` and `time_value` output column to something involving `version`, or keep the former and have the latter turn into two duplicate columns with both names?  Should we output an `epi_archive`?
[or should we call the time/version output column `ref_time_value`?]

> - Separate discussion: should we rename `max_version` parameter of `epix_slide` to `version`?
[Since we're moving to more consistently use an "implicit versioning" scheme, where last-version-of-each-observation-carried-forward is assumed everywhere in archives, this may make sense.  However, we might then need to think about the naming or discussion of the `$DT$version` column.]

Some other existing mismatches between slide operations that we might want to think about:
- `mutate`-like vs `summarize`-like
  - `epi_slide` is like `mutate`: it keeps existing columns and, given scalar output from `f`, broadcasts each slide result to the associated input locations to maintain size stability
  - `epix_slide` is like `summarize`: it only produces the grouping columns + `f` results, and doesn't broadcast
- Once `n` -> `before`, `after` / `before` hits:
  - `before=k`, `after` missing in `epi_slide` means a trailing/right-aligned window that will actually have data at that right side of the window (as we ensure `ref_time_values %in% unique(x$time_value)`), unless there are some variable-time-values-by-group things to think about
  - `before=k`, `after` missing/not-accepted-as-an-arg in `epix_slide` means a window extending infinitely far into the future, but in typical surveillance data cases, will only contain data up to some time before the associated `ref_time_value`; to call it trailing/right-aligned doesn't seem precise either way.

brookslogan:
Another idea to consider here: guess what label to use for the ref_time_values based on the user output: if they provide a(n epi_)df with a `time_value` column, then use `version`; else use `time_value`.

PR #290, opened 2023-03-28
Make `epix_slide` more like `reframe`

### LINKED ISSUE #275: Change `epix_slide` to be more `group_modify`-like
Migrated from #64.

Foreseen common use cases of `epix_slide`:
* Pseudo-prospective forecasting/pancasting/modeling.
* Getting time lags of observations/functions-of-observations as of particular version lags, to:
  * Evaluate accuracy of real-time/bleeding-edge data
  * Prepare apples-to-apples training data for revision-aware forecasters
* Summarizing statistics regarding latency of data sources, missingness/zero-ness/duplications in bleeding-edge data.

Note that:
1. For pseudoprospective forecasting, we typically want multiple rows per group-reftime, as we may be predicting multiple targets, multiple quantiles, etc.  These multiple rows are based on new key columns introduced in the slide computation (e.g., the target date and quantile level), not the old non-grouping parts of the epikey (e.g., the `geo_value` if we are fitting all geos simultaneously).  We might want to output a different epikey set if we are performing/tacking on a geo/age/etc. aggregation.
2. For extracting version&time-lags of functions of data: we probably want either (a) exactly 1 or (b) either 0 or 1 row per group-reftime (with multiple columns for different functions, lags, etc.).  We might want to output a different epikey set if we are performing a geo/age/etc. aggregation.
3. For summary statistics, we might want to output any number of rows, depending on the analysis.

Except maybe 2(a) with no epikey aggregation, we don't want the 1-per-reftime-epikey or 1-per-epikey-broadcasted-to-epikeys behavior.

So, we should make `epix_slide` more `reframe`/`group_modify`-like than grouped-`mutate`-like.  (Its output column behavior is already like the former; this would focus on row handling.)
- [x] Allow the computations to output any number of rows.  Do not perform row broadcasting. Be careful to remove descriptions of size stability and exact forecast compatibility with `epi_slide` if not true.
- [x] Adjust or deprecate all_rows for epix_slide.
- ~Check that returning computation results with new key columns (e.g., target date, quantile level, ...) result in an appropriately-keyed output, or consider a parameter to express the introduction of new key variables.~ [punted]

brookslogan:
Resolves #275, addresses _pkgdown.yml typo, adds GitHub Actions checks to dev branch PRs, implements current approach to #213, adds missing `test-grouped_epi_archive.R` file.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [main, master]
   pull_request:
-    branches: [main, master]
+    branches: [main, master, dev]
 
 name: R-CMD-check
 
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [main, master]
   pull_request:
-    branches: [main, master]
+    branches: [main, master, dev]
   release:
     types: [published]
   workflow_dispatch:
--- a/R/archive.R
+++ b/R/archive.R
@@ -640,7 +640,7 @@ epi_archive =
           slide = function(f, ..., before, ref_time_values, 
                            time_step, new_col_name = "slide_value",
                            as_list_col = FALSE, names_sep = "_",
-                           all_rows = FALSE, all_versions = FALSE) {
+                           all_versions = FALSE) {
             # For an "ungrouped" slide, treat all rows as belonging to one big
             # group (group by 0 vars), like `dplyr::summarize`, and let the
             # resulting `grouped_epi_archive` handle the slide:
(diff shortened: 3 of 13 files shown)


brookslogan:
Sometimes we can get an `epi_df` out of `epix_slide` as I believe was originally intended.  But not very elegantly:
* We don't output `epi_df` when ungrouped
* We output `epi_df` when grouped by `geo_value`, but if slide computation was trying to output an `epi_df` or add a key column, it's ignored/corrupted by default, or raises errors with `names_sep=NULL`.

And if we want an `epi_df` out of `epix_slide` with `time_value`s from the slide computations, not the `ref_time_values`, we may be stuck trying to wrap `epi_df`s in our computation inside a list column of a tibble output, then attempt to unnest and deal with naming conflicts simultaneously, which may not be possible, so we may have to rename & lose any outer `epi_df`-ness, then unnest, then add an other_keys entry for whatever the `ref_time_values` label was.  Or, alternatively, we don't wrap, do the rename shuffling, then have to reclass manually and reset metadata, some of which would have been discarded in intermediate steps (which means hardcoding or breaking the pipe to assign to a temporary to allow later access; both awkward).

I'm not going to try to make this easier, because it again raises questions of whether we really do want an `epi_df` out sometimes, or if we want something else.  Perhaps we should only output a tibble like with `all_versions=TRUE`, or consider alternatives such as another `epi_archive` or a generalized `archive`.  And revisit #163.

brookslogan:
This PR is based off of a branch from the wrong remote.  It is replaced by #311.

Issue #296, opened 2023-04-07
Grouped `epi_df` can/always loses `epi_df`ness on `summarise`


brookslogan:
See https://github.com/cmu-delphi/epiprocess/issues/298.

brookslogan:
The original issue from epipredict has been transferred to this repo, so this is now a duplicate.

PR #297, opened 2023-04-07
Keep `epi_df`-ness in `summarise` + fix roxygen typos

brookslogan:


brookslogan:
Closes #296.  But this may not be the best way to do it.  Maybe we'd get some of the intended behaviors of these reconstructing wrappers via implementing `group_data.epi_df`.  But some might be unavoidable depending on `grouped_df` method implementations.  There's also still issues with the metadata when binding rows which might be easier to resolve in some wrappers like the current approach (though none of them do that right now).  Especially with dplyr 1.1.0 summarize that deprecates returning != 1 row per group, we may also have some re-keying hints that we might need to act upon.

brookslogan:
I think an alternative approach involving an `"grouped_edf"` class is going to be preferable.

Issue #168, opened 2022-07-26
Validate `f` function in `epi[x]_slide`; give better feedback if doesn't take enough args


brookslogan:
See this example:
```
epi_slide(jhu_csse_county_level_subset, function(x) print(x))
Error in .f(.x, ...) : unused argument (list())
```
The error message isn't that informative, and seems like it might be fairly commonly encountered.  (I have trouble remembering the `g` arg despite discussing it for a while.)

We can check that `f` takes enough arguments beforehand.  I believe I have something along these lines for a different function in #102.  (See `fn_max_n_args` definition and use.)

brookslogan:
I've hit this while trying to use `lm.fit` or `slice_min` as the computation, which may give more confusing error messages than the above.

brookslogan:
Having checks also hit the above cases might require something more detailed than checking `fn_max_n_args`, such as requiring the number of positional args before seeing any `...` to be at least the number of mandatory arguments for the slide.  (This might require explicitly forwarding some things to `comp_one_group` that previously we forwarded via `...`, but might catch `slice_min` even though it takes more than enough args.)

nmdefries:
> I've hit this while trying to use lm.fit or slice_min as the computation

nmdefries:
> I've hit this while trying to use lm.fit or slice_min as the computation

Both `lm.fit` and `slice_min` take two positional args before `...`, and are successfully passed `x` and `g` by `epi_slide`. The errors

```
epi_slide(jhu_csse_county_level_subset, slice_min, before=2L)
# Error in `.f()`:
# ! Problem while computing indices.
# Caused by error in `UseMethod()`:
# ! no applicable method for 'group_keys' applied to an object of class "rlang_data_pronoun"
```

```
epi_slide(jhu_csse_county_level_subset, lm.fit, before=2L)
# Error in .f(.x, ...) : incompatible dimensions
```

both come from internal arg validation by the provided function, so this behavior seems correct. (Also, since both functions already take 2 args before `...` the approach proposed above wouldn't catch them; we could check arg names?)

nmdefries:
Closed in https://github.com/cmu-delphi/epiprocess/pull/302

PR #302, opened 2023-04-21
Check that the `f` passed to `epi[x]_slide` takes enough args

nmdefries:
Currently we expect `f` to take 2 args, `x` and `g`.

If `f` has `...` as one of its args, it ideally would take 2 or more args before `...`; a warning is raised if not. We don't error in this case to add flexibility. A user may want to handle only `x`, like `function(x, ...) { <do stuff with x only> }` if they don't need the group key or ref time value, letting `...` capture the other args.

If `f` doesn't have `...` as one of its args, it must take 2 or more args in total, and exits with an error if not.

Closes https://github.com/cmu-delphi/epiprocess/issues/168


R/utils.R
```diff
@@ -100,6 +100,36 @@ paste_lines = function(lines) {
...
+#' @noRd
+check_sufficient_f_args <- function(f) {
+  n_mandatory_f_args <- 2
+  arg_names = names(formals(args(f)))
+  if ("..." %in% arg_names) {
+    # Keep all arg names before `...`
+    dots_i <- which(arg_names == "...")
+    arg_names <- arg_names[seq_len(dots_i - 1)]
+
+    if (length(arg_names) < n_mandatory_f_args) {
```
brookslogan, line 117: issue: While this should catch the `slice_head` issue, I don't think this will catch something like trying to use `lm` and having some of its parameters-that-have-defaults filled with the group key / ref_time_value.  This could probably be checked by seeing if any of the first `n_mandatory_f_args` `formals(args(f))` are `identical` to `quote(expr =)` or `rlang::expr()` --- see https://rlang.r-lib.org/reference/missing_arg.html.

R/utils.R
```diff
@@ -100,6 +100,36 @@ paste_lines = function(lines) {
...
+      Warn(sprintf("`f` only takes %s positional arguments before the `...` args, but %s were expected; this can lead to obtuse errors downstream", length(arg_names), n_mandatory_f_args),
```
brookslogan, line 118: except this is already worded in terms of arguments.  seems okay.

tests/testthat/test-epi_slide.R
```diff
@@ -86,3 +86,15 @@ test_that("these doesn't produce an error; the error appears only if the ref tim
...
+
+test_that("epi_slide alerts if the provided f doesn't take enough args", {
+  f_xg = function(x, g) dplyr::tibble(value=mean(x$value), count=length(x$value))
+  # If `regexp` is NA, asserts that there should be no errors/messages.
+  expect_error(epi_slide(grouped, f_xg, before = 1L, ref_time_values = d+1), regexp = NA)
+  expect_warning(epi_slide(grouped, f_xg, before = 1L, ref_time_values = d+1), regexp = NA)
+
+  f_x_dots = function(x, ...) dplyr::tibble(value=mean(x$value), count=length(x$value))
+  expect_warning(epi_slide(grouped, f_x_dots, before = 1L, ref_time_values = d+1),
+    regexp = "positional arguments before the `...` args",
```
brookslogan, line 98: note: probably don't need to specify the regexp if you have the class

brookslogan (changes requested):
Looks good, but I have a few edits (will push momentarily) + we should probably try to warn on attempts to use functions like `lm`.

R/utils.R
nmdefries: Added in https://github.com/cmu-delphi/epiprocess/commit/d5466d38db3a149fe2a990e6f91b1561836abab0

I used `rlang::is_missing` since it [already contains the `identical(x, quote(expr =)` check](https://github.com/r-lib/rlang/blob/84bd450/R/arg.R#L463-L465) you mentioned.

R/utils.R
```diff
@@ -100,6 +100,36 @@ paste_lines = function(lines) {
...
 
+#' Check that a sliding computation function takes enough args
+#'
+#' @param f Function; specifies a computation to slide over an `epi_df` or
+#'  `epi_archive` in `epi_slide` or `epix_slide`.
+#'
+#' @noRd
+check_sufficient_f_args <- function(f) {
+  n_mandatory_f_args <- 2
+  arg_names = names(formals(args(f)))
```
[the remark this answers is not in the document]
nmdefries: I took this from [your code](https://github.com/cmu-delphi/epiprocess/blob/4ae435944ddbfe6b98d498f90def57dc1adc31e8/R/utils_conditions.R#L357-L358) on https://github.com/cmu-delphi/epiprocess/pull/102... so good job yourself!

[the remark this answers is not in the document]
brookslogan: Well, definitely don't remember that, lol.  Glad I "found out" too late that `rlang::fn_fmls_names` was "just better".

brookslogan (approved):
I used a `suppressWarnings` to prevent some warnings from a warning+error case from popping up during `test()`.  Not sure if we want warning+error but it's probably okay.

I also (hopefully) addressed some technicalities dealing with dots forwarding; this could use some review.  I'll go ahead and mark this approved so you can go ahead and merge if it looks all right.

R/utils.R
```diff
@@ -103,46 +103,72 @@ Warn = function(msg, ...) rlang::warn(break_str(msg, init = "Warning: "), ...)
...
+  # calling args. If `f` has fewer than n_mandatory_f_args before `...`, then we
+  # only need to check those args for defaults. Note that `dots_i - 1L` is
+  # length 0 if `f` doesn't accept `...`.
+  n_remaining_args_for_default_check = min(c(dots_i - 1L, n_mandatory_f_args))
+  default_check_args = remaining_args[seq_len(n_remaining_args_for_default_check)]
+  default_check_args_names = names(default_check_args)
+  has_default_replaced_by_mandatory = map_lgl(default_check_args, ~!is_missing(.x))
+  if (any(has_default_replaced_by_mandatory)) {
+    mandatory_args_replacing_defaults =
+      mandatory_f_args_labels[seq_len(sum(has_default_replaced_by_mandatory))]
```
nmdefries, line 166: issue: `mandatory_args_replacing_defaults` should be defined positionally. The `seq_len(sum())` is causing both cases `f(x=1, g)` and `f(x, g=1)` to warn that "`epi[x]_slide` would pass the window data" to the arg with the default. But the second case would pass the group key arg instead.

```suggestion
      mandatory_f_args_labels[has_default_replaced_by_mandatory]
```

R/utils.R
```diff
@@ -103,46 +103,72 @@ Warn = function(msg, ...) rlang::warn(break_str(msg, init = "Warning: "), ...)
...
-    arg_names_before_dots <- arg_names[seq_len(dots_i - 1)]
-
-    if (length(arg_names_before_dots) < n_mandatory_f_args) {
-      Warn(sprintf("`f` only takes %s positional arguments before the `...` args, but `epi[x]_slide` will call it with at least %s positional arguments; if `f` doesn't expect those arguments, it may produce confusing error messages", length(arg_names), n_mandatory_f_args),
-           class="epiprocess__assert_sufficient_f_args__f_needs_min_args_before_dots",
-          epiprocess__f = f,
-          epiprocess__arg_names_before_dots = arg_names_before_dots)
+    mandatory_args_mapped_names <- remaining_args_names[seq_len(dots_i - 1L)]
+
+    if (dots_i - 1L < n_mandatory_f_args) {
```
nmdefries, line 134: suggestion (non-blocking): We use `dots_i - 1L` a number of times. Defining a variable to store it would make the code more readable.

nmdefries (commented):
@brookslogan Mostly looks good. Thanks for handling dots, I forgot that that was a complication.

One suggestion that I'll add in, and what looks like an issue -- would like you to verify.

R/utils.R
```diff
@@ -100,6 +100,80 @@ paste_lines = function(lines) {
...
+  remaining_args_names = names(remaining_args)
+  # note that this doesn't include unnamed args forwarded through `...`.
+  dots_i <- which(remaining_args_names == "...") # integer(0) if no match
+  n_f_args_before_dots <- dots_i - 1L
+  if (length(dots_i) != 0L) {
+    # Keep all arg names before `...`
+    mandatory_args_mapped_names <- remaining_args_names[seq_len(n_f_args_before_dots)]
+
+    if (n_f_args_before_dots < n_mandatory_f_args) {
+      mandatory_f_args_in_f_dots = tail(mandatory_f_args_labels, n_f_args_before_dots)
```
nmdefries, line 137: issue: The list of mandatory args that will be passed to `f`'s dots should be the last `n` values in `mandatory_f_args_labels`, where `n` is the # of mandatory args in excess of the # of positional args that `f` takes.

```suggestion
      mandatory_f_args_in_f_dots = tail(mandatory_f_args_labels, n_mandatory_f_args - n_f_args_before_dots)
```

brookslogan: Nice catches!  For this one, I'll try a shorter version with 2nd arg just `-n_f_args_before_dots`.

brookslogan: nope, actually I just made a bug trying to be fancy.  Fixing to your approach.

brookslogan: Still working out some formatting details and a couple bugs I discovered when adding regexp's for the relevant tests.

R/utils.R
brookslogan: This actually had a slight bug: `mandatory_f_args_labels` isn't necessarily of the same length as `has_default_replaced_by_mandatory`.  Should be fixed now.  Would have been easier to spot if we had better names for some of these variables, but I can't think of better names at the moment.

PR #315, opened 2023-05-11
Djm/remove fabletools

dajmcdon:
Removes some package Imports used only in `detect_outlr_stl()`.

* All checks pass. Some local comparisons of the guts is a few cases matched up to reasonable precision.
* Vignette output looks visually identical.
* No new tests have been added ...


brookslogan:
The `Collate:` is to get some roxygen documentation with the same `@rdname` spread across multiple files to render in the way I wanted.  Following up with some other questions on Slack.

dajmcdon:
I've made the fixes based on our Slack discussion. Passes local checks, though there are some warnings due to functions I haven't touched.

I moved `lubridate` to Suggests. There were two imported functions that weren't used anywhere that I could find, so that is removed as well. We still use it in some tests and link to it in documentation, but it isn't needed for any code AFAICT. Maybe when CI runs, it will throw an error though.

brookslogan:
Rebased to address merge conflicts, apply styling, and squash some commits.  Tested that detect_outlr example still gives effectively the same results (A simple `waldo::compare(old, new)` complains of differences, but at `tolerance = 1e-8` will not).

brookslogan (commented):
I think I've finished addressing merge conflicts now via rebasing.  I've left a few notes here to end the day.  Most are minor things that I can probably quickly clean up.  But I was sidetracked by a larger surprise, which is that our method of non-seasonal fitting is sensitive to the fake seasonal settings.  I can continue looking into whether there's a way to get a "cleaner" non-seasonal fit a bit, but would be glad if you have ideas/plans here.  I'm quite unfamiliar with STL&LOESS.

R/outliers.R
```diff
@@ -222,6 +222,7 @@ detect_outlr_rm <- function(x = seq_along(y), y, n = 21,
 #' @param seasonal_period Integer specifying period of seasonality. For example,
 #'   for daily data, a period 7 means weekly seasonality. The default is `NULL`,
 #'   meaning that no seasonal term will be included in the STL decomposition.
+#'   If specified, it must be strictly larger than 1.
 #' @template outlier-detection-options
 #' @template detect-outlr-return
 #'
```
brookslogan, line 228: todo: below, remove ref to `feasts`

brookslogan:
I've once again rebased to
- add `lubridate` back to Imports:, as the new function `epi_slide_opt()` uses `lubridate::as.period()`.

I've also pushed changes to address my feedback.  I gave up on trying to actually remove the seasonal component and instead just tried to make the function slightly more flexible and make the documentation not lie.

brookslogan:
Hi @nmdefries, could you please take a quick sanity check pass over this?  I've done some git surgery on it + tried to fix a few things and it'd be helpful to have an independent pass.

R/outliers.R
```diff
@@ -216,18 +217,28 @@ detect_outlr_rm <- function(x = seq_along(y), y, n = 21,
...
+#'   predictions, so large seasonal(/weekly) components will not lead to
+#'   flagging points as outliers. `TRUE` may instead consider the extrema of
+#'   large seasonal variations to be outliers; `n_seasonal` and
+#'   `seasonal_period` will still have an impact on the result, though, by
+#'   impacting the estimation of the trend component.
 #' @template outlier-detection-options
 #' @template detect-outlr-return
 #'
-#' @details The STL decomposition is computed using the `feasts` package. Once
+#' @details The STL decomposition is computed using the [`stats::stl()`]. Once
```
nmdefries, line 239:
```suggestion
#' @details The STL decomposition is computed using [`stats::stl()`]. Once
```

nmdefries (commented):
Looks reasonable enough, but it would be nice to have some tests to verify behavior. I assume existing tests would break with these changes.

brookslogan:
Thanks for the pass.  Thinking about testing:
- [Golden reference / characterization tests](https://github.com/cmu-delphi/forecasting-team-meta/issues/13) for all of our examples and vignettes would be nice.  I actually manually did this for the example + `x` in the outliers vignette and they actually PASS `waldo::compare(tolerance = 1e-6)` if you renaming "nonseasonal"/"reseasonal" to the same thing!  It looks like fabletools must have inferred the same seasonality frequency we used (7).
  - testthat has support for text-based golden tests + I hacked together an RDS-based one somewhere [in this repo](https://github.com/ryantibs/quantgen/issues/24).  Maybe I should apply one of these approaches to (some fixed subsampling of?) the vignette's `x`.
- Other tests / actual tests in CI... I'm not sure what they should be, and I'm not sure if we might do some sort of major overhaul of outlier detection if we ever get around to focusing on the "methods" part of it.  Going to forget about these for now.

nmdefries:
IIRC, @dshemetov did some work on comparing vignettes before/after a change. I wonder if he turned up anything there that we can run automatically.

dshemetov:
[Here's](https://github.com/cmu-delphi/epiprocess/pull/431#issuecomment-2083983708) what I ended up doing. I didn't keep it in, because I was concerned committing vignette outputs would balloon the repo size.

brookslogan:
Thanks, I forgot that you already implemented something for this.  Agree that if we want it in CI, it'd be better to find some way to keep the vignette outputs out of the repo (e.g., using GitHub artifacts).  I ended up just visually comparing the vignettes and they look similar enough.

[Again](https://github.com/cmu-delphi/epiprocess/pull/315#issuecomment-2183374466), I'm going to forget about adding other tests for now and go ahead and merge.

brookslogan (approved):
Approving in an [unsuccessful] attempt to wake up or regenerate CI.  [Using workflow dispatch instead.]

brookslogan:
The workflow dispatch [R CMD check run](https://github.com/cmu-delphi/epiprocess/actions/runs/9980372696) run didn't attach to this PR.  It has some unrelated failures that will be skipped by #480.  Force-push above is to rebase on top of dev to get these test skips.

PR #319, opened 2023-05-23
Update to use latest `epidatr` (`fetch_tbl` -> `fetch`)

brookslogan:


brookslogan:
The warning checks are being addressed in another branch.  This PR isn't going to address its real purpose when we're just merging to dev.  But dev has changes we plan to revert, and additional breaking changes planned.  We should decide what to do at the tooling meeting.

Issue #327, opened 2023-06-08
Properly attribute code and data sources


nmdefries:
`as_slide_computation` borrows heavily from `rlang::as_function`. Currently attribution is given informally via the `roxygen` `@source` tag and in the function description. We would also want to list JHU for data usage.

Logan looked into [possibilities here](https://github.com/cmu-delphi/epiprocess/pull/313#discussion_r1206149945), but it looks like there's no single official way.

brookslogan:
Potentially missing from the above investigation is attaching a file LICENSE that notes the (non-MIT) data set licenses.

nmdefries:
In the `knitRProgressBar` package, the author [added the original license text and a description of changes](https://github.com/rmflight/knitrProgressBar/commit/c316878b55a63b51086ed0537490680fc740e2f4) to the source code, and [added the original authors to the DESCRIPTION](https://github.com/rmflight/knitrProgressBar/commit/4d9b28e3b2245bfd5e223ce168c46a94e1edd138) as contributors (`ctb`). No other changes were made for attribution.

dajmcdon:
Are these attributions actually necessary? I'm not sure that `knitRProgressBar` is the appropriate model. How do packages in the tidyverse/r-lib actually approach this? That is, do the people who work directly with `rlang` contributors credit `rlang` in this way?

(I get that the Writing R packages document recommends this, but if followed, I suspect that all packages on CRAN would have dozens of aut/ctb fields, which doesn't seem to be the case.)

nmdefries:
The motivation for this attribution is more than just that `epiprocess` _uses_ `rlang` functions. The initial version of `epiprocess:::as_slide_computation` was copy-pasted, including documentation, from `rlang`'s source code. I added maybe a line of code.

Our function now has more differences, but it is more than "inspired" by `rlang::as_function`, although it's not a substantial amount of code.

As for `tidyverse`, the only member package with additional external authors listed is `readr`, see its [DESCRIPTION](https://github.com/tidyverse/readr/blob/main/DESCRIPTION#L12-L17). `mio` and `grisu3` seem to be C++ packages that `readr` copy-pasted code from, e.g. see [`readr`'s `mio.h`](https://github.com/tidyverse/readr/blob/0648707/src/mio.h#L105-L567) vs [`mio`'s `mmap.hpp`](https://github.com/mandreyel/mio/blob/master/include/mio/mmap.hpp). Both C++ packages use the MIT license.

[...]

PR #338, opened 2023-06-29
Attribute code borrowed/modified from `rlang` for use in `as_slide_computation`

nmdefries:
### Description
Add `rlang` authors and copyright holder to DESCRIPTION file. Add COPYRIGHT file. Add `rlang` license text to the documentation for `as_slide_computation`, noting aspects of the original code that were kept and changed.

### Fixes
Partially addresses https://github.com/cmu-delphi/epiprocess/issues/327

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -24,7 +24,7 @@ Description: This package introduces a common data structure for epidemiological
     work with revisions to these data sets over time, and offers associated
     utilities to perform basic signal processing tasks.
 License: MIT + file LICENSE
-Copyright: file COPYRIGHTS
+Copyright: file inst/COPYRIGHTS
 Imports:
     cli,
     data.table,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -16,13 +16,15 @@ Authors@R: c(
     person("Dmitry", "Shemetov", role = "ctb"),
     person("Ryan", "Tibshirani", , "ryantibs@cmu.edu", role = c("aut", "cre")),
     person("Lionel", "Henry", role = "ctb", comment = "Author of included rlang fragments"),
-    person("Hadley", "Wickham", role = "ctb", comment = "Author of included rlang fragments")
+    person("Hadley", "Wickham", role = "ctb", comment = "Author of included rlang fragments"),
+    person("Posit", role = "cph", comment = "Copyright holder of included rlang fragments")
   )
 Description: This package introduces a common data structure for epidemiological
     data reported by location and time, provides another data structure to
     work with revisions to these data sets over time, and offers associated
     utilities to perform basic signal processing tasks.
 License: MIT + file LICENSE
+Copyright: file COPYRIGHTS
 Imports:
     cli,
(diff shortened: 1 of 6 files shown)


inst/COPYRIGHTS
```diff
@@ -0,0 +1,26 @@
+Copyright for most of the included code is held by the Delphi group.
```
nmdefries, line 1: > held by the Delphi group

Maybe?

nmdefries: Wondering about the format for `inst/COPYRIGHT`.

Looks like there is no single standard format. We could do a free-form version, but it will be unwieldy and unreadable if we include lots of sources and licenses.
Some formats from other R packages that seem particularly appealing when listing multiple sources are:
- [Block for each library](https://github.com/zhengxwen/gdsfmt/blob/d82c4f0/inst/COPYRIGHTS) that includes the library's license and/or notes on usage
- [Debian package standard for copyright files](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0)
    - [Simple example](https://github.com/kaskr/adcomp/blob/master/TMB/inst/COPYRIGHTS)
    - [More complex example](https://github.com/trevorld/r-optparse/blob/f7d6363/inst/COPYRIGHTS)

DESCRIPTION
```diff
@@ -14,13 +14,17 @@ Authors@R: c(
...
+    person("Lionel", "Henry", role = "ctb", comment = "Author of included rlang fragments"),
+    person("Hadley", "Wickham", role = "ctb", comment = "Author of included rlang fragments"),
+    person("Posit", role = "cph", comment = "Copyright holder of included rlang fragments")
   )
 Description: This package introduces a common data structure for epidemiological
     data reported by location and time, provides another data structure to
     work with revisions to these data sets over time, and offers associated
     utilities to perform basic signal processing tasks.
 License: MIT + file LICENSE
+Copyright: file COPYRIGHTS
```
brookslogan, line 27:
```suggestion
Copyright: file inst/COPYRIGHTS
```

brookslogan (changes requested):
Looks great overall.  A few typos + a re`document()` needed, plus resolution/punting of a huge headache: line 1 of inst/COPYRIGHTS.

inst/COPYRIGHTS
nmdefries: Since we only are including one external copyright/license, not going to worry about the format for now.

I'm going to go with option a) and not say anything about the majority of the code. Will open an issue for figuring out the copyright on our end.

nmdefries: https://github.com/cmu-delphi/epiprocess/issues/348
https://github.com/cmu-delphi/epiprocess/issues/349

brookslogan:
I missed that this was merging to a feature branch; maybe that's why GitHub's saying rebase+merge.  I'll leave it to you to press the button whichever way.  But if you think rebase should/shouldn't be the way we do things on main please let me know.

PR #346, opened 2023-07-13
ci: stop building pkgdown on forks

dsweber2:
This makes it so that pkgdown will only run if the repo is owned by cmu-delphi directly. I'd rather pkgdown stopped trying to run on my fork whenever I sync it with main or dev, causing failing build notifications. This should solve it, taken from [here](https://github.com/orgs/community/discussions/26409), and working on epidatr.

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -15,6 +15,7 @@ name: pkgdown
 
 jobs:
   pkgdown:
+    # only build docs on the main repository and not forks
     if: github.repository_owner == 'cmu-delphi'
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


brookslogan (approved):
Interesting, I thought that this type of CI would automatically not run on forks.  Right now our workflow is to have branches off of the cmu-delphi repo rather than in forks; LMK if you are lacking permissions.

todo: please add on to the line 4 comment to describe this change.

dsweber2:
Added. Apparently this needs approval from both you and @dajmcdon? Y'all may want to make that either, rather than both, for future pull requests.

I currently can't add branches directly to epiprocess or epipredict

PR #378, opened 2023-11-16
style: run package through styler

dshemetov:


R/epi_df.R
```diff
@@ -107,64 +107,68 @@ NULL
...
+    else if ("as_of" %in% names(x)) {
+      as_of <- max(x$as_of)
+    } else if ("issue" %in% names(x)) {
+      as_of <- max(x$issue)
+    } else if ("version" %in% names(x)) {
+      as_of <- max(x$version)
+    } # If we got here then we failed
+    else {
+      as_of <- Sys.time()
+    } # Use the current day-time
```
brookslogan, line 154:
```suggestion
    } else {
      # If we got here then we failed; use the current datetime
      as_of <- Sys.time()
    }
```

brookslogan (commented):
From a skim, looks fine.  But what's the process for pending PRs?  Git rebase --exec, something like [this](https://blog.scottlogic.com/2019/03/04/retroactively-applying-prettier-to-existing-branches.html)?

brookslogan:
Oh, not sure if rebase styling would even be needed if we are merging PRs rather than rebasing PRs.

brookslogan:
Non-rebase approach is also going to make reviewing a hassle if it's not done at the very very end.  Might ask PR authors if we can apply the rebase approach & do that then.  Locally at least we should make sure to back up the old version until everything looks good.

brookslogan (approved):

brookslogan:
As I'm trying to clean up some old PRs, I'm revisiting these handy commands.  Post-merge of the styler, it looks like `git rebase origin/dev` will complain about conflicts due to the merged styling.  I might have been able to just skip that for one of the PRs, but for another, it looks like maybe it should first be rebased on 58ed6b45f77eb252ee5f721ff0ae8ef93888e4a6, the commit preceding the initial styling commit.  If we use this approach as a template for other mass-restyling operations, we should probably tag that pre-initial-styling commit to do the same.

PR #377, opened 2023-11-16
fix: update for epidatr 1.0.0

dshemetov:
- Fixes vignettes currently failing.
- Fixes dataset downloading code that relies on epidatr. It might seem like there a lot of changes, but most of those are my IDE automatically applying styler when I saved.
- Fixes formatting errors in NEWS.md.

PR #379, opened 2023-11-27
Djm/pkgdown prettier

dajmcdon:
Improve the pkgdown site (nice and red and matches `epipredict`).

Checks fail until #377 is merged.


dajmcdon:
@brookslogan needs `ds/clean` to pass checks (at least).

brookslogan:
I think we're still trying to decide on branch organization --- it appears that pkgdown lets you build main & dev documentation sites independently of how you organize your branches (instead basing it on whether you use a 3-number or 4-number version tag) --- but I believe we decided to try to standardize the appearance across both main & dev sites rather than maintain separate styles.  So I will try to merge this in.  One thing we need to check is whether there is some auto-differentiation of style by pkgdown using red text for dev version numbers that may not be visible against a red bar at the top.

brookslogan:
Looks good, but before approving & merging, checking about provenance.  (Certain themes are exclusive to tidyverse or some other universe/org.)

brookslogan (approved):
Provenance: epipredict styling copied here was by @dajmcdon following https://pkgdown.r-lib.org/articles/customise.html.

brookslogan:
Because `dev` currently has a 4-digit version number, merging this will likely only update the dev site... which appears not to have been built successfully yet; [the expected landing page](https://cmu-delphi.github.io/epiprocess/dev) gives a 404.  Maybe merging will give it a chance to build with the epidatr sync PR.

brookslogan:
Oh, dev didn't exist because #383 wasn't finished yet.  No wonder there were no conflicts.

PR #383, opened 2023-11-30
Update pkgdown actions based on branch reorganization

brookslogan:
Goal is to have separate main & dev documentation sites, with different styling, like cmu-delphi/epidatr.  Plus some miscellaneous small updates thrown in.  [Before, we only would run pkgdown checks on PRs to dev, not render&update the actual site on pushes to dev, since it would clobber the main version of the site.]


nmdefries (changes requested):
~The DESCRIPTION file on `dev` needs a 4-component version number (x.x.x.x) to trigger the dev-site behavior. Without this, the `dev` site will keep overwriting the `main` public version. The `main` version needs to be a 3-component number (x.x.x). Some more info and links in [this PR](https://github.com/cmu-delphi/epidatr/pull/215).~

4-component versions complicates automated version bumping, if that's set up in this repo. That will need to be addressed in any workflow files, or noted in any instructions for manual releases.

Recommend including info somewhere about how this works on the public site and how to generate local previews. This is the [text](https://github.com/cmu-delphi/epidatr/blob/2faf053a727de4290a2850251f6102f641770540/DEVELOPMENT.md#developing-the-documentation-site) I included for developers of `epidatr`.

dshemetov:
Ah I see the version bump happened in #377

dshemetov (approved):
should be good to go after #385

brookslogan:
Thanks @dshemetov for addressing the issues.  Are we still missing release process documentation as mentioned by @nmdefries?  If so, should we put that in another Issue or handle it here?

dshemetov:
Fine to handle it here. Let's also take care of the 0.7.1 release here too.

- [ ] Add Nat's dev site docs documentation.
- [ ] Add a release note for 0.7.1
- [ ] Bump version to 0.7.1
- [ ] Merge to main
- [ ] Release

dajmcdon (approved):

nmdefries (approved):

PR #384, opened 2023-11-30
docs: update news for 0.7.0 release

dshemetov:
Since `main` is already on 0.7.0 and it's stable-enough, we should make it a release. This updates the NEWS with a few bullets from the release notes generated by the current `main` branch. 

After we merge this, we can make a release 0.7.0, then bump the dev branch to 0.7.0.9000 (see Nat's comment in #383) and starting adding PR change lines under that heading.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -52,10 +52,10 @@ inter-release development versions will include an additional ".9000" suffix.
   format (see `tidyr::chop`).
 * `epi_slide` now works properly with slide computations that output just a
   `Date` vector, rather than converting `slide_value` to a numeric column.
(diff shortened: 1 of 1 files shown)


brookslogan:
That check failure's probably bad.  Think we need a version range on epidatr.

dshemetov:
The check is failing because this doesn't have the ds/clean fixes in it, those are on dev.

brookslogan:
Okay, 0.7.0 should actually include ds/clean then?

dshemetov:
yea, doesn't seem worth trying to perfect 0.7.0. we should just try to get to a 0.8.0 release soon.

PR #385, opened 2023-11-30
sync: main -> dev

dshemetov:
This should be the final time we need to do this (after this all merges will be dev -> main), just catching up dev with a few 0.7.0 release notes.


dshemetov:
The CI for this merge is actually passing, not sure why the top 2 CI actions (which are for ds/0.7.0) are showing up here.

brookslogan (approved):
Failing checks are from "push", which I assume is triggered from merging the 0.7.0 doc update PR which would cause failing checks and a failure of the main documentation site to be updated.  Not sure how they're being referenced here either though.

We probably want to release a 0.7.1 with just ds/clean so the main site will actually get updated & will actually have running examples with the latest epidatr.

PR #386, opened 2023-12-04
Speedups for `epix_slide`

nmdefries:
Speeds `epix_slide` up ~2x on a series of test cases.


R/epi_df.R
```diff
@@ -159,7 +161,8 @@ new_epi_df = function(x = tibble::tibble(), geo_type, time_type, as_of,
   
   # Reorder columns (geo_value, time_value, ...)
   if(sum(dim(x)) != 0){
-    x = dplyr::relocate(x, "geo_value", "time_value")
+    # Moves columns in place
+    setcolorder(x, c("geo_value", "time_value"))
```
brookslogan, line 165: `new_epi_df` is part of our copy-on-write-compatible interface, so I don't think we can use `setcolorder` here without potentially introducing surprises in user code.

R/grouped_epi_archive.R
```diff
@@ -357,15 +360,15 @@ grouped_epi_archive =
...
+                dplyr::group_by(as_of_df, !!!syms(private$vars),
                                 .drop=private$drop) %>%
                   dplyr::group_modify(group_modify_fn,
                                       f = f, ...,
                                       ref_time_value = ref_time_value,
                                       new_col = new_col,
                                       .keep = TRUE)
               )
-            })
+            }) %>% rbindlist()
```
brookslogan, line 371: Nice speedup!  Is this going to change the output class of our result?  This might need to have a `setDF` (but check to make sure we can't possibly be aliasing something, e.g., in the 1-reference-time case).

R/epi_df.R
brookslogan: (All `epi_df` stuff is expected to have this interface.  We only have the dual interface for `epi_archive`.  But we want to get rid of it as well and have only a COW-compatible one.)

brookslogan:
These are nice cheap speedups.  FYI a couple other potential avenues at a much higher cost:
- Changing backend... I've checked out a few, but so far no improvements.  Might still need to check if some of these alternatives have explicit indexing options to speed them up.
- Partitioning DT based on the `ref_time_values` requested, taking each element of the partition and "squash" its updates using `unique fromLast=TRUE` (or a rolling join if it's faster), then apply those updates sequentially to build the snapshots sequentially.
- Parallelism stuff.  Conflicts with the sequential trick above.
- Maybe... partitioning DT by group or epikey before looping over the as-of logic (rather than while processing each snapshot), especially if rolling joins are involved.

These are probably recorded in another issue or issues somewhere.  All definitely seem more involved than these super-simple improvements, but if any catch your eye, feel free to try them!  (Probably separately, it will be nice to have these simple improvements first!)

R/grouped_epi_archive.R
```diff
@@ -280,16 +281,18 @@ grouped_epi_archive =
               if (! (is.atomic(comp_value) || is.data.frame(comp_value))) {
                 Abort("The slide computation must return an atomic vector or a data frame.")
               }
+              # Convert from data.frame to tibble for speed.
+              # Label every result row with the `ref_time_value`
+              res <- as_tibble(data.frame(time_value = ref_time_value))
```
brookslogan, line 286: IIRC `as_tibble` can also be a good amount of overhead in a tight loop.  Maybe see if this can be done after collecting the different results?

brookslogan: Also, maybe we should consider removing `%>%`s within loops.  Probably still can't use `|>` yet though.  Only time when our loops are tight enough that this might matter is in latency analysis and MAYBE lag extraction (though snapshot completion probably makes it not a tight loop in the latter case).

R/epi_df.R
nmdefries: What do you think about [this approach](https://github.com/cmu-delphi/epiprocess/pull/386/commits/e00dc125415b6cdc30a4e076a3786839d1bc820b)? It's about the same speed as `data.table::setcolorder`; I changed to `setcolorder` for brevity.

R/grouped_epi_archive.R
nmdefries: > maybe we should consider removing %>%s within loops

I can add this in. You're imagining changes like `select(...) %>% filter(...)` -> `filter(select(...), ...)` for now? At least where it doesn't impact readability.

nmdefries:
> other potential avenues

@brookslogan If there are other particular `epiprocess` usecases you'd like to be faster, I can profile and investigate those as well.

The above changes were based on a task from the examples,

```
 # dataset is 2000 obs, but individual changes were benchmarked on datasets of varying size (10-10k)
  dataset %>%
    group_by(geo_value) %>%
    epix_slide(f = ~ 5, before = 7) %>%
    ungroup()
```

R/grouped_epi_archive.R
nmdefries: I fixed the class aspect, but not sure about the aliasing. My attempt at investigating this using `lobstr::ref` to compare addresses of list of epi_dfs containing computed values and tibble of computed values after using rbindlist:

```
> data$DT

# Input data
geo_value sex time_value    version value
1:        al   M 2012-01-01 2012-01-01    25
2:        ca   M 2012-01-01 2012-01-01   100

> data %>%
  +   epix_slide(f = ~ 5, before = 0, ref_time_values = as.Date("2012-01-01")) %>% 
  +   ungroup()

# output from lobstr::ref(<list of epi_dfs containing computed values >, <tibble of computed values after using rbindlist>)
█ [1:0x559b7f330b08] <list> 
  └─█ [2:0x559b80ee4dd8] <tibble[,2]> 
  ├─time_value = [3:0x559b7f330b40] <date> 
  └─slide_value = █ [4:0x559b7f467918] <list> 
  └─[5:0x559b7f331160] <dbl> 
  
  █ [6:0x559b80ee37d8] <tibble[,2]> 
  ├─time_value = [7:0x559b7f4842f8] <date> 
  └─slide_value = █ [8:0x559b7f484288] <list> 
  └─[5:0x559b7f331160] 

# Output of epix_slide
# A tibble: 1 × 2
time_value slide_value
* <date>           <dbl>
  1 2012-01-01           5
```

R/grouped_epi_archive.R
```diff
@@ -366,6 +370,9 @@ grouped_epi_archive =
                                       .keep = TRUE)
               )
             })
+            x <- rbindlist(x) %>% setDF() %>% as_tibble() %>%
+              # Reconstruct groups
+              group_by(!!!groups(x[[1L]]), .drop = attr(attr(x[[1L]], "groups"), ".drop"))
```
nmdefries, line 375: Is there a better way to carry forward groups and `.drop` setting? This seems awkward.

nmdefries:
Odd suggestion. `dplyr::group_map` and `dplyr::group_split` look like

```
function (.data, .f, ..., .keep = FALSE) 
{
    lifecycle::signal_stage("experimental", "group_map()")
    UseMethod("group_map")
}
```

with the `lifecycle` bit taking a lot of time. After changes above, just the call to `signal_stage` takes ~1 of 8 s in a test call. This seems easy to remove, if we can get dplyr's other functions to use a modified version of these two functions. I'm assuming this is bad practice? Maybe not impactful enough to be worth it.

R/grouped_epi_archive.R
nmdefries: Oh, I guess I can access private$drop and private$vars here

R/grouped_epi_archive.R
nmdefries: > `as_tibble`... can be done after collecting the different results?

I can't find a nice way. I changed this section a bit to make it faster and, even though it still calls `as_tibble`, is about as fast in my benchmarking and profiling as a janky non-`as_tibble` approach. Will leave this as-is for now.

nmdefries:
Yeah, it's a bad idea... but tempting since the change is so small :grinning: We use `group_map` and `group_split` indirectly. I'm not entirely sure where they're coming from.

dshemetov:
It does seem like an easy change for ~10% speed boost! 

Hm, since they're used indirectly, does this mean loading epiprocess would have to overwrite those functions in the dplyr namespace?

nmdefries:
> epiprocess would have to overwrite those functions in the dplyr namespace?

Yep. I was hoping there could be a workaround, but unclear.

R/grouped_epi_archive.R
nmdefries: (Another approach creating a new `data.frame` every iteration was about as fast in benchmarking as the current version.)

R/grouped_epi_archive.R
brookslogan: Does this maintain the speed benefits?  `group_by` could potentially be costly.  I'll try to double-check the no-aliasing part.

nmdefries: > Does this maintain the speed benefits? group_by could potentially be costly.

Yes, it seems to. I was also wondering about the time for that group_by but it didn't seem to impact anything.

brookslogan: Great!  Regarding aliasing:
- Actually, we don't need to avoid it; we are fine with out outputs aliasing stuff I think.
- I've updated the docs to reflect the above and some other mutation&aliasing details.  Please sanity check.  (I hope we can eventually simplify & make all these notes internal by providing a filter method & hiding away the DT from the user.)
- We already probably don't ever alias.
  - `all_versions = FALSE` gets input that has gone through a DT -> tibble conversion using `as_tibble` with no `setDT`, which should prevent DT/column aliasing
  - `all_versions = TRUE` has a manual step making sure we don't alias the `DT` (I'm actually not sure why now... maybe I was thinking we were going to be mutating like in `fill_through_version`) but might potentially somehow alias the columns (I'm not sure if this is possible... but since the aliasing case of `unique` didn't appear to be documented before, we probably should assume it could change to alias columns instead of the DT).

R/grouped_epi_archive.R
brookslogan: > > `as_tibble`... can be done after collecting the different results?
> 
> I can't find a nice way. I changed this section a bit to make it faster and, even though it still calls `as_tibble`, is about as fast in my benchmarking and profiling as a janky non-`as_tibble` approach (could be different with a different test query, of course). Will leave this as-is for now.

Yes, something like that.  Or maybe, for easier readability, though I don't particularly like this pattern due to making debugging harder (and it might actually garbage-collect differently, due to a potentially higher refcount and potentially setting the "named" garbage collection flag whatever that does):
```r
tmp <- `select(...)`
tmp <- `filter(tmp, ...)`
```

brookslogan:
> @brookslogan If there are other particular epiprocess usecases you'd like to be faster, I can profile and investigate those as well.

I don't know yet.  I didn't expect `epix_slide` overhead to even be a deal for pseudoprospective forecasting since the fitting routines we were using back a while ago were pretty slow.  In the process of finishing up some chapters I will hopefully get a sense what it feels like for lag extraction and might ask you to profile if it seems slow.  There's a lot of specialized stuff we could do for lag extraction, though; it doesn't have to be a slide operation necessarily.

Re. `group_map` and `group_split`...  If we could refactor to call them directly, then we could also easily refactor to call our own version of `group_map` and `group_split` that `UseMethod` without the lifecycle call, though this does have some chance of breakage if they decide to move some pre/post-processing into the generic.  But we also might want to try to "group split" the archive on the "outside" and loop over ref_time_values on the "inside" like we do in `epi_slide`, rather than the other way around; this seems like it could potentially avoid some unneeded sorts by `group_by` inside the `ref_time_value` loop, though it might be hard to determine beforehand which approach incurs more function call overhead.

R/grouped_epi_archive.R
nmdefries: Thanks for adding all of those comments, they look good.

I'll remove some pipes as a final change.

R/grouped_epi_archive.R
brookslogan: Another approach?: I think there may actually be something like `new_tibble` or `new_data_frame` or other related constructs for disabling some validation.

nmdefries: So the alternate approach I came up with was

```
df <- data.frame(time_value = ref_time_value)
df[[new_col]] <- I(list(comp_value_raw))
return(df)
```

and then do the rbindlist -> as_tibble outside the loop. So there is no tibble conversion inside the loop and we still get our loop output col in list format. At least in benchmarking, this wasn't quite as fast as what I have in the code now (list -> as_tibble in loop), but that wouldn't necessarily be true in real calls. `new_tibble` is faster, but do we then need to run `validate_tibble` as well?

nmdefries: (At this point, this step isn't taking that much time, esp compared to the group_by-group_modify in the loop, so we don't need to think about it super in-depth)

nmdefries: Added `validate_tibble` in for safety. It makes using `new_tibble` slower, but still faster than other methods.

nmdefries:
@brookslogan if you review this before I'm back and everything looks good, feel free to merge.

R/grouped_epi_archive.R
brookslogan: Did removing pipes actually make a difference in your testing?  I just tried re-introducing magrittr pipes to what I thought would be the highest-traffic pipe operations, the `group_by` and `group_modify` inside the `FUN` arg to `lapply`, but don't see any real change.

brookslogan: Also it'd be nice if you could maybe summarize in general what seemed like the highest-impact changes here or any other places you think might benefit significantly from optimization.

brookslogan (approved):
Looks good to me!  Have some trailing requests regarding documenting whether the de-pipe-ing was worth it, and which optimizations had the most impact, but this seems ready to merge.

R/grouped_epi_archive.R
nmdefries: I don't see that de-piping changed runtime. Theoretically, extra function calls will always take [more time](https://www.r-bloggers.com/2011/10/overhead-cost-of-a-function-call/), but the cost is just so low compared to everything else we're doing here.

On the other hand, not using pipes does make profiles easier to read. Using pipes somehow causes functions to be marked as not having a sensible parent call, so e.g. `group_modify` shows up in a spot on the profile that you don't expect and that doesn't match the code.

nmdefries:
Slow bits in epix_slide from a test call. 23.5 s is the baseline runtime.

- [done] 25% of time (5.8 of 23.5 s) spent in x$slide -> map_dfr -> bind_rows
- [done] 5% (1.2 of 23.5 s) spent in as_epi_df.tbl_df -> new_epi_df -> relocate.data.frame 
- [done] ?? spent in group_by-across-tidyselect::all_of
- [probably intractable] 16% of time (3.7 of 23.5 s) spent in group_modify.grouped_df -> group_map -> lifecycle::signal_stage or group_modify.grouped_df -> group_map.data.frame -> group_split -> lifecycle::signal_stage
- [intractable] 25% of time (6 of 23.5 s) spent in `signal` from group_modify.epi_df -> dplyr_reconstruct -> dplyr_reconstruct.grouped_df -> grouped_df -> compute_groups -> signal and group_split -> signal and other places (apparently a logging feature, not necessary for production https://github.com/tidyverse/dplyr/pull/4751/files ). Too deep in `dplyr` to change.

It was hard to tell in [Dmitry's tests profile](https://github.com/cmu-delphi/exploration-tooling/issues/60) which functions were coming from `epiprocess` (pipe-caused "unaffiliation" might be part of the problem). I could tell that epix_slide -> anonymous -> purrr::map_dfr -> group_by was slow, and  epi_df.R::new_epi_df -> relocate.data.frame was slow. Besides those, there were a lot of dplyr functions that were slow but not clearly coming from `epiprocess` (unnest.data.frame, arrange.data.frame, rename.data.frame, group_by.data.frame, select.data.frame, mutate.data.frame).

nmdefries:
Code to profile, and output of baseline profile.

[baseline_profile_example.tar.gz](https://github.com/cmu-delphi/epiprocess/files/13678558/baseline_profile_example.tar.gz)

nmdefries:
Remaining slow parts are doing `group_by` and `group_modify` on different objects.

Screenshot from 2023-12-14 16-40-18

[20231214_profile.tar.gz](https://github.com/cmu-delphi/epiprocess/files/13678705/20231214_profile.tar.gz)

We can't just swap in a different function here, so like Logan was saying above ([here](https://github.com/cmu-delphi/epiprocess/pull/386#issuecomment-1839770865) and [here](https://github.com/cmu-delphi/epiprocess/pull/386#issuecomment-1851158312)), this gets into more demanding changes. We should figure out what particular types of requests/computations are slow and how much slower than desired before spending a lot of time on a rewrite.

PR #310, opened 2023-05-04
Round out dplyr&tidyr functionality for grouped `epi_df`s

brookslogan:
This is a work in progress and is currently buggy.  See related thread around [this comment](https://github.com/cmu-delphi/epiprocess/issues/223#issuecomment-1524587983) for bugs and potential ways forward.

Issue #186, opened 2022-08-05
"Promote" `other_keys` to be printed, a constructor parameter, and more clearly documented


brookslogan:
Migrated from #113 and #183.
- `print.epi_df` should print the `other_keys` metadata.
- `other_keys` should be its own parameter to `new_epi_df` and `as_epi_df`
- Unfortunately, I don't think we ever explained what "keys" were in the context of an `epi_df`, and googling "tibble key" is surprisingly unhelpful (currently giving links to lock picks and decoder tools) [so the `{new,as}_epi_df` docs may need explication].  Maybe explaining this could just be rolled into the issue for adding `other_keys` as a parameter, though; it'd be easier to do in its own `@param`.

If this is causing recurring issues in epipredict maybe it should be changed to higher priority.

brookslogan:
We should probably also make `other_keys=character(0)` the default.

brookslogan:
And we should also print the rest of the epikey in `epi_archive`'s print method as well (e.g., after the geo and time type).

brookslogan:
We need to regenerate the saved data sets after this change, or maybe right now due to the `other_keys` default change, or add notes in implementation to try to maintain backwards compatibility with saved edf's, or...

dsweber2:
When "you say after this change", are you referring to the default of `other_keys` switching to `character(0)`?

PR #390, opened 2024-01-09
patch: get select working with grouping

### LINKED ISSUE #388: `select` causes type conversion
Calling `select(epi_data, -col_name)` on some `epi_df` converts it into a tibble. There are probably related issues, since this is only one of many dplyr verbs to support.

dsweber2:
Straightforward patch of `select.epi_df` so that the grouped version doesn't drop the `epi_df` type when selecting. closes #388.


dshemetov:
@brookslogan does this overlap #310 work?

brookslogan:
Yes, this overlaps with #310, but while #310 would lend itself to something more general/proper in some cases, it's probably good to just do the mass-S3-implementation approach for now.

Something for `select` in particular: we can't just restore the old metadata as we may have dropped keys. We need to preserve the sort of stuff that goes on in `[.epi_df` and `names<-.epi_df`.  I realize now it may not be as simple as some of the other ones with just a `dplyr_reconstruct` or `reclass` whatever.

brookslogan:
@dsweber2 I think the R/ folder changes still need to be pushed?

dsweber2:
lol, whoops. I forgot to push the new file. I was trying to separate out the patch-style methods from the proper/more permanent ones.

Also, there's a ton of warnings which I don't think I caused, and the package has no styler, which means my commits may come with random formatting. Sorry about that.

> [degrading if we drop required columns, adjusting other_keys if we drop or rename them]

Good point, I'll add some tests for this and adjust. Though the times I've actually seen `select` used in our code, it was for dropping a data-column.

tests/testthat/test-epi_df.R
```diff
@@ -36,6 +36,22 @@ test_that("as_epi_df errors when additional_metadata is not a list", {
...
+    x = 1:10, y = 1:10,
+    time_value = rep(seq(as.Date("2020-01-01"), by = 1, length.out = 5), times = 2),
+    geo_value = rep(c("ca", "hi"), each = 5)
+  )
+
+  epi_tib <- epiprocess::new_epi_df(tib)
+  epi_tib
+  grouped_epi <- epi_tib %>% group_by(geo_value)
+  selected_df <- grouped_epi %>% select(-y)
+  expect_true("epi_df" %in% class(selected_df))
```
dshemetov, line 56: Can we add a test that makes sure whatever metadata epi_df adds is preserved?

dshemetov (commented):
Just one test request. Are you planning to do more in this PR?

brookslogan:
Also, here or in a separate PR, it'd be nice to have a `complete` that doesn't drop `epi_df`ness?

tests/testthat/test-epi_df.R
dsweber2: yeah, that's reasonable. I am going to add stuff to address the edge-cases Logan was talking about

dsweber2:
I'd rather this not blossom into a weeks long project. I made an issue #391 for patching more generally; if any feel particularly important feel free to rearrange/add some.

dshemetov:
Probably remove the list from the OP so it doesn't seem like this PR is aiming to check all those boxes.

dsweber2:
So I added some tests for the cases you were talking about @lcbrooks, and it seems like whoever implemented `dplyr_reconstruct` covered all these bases, since it passed the tests without me changing the core code.

brookslogan:
An edge case to fix or file off:
```r
tibble::tibble(geo_value = 1, time_value = 1, age = 1, value = 1) %>%
  as_epi_df(additional_metadata = list(other_keys = "age")) %>%
  select(geo_value, time_value, age_group = age) %>%
  attr("metadata")
#> $geo_type
#> [1] "hhs"
#> 
#> $time_type
#> [1] "custom"
#> 
#> $as_of
#> [1] "2024-01-12 16:48:48 PST"
#> 
#> $other_keys
#> character(0)
```

brookslogan:
Might be a regression from `dplyr_reconstruct`ing in `names<-.epi_df`.  Taking a look.

brookslogan:
Okay, I think I fixed `names<-.epi-df`.  The problem is that it doesn't propagate through correctly via `select`.  Now, based on `?dplyr_extending`, you'd expect we could do something like:
```
select.epi_df <- function(.data, ...) {
  selected <- NextMethod(.data)
  might_decay <- reclass(selected, attr(selected, "metadata"))
  return(dplyr_reconstruct(might_decay, might_decay))
}
```
but that doesn't seem sufficient.  It seems that `select.data.frame`'s `out <- set_names(out, names(loc))` just destroys our attributes when we are grouped, preventing us from seeing what was renamed.  So we may need to do some `eval_select`ing ourselves.

brookslogan (approved):
Okay, so I kind of tricked you with some pre-existing bugs and missing tests + buggy wrappers along these lines.  We can't always reconstruct using the input's metadata; `dplyr_reconstruct.epi_df` tries to account for things like dropped other key columns, but can't detect a rename.  I think I've fixed it up, along with some of the pre-existing things masking the problem, and some other stuff to make tests pass (inconsistencies between `NULL` vs. `character(0L)` for representing no other keys).

Please sanity-check these updates & merge.

dajmcdon:
See also @brookslogan's comment on #186 about the `character(0)` default.

PR #394, opened 2024-01-13
Make epi_slide faster

nmdefries:
About 60% faster. Turns group keys into a list for quick lookup, but just pastes together values, so not potentially not robust to odd/complex objects used as keys. `deparse` then `paste` should be more robust (but will take a little longer).


nmdefries:
@dsweber2 if your group keys are well-behaved, this might work as-is as a quick fix.

nmdefries:
Superseded by https://github.com/cmu-delphi/epiprocess/pull/397.

PR #397, opened 2024-01-17
Step through pre-calculated start times for each group using closure rather than using `.real` col in `epi_slide`

nmdefries:
Instead of re-calculating `.ref_time_value`s, use pre-calculated values stored in `starts`. For each group, use a closure to keep track of position in `starts` vector. This avoids the [slow `.real` filtering and removal](https://github.com/cmu-delphi/epiprocess/pull/318#issuecomment-1560215043) steps, and simplifies the code by removing various bits of `.real` handling.

This is ~4x faster than the old version.


nmdefries:
Seems to be a problem with the new `withr` (3.0.0, released yesterday).

nmdefries:
All the `man/*.Rd` files are from styling changes.

R/slide.R
```diff
@@ -230,37 +230,15 @@ epi_slide <- function(x, f, ..., before, after, ref_time_values,
...
-  before_time_values_df$.real <- FALSE
-
-  x <- bind_rows(before_time_values_df, x)
-
   # Arrange by increasing time_value
   x <- arrange(x, time_value)
 
   # Now set up starts and stops for sliding/hopping
-  time_range <- range(unique(x$time_value))
+  time_range <- range(unique(c(x$time_value, min_ref_time_values_not_in_x)))
```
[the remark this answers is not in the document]
brookslogan: Okay, I tried addressing the above + some other random things [here](https://github.com/cmu-delphi/epiprocess/pull/399).  But current implementation looks okay as-is.

R/slide.R
```diff
@@ -394,14 +379,5 @@ epi_slide <- function(x, f, ..., before, after, ref_time_values,
...
 
-  # Remove any remaining phony observations. When `all_rows` is TRUE, phony
-  # observations aren't necessarily removed in `slide_one_grp`.
-  if (all_rows) {
-    x <- x[x$.real, ]
-  }
-
-  # Drop helper column `.real`.
-  x$.real <- NULL
-
```
brookslogan, line 405: Requests regarding other files:
- Update `test-epi_slide.R` "Also checks that we correctly remove extra rows and columns (`.real`) used [...]" comment.
- Maybe review & merge the [`epi_slide` tidying PR](https://github.com/cmu-delphi/epiprocess/pull/399) into this one?
- Bump version in DESCRIPTION, add entry to NEWS.md, merge.

brookslogan (approved):
Looks good, please see minor TODOs above.

Nice 4x!!!  Was this on something like 7-day averaging?  [I'm getting something like a 2x doing `jhu_csse_daily_subset %>% group_by(geo_value) %>% epi_slide(before = 6, cases_7davish = mean(cases))`, 3x on an ungrouped version.  Still super nice.]

nmdefries:
My (main) test case was indeed a 7-dav, calling the `exploration-tooling` [rolling mean fn](https://github.com/cmu-delphi/exploration-tooling/blob/c6913c91826d27acaabe6a373881d79f09537f81/R/data_transforms.R#L38-L47) on a synthetic dataset,

```
n_days <- 4000
removed_date <- 10
simple_dates <- seq(as.Date("2012-01-01"), by = "day", length.out = n_days)
simple_dates <- simple_dates[-removed_date]
rand_vals <- rnorm(n_days - 1)

# Three states, with 2 variables. a is linear, going up in one state and down in the other
# b is just random
# note that day 10 is missing
epi_data <- epiprocess::as_epi_df(rbind(tibble(
  geo_value = "al",
  time_value = simple_dates,
  a = 1:(n_days - 1),
  b = rand_vals
), tibble(
  geo_value = "ca",
  time_value = simple_dates,
  a = (n_days - 1):1,
  b = rand_vals + 10
), tibble(
  geo_value = "fl",
  time_value = simple_dates,
  a = (n_days - 1):1,
  b = rand_vals * 2
)))
```

Surprisingly, 

```
jhu_csse_daily_subset %>% group_by(geo_value) %>% epi_slide(before = 6, ~mean(.x$cases), new_col_name = "cases_7davish")
```

is twice as fast as

```
jhu_csse_daily_subset %>% group_by(geo_value) %>% epi_slide(before = 6, cases_7davish = mean(cases))
```

It looks like converting a quosure to fn is the bottleneck, with [this line](https://github.com/cmu-delphi/epiprocess/blob/b95685ee5251e8c994be1907cf5283a185557f3f/R/utils.R#L296) being particularly slow.

R/slide.R
nmdefries: I thought version bumping only happens on merge to `main`.

brookslogan: True... Eventually on release we will smash these down but it does look like they will be awkward in the mean time.  Some form of automation of the version bump etc. also sounds nice.  We could just do whatever for this PR and fix it up after deciding on an appropriate scheme.

R/slide.R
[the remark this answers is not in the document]
nmdefries: Thanks for further cleaning up this section! I'm looking at that PR now.

PR #398, opened 2024-01-18
Versioning Proposal

dsweber2:
This implements the github actions to make  `0.<increment on release>.<increment on PR>` usable by making the dev docs on the `dev` branch, and the release docs on the `main` branch. Along with this, I suggest that we make `main` the default branch, so that people have to opt-in to the dev version, rather than opt-out of it.

It also introduces a semver check to make sure that any PR includes a bump to the semantic version. Hypothetically this should fail on this PR for the first commit, to be fixed after confirming this (if only there were a way to test gh actions).


nmdefries:
Drive by comment: Thanks for adding these actions!

> if only there were a way to test gh actions

I hear there are some tools that let you emulate GH Actions environment locally.

>  I suggest that we make main the default branch

Doesn't this cause problems with PRs? e.g. opening a PR will default to being against `main` and could lead to accidentally merging changes into `main`.

> semver check to make sure that any PR includes a bump to the semantic version

Propose having an action to bump the version (from `dev`) on PR merge to dev. I've already run into the problem where my PR and another bumped the version to the same number, so I then had to re-bump. This also causes the changelog to be hard to maintain for the same reason.

dsweber2:
Ok, so I think I got the auto-version bumper working (hard to tell without merging this, but it was successfully bumping the version *every* commit before I switched that to only happen on `dev` and not every branch).

It's also set to auto-generate a tag for minor and major releases, and to do minor releases on any commit on `main`, and major releases on a `release` branch; we may want to mess with the settings there and make that one manual.

dsweber2:
>> I suggest that we make main the default branch

> Doesn't this cause problems with PRs? e.g. opening a PR will default to being against main and could lead to accidentally merging changes into main.

I guess it depends on if we're prioritizing user-experience, so that by default people get a more stable version, or dev experience, so we don't have to do some git history magic if we mess up. Doesn't look like github has a way to set a PR-only default branch, unfortunately. Just a thought while I was working on things that distinguish a "release" before we're on CRAN.

dshemetov:
We can get a good dev experience and a good user experience by keeping `dev` the default branch, but recommend users install from the `main` tag (which we already do in Getting Started `devtools::install_github("cmu-delphi/epiprocess", ref = "main")`).

dshemetov:
> Propose having an action to bump the version (from dev) on PR merge to dev. I've already run into the problem where my PR and another bumped the version to the same number, so I then had to re-bump. This also causes the changelog to be hard to maintain for the same reason.

@nmdefries @dsweber2 this is addressed now right (we're only bumping on merge to dev now)?

dshemetov:
Can we / should we automate merging back to dev from main? That will probably trigger a version bump in dev, but I'm fine with always starting on `x.y.1`.

dsweber2:
> Propose having an action to bump the version (from dev) on PR merge to dev

Is addressed by the fact that the version bump applies at merge time on the `dev` branch, rather than in the PR. Which kind of bump depends on whether its on `dev` or `main`.

> Can we / should we automate merging back to dev from main?

Oh, yeah this is probably a good idea; I guess we can just copy the action from a different repo?

dshemetov:
There's [this](https://github.com/cmu-delphi/delphi-epidata/blob/8a476fd08f5aef6913fb3d231f1fcceb691953be/.github/workflows/release-helper.yml#L129-L153), but I don't see a reason why we couldn't just merge back automatically without a PR.

.github/workflows/pkgdown_dev.yaml
```diff
@@ -40,7 +40,7 @@ jobs:
       - name: Build site
         env:
           DELPHI_EPIDATA_KEY: ${{ secrets.SECRET_EPIPROCESS_GHACTIONS_DELPHI_EPIDATA_KEY }}
-        run: pkgdown::build_site_github_pages(new_process = FALSE, install = FALSE)
+        run: pkgdown::build_site_github_pages(new_process = FALSE, install = FALSE, override=list(PKGDOWN_DEV_MODE="devel"))
```
nmdefries, line 43: issue (nonblocking): For maintenance, prefer to have both `dev` and `main` behavior in the same workflow. We should be able to set `PKGDOWN_DEV_MODE="devel"` or `PKGDOWN_DEV_MODE="release"` conditionally on the push/pull branch.

This is untested:

```suggestion
        run: |
            if (startsWith("${{ GITHUB_EVENT_NAME }}", "pull_request")) {
                mode <- ifelse("${{ GITHUB_BASE_REF }}" == "main", "release", "devel")
            } else {
                mode <- ifelse("${{ GITHUB_REF_NAME }}" == "main", "release", "devel")
            }
            pkgdown::build_site_github_pages(new_process = FALSE, install = FALSE, override=list(PKGDOWN_DEV_MODE=mode))
```

[GH actions var reference](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)

.github/workflows/release-helper.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Create pull request into dev
+        uses: peter-evans/create-pull-request@v3
+        with:
+          branch: bot/sync-main-dev
+          commit-message: "chore: sync main-dev"
+          base: dev
+          title: "chore: sync main->dev"
+          labels: chore
+          body: |
+            Syncing Main->Dev.
```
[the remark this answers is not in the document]
dsweber2: Not sure the release workflow is worth the effort to actually get it set up, given that that one is specific to python code, but good to know it exists

.github/workflows/version_bump_options.json
```diff
@@ -0,0 +1,26 @@
...
+      "trigger": "commit",
+      "bump": "minor",
+      "branch": "main",
+      "tag": true,
+      "reset": "build"
+    },
+    {
+      "trigger": "commit",
+      "bump": "major",
+      "branch": "release",
```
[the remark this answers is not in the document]
dsweber2: Honestly I probably should've just skipped this; I expect 1.0 to be a CRAN release and a process we will want to revisit at the time, no matter what I'd written here. I'm just gonna drop it and we can revisit later.

.github/workflows/pkgdown_dev.yaml
dsweber2: Not sure why `if (startsWith("${{ GITHUB_EVENT_NAME }}", "pull_request"))` exists, seems to be the same in either case.

.github/workflows/release-helper.yaml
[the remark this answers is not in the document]
nmdefries: Agree about maybe not worth the effort, butT that release workflow is for R -- python is just used to do the bumping. (`forecast-eval` doesn't have any python code.)

.github/workflows/pkgdown_dev.yaml
nmdefries: If it's a pull request, you have to get the branch name from a different variable (GITHUB_BASE_REF vs GITHUB_REF_NAME)

nmdefries: There might be a better way to do this conditional logic, too. This is just my initial thought on how to approach it.

dsweber2: gotcha, didn't read closely enough. Seems like it should work.

nmdefries (approved):
Looks good!

PR #405, opened 2024-01-25
patch version bumper for branch permissions

dsweber2:
not sure if this is quite the right way. I also added a repo secret `SEMVER_BUMPER_KEY`, which has both a [secret](https://github.com/cmu-delphi/epiprocess/settings/secrets/actions) and a [deploy key](https://github.com/cmu-delphi/epiprocess/settings/keys) with write access

### DIFF
--- a/.github/workflows/auto_semver_bump.yaml
+++ b/.github/workflows/auto_semver_bump.yaml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
       # Checkout action is required
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3.5.2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/auto_semver_bump.yaml
```diff
@@ -12,7 +12,9 @@ jobs:
 
     steps:
       # Checkout action is required
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3.5.2
+        with:
+          ssh-key: ${{ secrets.SEMVER_BUMPER_KEY }}
```
dshemetov, line 17:
```suggestion
          token: ${{ secrets.DMITRY_PAT }}
```

PR #409, opened 2024-01-26
Test autoplot and arg-checker functions

nmdefries:


dshemetov:
Woops, I just saw you made this PR after merging #410, my bad. I removed some of the functions you're testing here, sorry to null that work but on the plus side we don't need to unit test checkmate functions.

nmdefries:
I was expecting the issue to use checkmate functions to disappear into the issue void :P Would have been less work for everyone if I'd started there instead of the tests.

dajmcdon (approved):

dshemetov:
Hehe yea, reasonable :) I was particularly excited about that issue cause checkmate was very satisfying to implement in epidatr!

PR #410, opened 2024-01-26
refactor: use checkmate checks

### LINKED ISSUE #403: deprecate internal arg_check funs in favour of `{checkmate}`?
suggestion (optional): Looks like these type-check fns could be replaced by `checkmate::assert_double` or `checkmate::check_double` if you want to raise a custom error message, and analogues for other types.

_Originally posted by @nmdefries in https://github.com/cmu-delphi/epiprocess/pull/382#discussion_r1462207988_

dshemetov:
### Checklist

Please:

        brookslogan, nmdefries.

Removes custom arg checks and uses checkmate. This one probably doesn't require a NEWS line.

-   Resolves #403


R/autoplot.R
```diff
@@ -25,19 +25,21 @@
 #' @return A ggplot object
 #' @export
 #'
+#' @importFrom checkmate assert assert_int anyInfinite assert_character
```
dajmcdon, line 28: One minor suggestion: rather than putting these imports with the function they're used in, maybe better to put them all in a common location? That way we don't end up importing the same one in the Roxygen 15 different times.

dajmcdon (approved):

PR #412, opened 2024-01-30
refactor: unclutter epi_archive print method

### LINKED ISSUE #341: Unclutter `epi_archive`'s `print` method + add preview of `DT`
Pretty commonly, we'll print an `epi_archive` `x`, see a lot but not very useful, then print `x$DT` to get a better idea what things look like.  We should rebalance the `print` output to include the `DT`, exclude any redundant information, and move some other information into a `summary` method.  Then we might be able to make `DT` a private field without much issue, and avoid some worries about aliasing and mutation and `key` invalidation.

dshemetov:
### Checklist

Please:

        brookslogan, nmdefries.

- Removed some metadata prints and combined a few others into a single line
- Added a line that prints `key(DT) - c("geo_value", "time_value", "version")` 
- Added a `x$DT %>% print` at the end

-   Resolves #341


R/methods-epi_archive.R
```diff
@@ -1000,3 +1000,79 @@ epix_truncate_versions_after.epi_archive <- function(x, max_version) {
...
+#'
+#' @param x The `epi_archive` object.
+#' @param class Boolean; whether to print the class label header
+#' @param methods Boolean; whether to print all available methods of
+#'   the archive
+#' @param ... Additional arguments passed to methods.
+#'
+#' @method print epi_archive
+#' @export
+print.epi_archive <- function(x, class = TRUE, methods = TRUE, ...) {
```
dshemetov, line 1014: Sample print:

```r
r$> df <- data.frame(
    geo_value = "ca",
    time_value = as.Date("2020-01-01"),
    version = as.Date("2020-01-01") + 0:19,
    value = 1:20
  )
r$> ea1 <- as_epi_archive(df, compactify = FALSE)
r$> ea1
An `epi_archive` object, with metadata:
* geo_type  = state
* time_type = day
----------
* min time value = 2020-01-01
* max time value = 2020-01-01
* first version with update = 2020-01-01
* last version with update = 2020-01-20
* No clobberable versions
* versions end   = 2020-01-20
----------
Data archive (stored in DT field): 20 x 4
Columns in DT: geo_value, time_value, version, value
----------
Public R6 methods: initialize, as_of, fill_through_version, truncate_versions_after, 
                   merge, group_by, slide, clone
# A tibble: 20 × 4
   geo_value time_value version    value
   <chr>     <date>     <date>     <int>
 1 ca        2020-01-01 2020-01-01     1
 2 ca        2020-01-01 2020-01-02     2
 3 ca        2020-01-01 2020-01-03     3
 4 ca        2020-01-01 2020-01-04     4
 5 ca        2020-01-01 2020-01-05     5
 6 ca        2020-01-01 2020-01-06     6
 7 ca        2020-01-01 2020-01-07     7
 8 ca        2020-01-01 2020-01-08     8
 9 ca        2020-01-01 2020-01-09     9
10 ca        2020-01-01 2020-01-10    10
11 ca        2020-01-01 2020-01-11    11
12 ca        2020-01-01 2020-01-12    12
13 ca        2020-01-01 2020-01-13    13
14 ca        2020-01-01 2020-01-14    14
15 ca        2020-01-01 2020-01-15    15
16 ca        2020-01-01 2020-01-16    16
17 ca        2020-01-01 2020-01-17    17
18 ca        2020-01-01 2020-01-18    18
19 ca        2020-01-01 2020-01-19    19
20 ca        2020-01-01 2020-01-20    20
```

R/methods-epi_archive.R
```diff
@@ -1000,3 +1000,79 @@ epix_truncate_versions_after.epi_archive <- function(x, max_version) {
...
+  # --- Copied from R6 ---
+  if (class) cat("An `epi_archive` object, with metadata:\n")
+  cat(sprintf("* %-9s = %s\n", "geo_type", x$geo_type))
+  cat(sprintf("* %-9s = %s\n", "time_type", x$time_type))
```
dshemetov, line 1018: Do we need `geo_type` and `time_type`?

R/methods-epi_archive.R
```diff
@@ -1000,3 +1000,79 @@ epix_truncate_versions_after.epi_archive <- function(x, max_version) {
...
+      cat(sprintf("* %-9s = %s\n", names(m), m))
+    })
+  }
+  cat("----------\n")
+  if (length(x$DT$time_value) == 0 || all(is.na(x$DT$time_value))) {
+    min_time <- max_time <- NA
+  } else {
+    min_time <- Min(x$DT$time_value)
+    max_time <- Max(x$DT$time_value)
+  }
```
dshemetov, line 1030: Do we need to print min/max time statistics?

R/methods-epi_archive.R
```diff
@@ -1000,3 +1000,79 @@ epix_truncate_versions_after.epi_archive <- function(x, max_version) {
...
+  cat(sprintf("* %-14s = %s\n", "min time value", min_time))
+  cat(sprintf("* %-14s = %s\n", "max time value", max_time))
+  cat(sprintf(
+    "* %-14s = %s\n", "first version with update",
+    min(x$DT$version)
+  ))
+  cat(sprintf(
+    "* %-14s = %s\n", "last version with update",
+    max(x$DT$version)
+  ))
```
dshemetov, line 1040: Do we need min/max version statistics?

R/methods-epi_archive.R
```diff
@@ -1000,3 +1000,79 @@ epix_truncate_versions_after.epi_archive <- function(x, max_version) {
...
+  } else {
+    cat(sprintf(
+      "* %-14s = %s\n", "clobberable versions start",
+      x$clobberable_versions_start
+    ))
+  }
+  cat(sprintf(
+    "* %-14s = %s\n", "versions end",
+    x$versions_end
+  ))
```
dshemetov, line 1052: (Don't know what clobberable versions means) Do we need these?

R/methods-epi_archive.R
```diff
@@ -1000,3 +1000,79 @@ epix_truncate_versions_after.epi_archive <- function(x, max_version) {
...
+  cat("----------\n")
+  cat(sprintf(
+    "Data archive (stored in DT field): %i x %i\n",
+    nrow(x$DT), ncol(x$DT)
+  ))
+  cat(sprintf("Columns in DT: %s\n", paste(ifelse(length(
```
dshemetov, line 1058: Do we need DT shape information?

R/methods-epi_archive.R
```diff
@@ -1000,3 +1000,79 @@ epix_truncate_versions_after.epi_archive <- function(x, max_version) {
...
+    length(colnames(x$DT)[5:length(colnames(x$DT))]), "more columns"
+  )
+  ))))
+  if (methods) {
+    cat("----------\n")
+    writeLines(wrap_varnames(
+      initial = "Public R6 methods: ",
+      names(epi_archive$public_methods)
+    ))
+  }
```
dshemetov, line 1072: This can probably go, as we're moving away from R6, yea?

R/methods-epi_archive.R
```diff
@@ -1000,3 +1000,79 @@ epix_truncate_versions_after.epi_archive <- function(x, max_version) {
...
+
+  # Error in NextMethod() : generic function not specified
+  # NextMethod()
+
+  x$DT %>% tibble %>% print
```
brookslogan, line 1077: Generally favor `as_tibble` over `tibble` for converting objects to tibbles.  Also should probably add a comment here noting that `tibble`/`as_tibble` don't appear to copy the column data (what we want here, though it would lead to violations of `data.table` memory model in general use --- see `data.table:::as.data.frame.data.table` for what it probably wants to be happening in the general case).

R/archive.R
```diff
@@ -439,69 +439,6 @@ epi_archive <-
...
-        )
-        ))))
-        if (methods) {
-          cat("----------\n")
-          writeLines(wrap_varnames(
-            initial = "Public R6 methods: ",
-            names(epi_archive$public_methods)
-          ))
-        }
-      },
```
brookslogan, line 504: It's fine but unnecessary to move this from R6 -> S3 at this point.  (R6 takes care of providing a `print` S3 method forwarding appropriately.  If we were planning to stick with R6, we might want it here so we could potentially change some things into private members & still use them in the print implementation.  Users could also use `x$print()`, thought that's likely not in high demand.)

R/methods-epi_archive.R
brookslogan: Also this is going to invisibly return the tibble-ified DT.  We want to instead `return(invisible(x))`.

R/methods-epi_archive.R
```diff
@@ -1000,3 +1000,79 @@ epix_truncate_versions_after.epi_archive <- function(x, max_version) {
...
+    cat("----------\n")
+    writeLines(wrap_varnames(
+      initial = "Public R6 methods: ",
+      names(epi_archive$public_methods)
+    ))
+  }
+
+  # Error in NextMethod() : generic function not specified
+  # NextMethod()
+
```
brookslogan, line 1076:
```suggestion
```

README.md
```diff
@@ -20,28 +34,26 @@ contains the most up-to-date values of the signals variables, as of a given
...
-  an `epi_df` object over sliding windows in time;
-  
-- `epi_cor()`, for computing lagged correlations between variables in an
-  `epi_df` object, (allowing for grouping by geo value, time value, or any other
-  variables).
+-   `epi_slide()`, for iteratively applying a custom computation to a variable in
+    an `epi_df` object over sliding windows in time;
+-   `epi_cor()`, for computing lagged correlations between variables in an
+    `epi_df` object, (allowing for grouping by geo value, time value, or any other
+    variables).
```
brookslogan, line 44: Q: does additional space between `-` and following non-whitespace character change/improve appearance?

R/methods-epi_archive.R
brookslogan: Requested improvements on tibble print formatting:

1. Can you please replace
```
# A tibble: 129,638 × 5
```
with an explanation of what is being printed?

2. Print the dimensions only once, print the column names only once.
3. Include some of the final rows, like `data.table`'s `print` method does... this will be handy to help point out latency differences between signals.

brookslogan (approved):
This is what we had in mind for adding the DT snippet, but these remain:
> exclude any redundant information

and
> move some other information into a summary method

The first can be resolved pretty easily here (see code convos).  Second involves more opinions & could be spun off to a separate lower-priority Issue.

README.md
dshemetov: It doesn't change appearance, just something my .md formatter prefers.

R/archive.R
dshemetov: Got it, this is something we want to do eventually though right? Does it make sense to do the R6 -> S3 migration incrementally or should it be more discrete?

brookslogan: Eventually, yes.  It'd be nice to distribute the effort incrementally, but (i) it doesn't make sense for users to see a partial migration so it might interfere with releases, and (ii) the complete migration may involve some renaming and access pattern changes that might require scanning ALL of the archive code, so we might have to revisit "already-migrated" stuff.

R/methods-epi_archive.R
brookslogan: issue?
```
* min time value = 2020-01-01
* max time value = 2020-01-01
````
looks fishy

File: R/methods-epi_archive.R (hunk 2 above)
Comment (brookslogan) [reply R/methods-epi_archive.R:1018]: That's a matter of opinion.  Personally I don't think they're that essential.  (Though they'd be nice to see in the type descriptions in the tibble header... but that's a larger restructuring.)

File: R/archive.R (hunk 9 above)
Comment (dshemetov) [reply R/archive.R:504]: Alright, I'll roll it back then

File: R/methods-epi_archive.R (hunk 3 above)
Comment (brookslogan) [reply R/methods-epi_archive.R:1030]: Matter of opinion, this one probably more mixed.  We might try to combine these two lines if we keep them?  Or hold off on excluding them until with have a `summary` method that would give this info?  Or just exclude them.  I don't know.

File: R/methods-epi_archive.R (hunk 4 above)
Comment (brookslogan) [reply R/methods-epi_archive.R:1040]: Min version could be misleading.... archive should represent the FULL version history up to some point.  Just because the first report in that history was the min version doesn't mean that this doesn't/shouldn't represent the full history.

Max version, yes, probably.  Helps users spot when their archive might need a refresh.

File: R/methods-epi_archive.R (hunk 5 above)
Comment (brookslogan) [reply R/methods-epi_archive.R:1052]: You could check if `clobberable_versions_start` is NA; if it is, then you could skip printing this line.  If it's not NA.... not sure.  Might be another thing to keep in `print` only until we have a `summary` method.

Comment (brookslogan) [reply R/methods-epi_archive.R:1052]: Clobberable version tracking might not be necessary for this data structure, but got included from some prototyping with updating archives in mind.  (Updating archives = ones that will update themselves automatically from upstream sources with new issues, patched issues, etc.  "Clobberable" means e.g. an update (e.g. a patch) might re-write version history rather than adding onto it.)

File: R/methods-epi_archive.R (hunk 2 above)
Comment (dshemetov) [reply R/methods-epi_archive.R:1018]: So I'll remove these and we can think about how to bring them back in a cleaner format later.

File: R/methods-epi_archive.R (hunk 6 above)
Comment (brookslogan) [reply R/methods-epi_archive.R:1058]: Number of rows is probably important; number of columns less so if we're also printing column names / a preview.  But no need to print number of rows & columns manually and then print the `as_tibble` form which also prints the same info.

File: R/methods-epi_archive.R (hunk 7 above)
Comment (brookslogan) [reply R/methods-epi_archive.R:1072]: It should go away when we're actually doing the move from R6.  And maybe be replaced by some alternative guidance as to what you can actually do with archives.  Think this is going to be easier to do all at once later on.

File: R/methods-epi_archive.R [hunk 12]
```diff
@@ -1000,3 +1000,79 @@ epix_truncate_versions_after.epi_archive <- function(x, max_version) {
...
+    nrow(x$DT), ncol(x$DT)
+  ))
+  cat(sprintf("Columns in DT: %s\n", paste(ifelse(length(
+    colnames(x$DT)
+  ) <= 4, paste(colnames(x$DT), collapse = ", "),
+  paste(
+    paste(colnames(x$DT)[1:4], collapse = ", "), "and",
+    length(colnames(x$DT)[5:length(colnames(x$DT))]), "more columns"
+  )
+  ))))
```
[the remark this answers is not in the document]
brookslogan: This is redundant with the `as_tibble` sample.  (And just 4 column names was probably too little anyway.)

R/archive.R
brookslogan: issue: `grouped_epi_archive.R` still uses `$print`:

```
        private$ungrouped$print(class = FALSE, methods = FALSE)
```

R/methods-epi_archive.R
brookslogan: issue: `tibble` also hasn't been importFrom'd elsewhere yet so this breaks.

R/methods-epi_archive.R
dshemetov: Alright, I'll keep them and merge into one line.

brookslogan:
question: do you know if `  Warning: <anonymous>: ... may be used in an incorrect context` in the checks was pre-existing?

R/methods-epi_archive.R
dshemetov: Got it thanks.

R/methods-epi_archive.R
dshemetov: Sounds good

R/methods-epi_archive.R
dshemetov: Any pointers for how to replace the message from (1)? That seems to be a default from the tibble print method, is that customizeable?

PR #415, opened 2024-02-07
Ndefries/autocompletion slide

### LINKED ISSUE #256: Provide option to complete/skip incomplete windows, fill gaps in `epi[x]_slide` windows
[Refined scope: see these comments: [A](https://github.com/cmu-delphi/epiprocess/issues/256#issuecomment-1611758042), [B](https://github.com/cmu-delphi/epiprocess/issues/256#issuecomment-1611721332).  We can then split off any other ideas into a separate issue.]

It's very easy to write an erroneous 7-day-average or 7-day-sum slide: e.g., if we `epi_slide` with the natural `mean` or `sum`, we get the sum of less than `n` * {n key values in group} things for
- the first `n - 1L` `time_value`s
- wherever there's a gap in availability (e.g., a skipped `time_value` in a/all `geo_value`s, an unavailable `geo_value` across a wide span of `time_value`s, etc.)

Ideas:
* Have some sort of option to auto-complete the window data, as writing this is a little tedious.  Furthermore, we probably need this to be the default behavior for `epi_slide` or some sort of messaging to go on unless the user explicitly says that they're okay without completion.  When this completion is enabled, we should balk if the key columns don't form a unique key.  Counterpart for `epix_slide` likely needs to complete a lot of leading edges through the `ref_time_value`, inclusive, and balk if there's anything beyond it (e.g., if archive is holding forecasts).
* Have an option to skip over incomplete(-in-another-sense) windows for early `ref_time_value`s by filtering down `ref_time_values` (perhaps as an alternative to the default `ref_time_values` or manual specification, perhaps as a separate option that would allow for this filtering to be applied even to manually-supplied `ref_time_values`).  (An `epi_archive` equivalent might not be so simple due to the possibility of later versions extending the time series into the _past_ or removing `time_value`s from the past, so `min(time_value)` is not a fixed property of an archive group.)  --- This seems closest to `.complete=TRUE` for `slider::slide_index*`, but I'm not sure we would want to ignore gaps like these functions, and there might be details to check regarding when the `time_value` set available varies across groups and/or "epikey" (non-time, non-version key col) values.  
* Have option to skip over or return some missing result indicator when there are incomplete windows in any sense (either because we are at the very beginning/end of the time series and the window extends past the edge, or because there are gaps). 
* Reject `epi_df`s/`epi_archive`s with gaps / gaps in slide computation inputs unless the user explicitly allows them.  (For the user to easily address these gaps, we want #250 + something for `epi_archive`s.)

We might not really need all of these; some of them make the others unnecessary or not as useful; some brainstorming still is needed here.

nmdefries:
### Checklist

Please:

-   [ ] Make sure this PR is against "dev", not "main".
        brookslogan, nmdefries.
-   [ ] Makes sure to bump the version number in `DESCRIPTION` and `NEWS.md`.
-   [ ] Describe changes made in NEWS.md, making sure breaking changes

By default, complete input data so that computations are always handed `n obs * m groups`. This currently only adds this feature for `epi_slide`; we want to also add it for `epi_slide_mean` and `epix_slide`.

Closes #256.


R/slide.R
```diff
@@ -573,81 +663,11 @@ epi_slide_mean = function(x, col_name, ..., before, after, ref_time_values,
     after <- 0L
   }
 
-  pad_early_dates <- c()
```
nmdefries, line 576: This chunk is turned into a fn below so that `epi_slide` can also use it to get a complete date sequence using the complicated `time_step` handling.

R/slide.R
```diff
@@ -225,6 +248,67 @@ epi_slide <- function(x, f, ..., before, after, ref_time_values,
...
 
+  if (autocomplete_windows) {
+    assert_list(fill, names = "unique")
+    fill_vars_not_in_x <- !(names(fill) %in% colnames(x))
+    if (any(fill_vars_not_in_x)) {
+      Warn(
+        c(
+          "Some names provided in `fill` do not correspond to column names in
+           the input data. Any columns not specified in `fill` will be
+           completed with missing values (`NA`) by default."
```
brookslogan, line 259: Second sentence seems a little tangential.  Might want to focus on what they need to do to fix it.  (Does tidyr not abort when there's this sort of typo?)

R/slide.R
```diff
@@ -225,6 +248,67 @@ epi_slide <- function(x, f, ..., before, after, ref_time_values,
...
+
+    # Make a complete date sequence between min(x$time_value) and max
+    # (x$time_value), plus pad values. We need to include pad dates in
+    # the complete date sequence so that the first and last n - 1
+    # ref_time_values have a complete window to use.
+    date_seq_list <- full_date_seq(x, before, after, time_step)
+    all_dates <- c(
+      date_seq_list$pad_early_dates,
+      date_seq_list$all_dates,
+      date_seq_list$pad_late_dates
```
nmdefries, line 284: `full_date_seq` returns `all_dates` and pad dates separately to support current `epi_slide_mean` implementation, but that can be changed to simplify this.

R/slide.R
```diff
@@ -225,6 +248,67 @@ epi_slide <- function(x, f, ..., before, after, ref_time_values,
...
+    #    doesn't change the impact of completion because we apply the
+    #    computation by group as well.
+    #  - If a geo first appears halfway through the dataset, it will be
+    #    completed all the way back to the beginning of the data.
+    x <- tidyr::complete(x,
+        expand(x, nesting(!!key_cols_no_time), data.frame(time_value = all_dates)),
+        # `complete` checks that fill types match existing column types.
+        fill = fill,
+        # Existing missings will be replaced by `fill`, too.
+        explicit = TRUE
```
brookslogan, line 305: `explicit = TRUE` doesn't seem very natural to me.  Could we also make this a parameter?  (And maybe use a more sane name&default?  Tradeoff: I guess that could also surprise people if `fill` makes them think we're mirroring tidyr defaults, or make them read more docs.  Not sure what's preferable.)

R/slide.R
```diff
@@ -225,6 +248,67 @@ epi_slide <- function(x, f, ..., before, after, ref_time_values,
...
+      )
+
+    # `complete` strips epi_df format and metadata. Restore them.
+    x <- reclass(x, attributes(x)$metadata)
```
brookslogan, line 309: do we need to use pre-recorded metadata beforehand?

R/slide.R
```diff
@@ -225,6 +248,67 @@ epi_slide <- function(x, f, ..., before, after, ref_time_values,
...
+    key_cols_no_time <- kill_time_value(key_cols)
+
+    # Add in rows for each present key column combination for all dates.
+    #  - This happens within each group if the data is grouped. This
+    #    doesn't change the impact of completion because we apply the
+    #    computation by group as well.
+    #  - If a geo first appears halfway through the dataset, it will be
+    #    completed all the way back to the beginning of the data.
+    x <- tidyr::complete(x,
+        expand(x, nesting(!!key_cols_no_time), data.frame(time_value = all_dates)),
```
brookslogan, line 301: Ouch, this is complicated.  Two issues.  They feel a bit edge-case-y, but matter more if we're trying to have some consistency with `epix_slide`, or maybe a future `epi_slide_reframe` that doesn't try to match the existing keys in the output.
- If reporting starts for one epikey later on, or if there are large gaps in its reporting, then even without data revisions, you could have later epi_slides on expanded epi_dfs outputting different results for earlier ref_time_values, as they will be considering additional epikeys.  If grouping by all epikeys, then maybe this can't happen for `epi_slide` (but could for an analogous `epix_slide`).  But if you don't group by anything, you might turn earlier national sums into a bunch of NAs.
- Should we be using only the epikeys that have appeared, or the cross product of all seen geos x all seen age groups x etc.?  If we have to pick one, I think the current approach of using only epikeys that have appeared is right; user can `complete` or join or whatever if they don't like that.  (Again, for `epi_slide`, this probably only matters about whether we make not-grouped-by-all-epikeys epi_slides output NAs...)

R/slide.R
```diff
@@ -756,3 +776,93 @@ epi_slide_mean = function(x, col_name, ..., before, after, ref_time_values,
...
+        class = "epiprocess__epi_slide_mean__unmappable_time_type",
+        epiprocess__time_type = ttype
+      )
+    }
+
+    # `seq` `by` arg can be any of `c("days", "weeks", "months", "quarters", "years")`.
+    all_dates <- seq(min(x$time_value), max(x$time_value), by = by)
+
+    if (before != 0) {
+      pad_early_dates <- Start(all_dates) - before:1
```
brookslogan, line 836: This should be something like `time_step(before:1)` or maybe something more complicated factoring in time_value class?  So core slide impl actually grabs the expected pad.  (same for late padding)

R/slide.R
brookslogan: One more thing: I couldn't find where the matching-old-epikey-time_value/broadcasting part is.  I wonder if it's using this completed `x`.  If it is, then the two questions above actually are more relevant for `epi_slide` since it involves also changing the epikeysets output for earlier time values simply by adding an epikey in rows for later time_values, and would also mean outputting values for gaps, and maybe also NA's at the leading and trailing edges.  Ideally we should consciously make decisions here, or make it configurable.

brookslogan (commented):
Some notes as I ran through.  Unfortunately, there are more design questions I forgot about than I expected; maybe we can brainstorm on Slack/call.

R/slide.R
brookslogan: (you may be able to get rid of the `expand` call)

R/slide.R
nmdefries: Good catch. I'll add this on https://github.com/cmu-delphi/epiprocess/pull/400 + a test for this problem. Should be able to use `seq(from, by, length.out)` here to avoid having to figure out the `time_step` to accompany the `by`/`time_type`.

R/slide.R
nmdefries: Does/could metadata get modified by prior processing?

R/slide.R
brookslogan: A full_seq may be missing in definition of all_dates.  If there is a gap for all epikeys then completion may not fill it in.

nmdefries: `full_date_seq` fn below should take care of that. The three cases ([tsibble class or numeric](https://github.com/cmu-delphi/epiprocess/blob/adbdcd414cf8d1366af2bb08eded088dc6eb51b8/R/slide.R#L797), [else `time_step` not defined](https://github.com/cmu-delphi/epiprocess/blob/adbdcd414cf8d1366af2bb08eded088dc6eb51b8/R/slide.R#L833), [else `time_step` defined](https://github.com/cmu-delphi/epiprocess/blob/adbdcd414cf8d1366af2bb08eded088dc6eb51b8/R/slide.R#L852)) all generate date sequences by stepping by `time_step` (or similar) between the input data's min and max time_values.

R/slide.R
```diff
@@ -225,6 +248,67 @@ epi_slide <- function(x, f, ..., before, after, ref_time_values,
...
+    all_dates <- c(
+      date_seq_list$pad_early_dates,
+      date_seq_list$all_dates,
+      date_seq_list$pad_late_dates
+    )
+
+    # A helper column marking real observations.
+    x$.real <- TRUE
+    # Fill the `.real` column with `FALSE` for rows added during completion.
+    fill$.real <- FALSE
```
brookslogan, line 290: is `.real` handling still computationally significant?  Wonder if in autocomplete case it could just go away.

nmdefries: The prior slow `.real` handling was in `f_wrapper` around the user-passed computation, so it got called a ton. This should be pretty fast since it only happens once. I'm curious what the alternative to the `.real` col would be, if we still plan on dropping completed rows.

R/slide.R
nmdefries: [Test](https://github.com/cmu-delphi/epiprocess/pull/400/commits/a429145c96c4b692ec442758630a3947ae5e9db3) and [change](https://github.com/cmu-delphi/epiprocess/pull/400/commits/4bf2a29e2583e7cafc4d6c86cca285c87c3a9f09).

R/slide.R
nmdefries: `tidyr::complete` silently ignores list items that don't correspond to observed column names.

R/slide.R
nmdefries: Key cols are broadcast backwards here:
```
expand(x, nesting(!!key_cols_no_time), data.frame(time_value = all_dates))
```
This wouldn't change output epikeysets for earlier time values. In the non-aggregating case, computations are both completed across and computed across all keycols, and we later (at the end of this function) filter out not-`.real` obs. Epikeysets that were broadcast back in time will be removed by the filter.

In the aggregating case (e.g. states -> national sum), we're outputting all-new epikeysets. We should end up with one row per group. As long as the input group data contains one `.real = TRUE` obs, the output should be kept (?)(I'm not sure how `.real` is handled right now -- need to check/implement that logic).

dshemetov:
Many of the ideas here were re-implemented in #519. Move to close?

PR #425, opened 2024-03-05
ci: add document, lint, style like epidatr

dshemetov:
### Checklist

Please:

        brookslogan, nmdefries.

Adds the same CI actions we have in epidatr: style, lint, document.


.github/workflows/style.yml
```diff
@@ -0,0 +1,82 @@
...
+  push:
+    paths:
+      [
+        "**.[rR]",
+        "**.[qrR]md",
+        "**.[rR]markdown",
+        "**.[rR]nw",
+        "**.[rR]profile",
+      ]
+    workflow_dispatch:
```
brookslogan, line 13: question: what does this do at this indentation level?  should it be on level to the left?

question: I thought you found styling & documenting interfered with each other sometimes, and chose to make like... documentation pushes automatic, and styling into just a check + maybe a workflow_dispatch command to actually do the styling & push?

.github/workflows/document.yaml
```diff
@@ -0,0 +1,52 @@
...
+          needs: |
+            devtools
+            roxygen2
+
+      - name: Document
+        run: roxygen2::roxygenise()
+        shell: Rscript {0}
+
+      - name: Build README.md from README.Rmd
+        run: Rscript -e 'devtools::build_readme()'
```
brookslogan, line 42: suggest: make some notes at top of file of things like this & the workflow dispatch stuff below that have been changed from the original files, so if we want to use `usethis` or that repo to update things, we'll know we need to rebase our own changes on top.

.github/workflows/style.yml
brookslogan: suggest: if you've set up workflow_dispatch commands for this, then please document them somewhere... maybe the PR template?

dshemetov: - Ah yes, good catch, the indentation is off.
- It's less that styling and documentation interfere with each other and more that once the document action pushes, the pushed commit doesn't rerun CI. So workflow_dispatch allows us to retrigger without another push.
- Can do that.

.github/workflows/document.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Commit and push changes
+        run: |
+          git config --local user.name "$GITHUB_ACTOR"
+          git config --local user.email "$GITHUB_ACTOR@users.noreply.github.com"
+          git add README.md
+          git add man/\* NAMESPACE DESCRIPTION
+          git commit -m "docs: document (GHA)" || echo "No changes to commit"
+          git pull --rebase
```
brookslogan, line 51: issue: it might be possible to get into bad race conditions here with document + style on examples.  If you think Git will catch this and balk, then maybe it's okay this way.  But this ff-only + [this concurrency group thing](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-concurrency-groups) might work better if it does what I think.

brookslogan (approved):
These sound nice.  Some minor requests & questions, some because I'm probably a bit out of date on the intended workflow for using these actions.

.github/workflows/document.yaml
dshemetov: Hm, given that we've been exposed to this potential race condition in epidatr for a while now and haven't had an issue, I figure Git is taking care of it. Can find a fix if we get a mysterious error here.

.github/workflows/style.yml
brookslogan: Oh, that's what workflow_dispatch means?  I mean, that's what is sounds like, but not what skimming a few lines of docs made me think / what I thought earlier plan was; I thought these were special slash commands for comments.

dshemetov: Ah yes, it lets you trigger the workflow from the Actions tab at the top, like [here](https://github.com/cmu-delphi/epiprocess/actions/workflows/pkgdown.yaml).

.github/workflows/document.yaml
brookslogan: So there may be some more low-probability races between the final push&pulls; I think Git can detect those.  But more commonly, I imagine you could have:
- doc & style start simultaneously
- styler finishes, pulls, pushes
- document finishes, pulls & rebases, pushes
- document still needs to run another time because styler updated example source.  Git does not detect conflict unless styler also re-documents after touching examples.  What does workflow_dispatch: do?  (And if it saves us here, how does it not get into infinite loops?)

brookslogan: I guess concurrency groups probably aren't the right thing either.  Just need an order: style first, then document.  Not sure about lintr thing, maybe that's just a check that could be in parallel.

dshemetov: workflow_dispatch lets you queue another document command that this time contains the styling changes, if you wish. In practice, the next commit you push will run one or the other again and that will get you back to a stable state, so it hasn't been an issue.

brookslogan: Oh, I guess maybe document will be triggered to re-run if styler does update examples.  Okay, maybe there's no uncaught bad stuff that I can think of happening.

brookslogan: Yeah, if there is that super rare chance of the final pull-pushes causing Git to balk and error out one of the actions, we can just dispatch it.

PR #431, opened 2024-03-20
wip feat: replace R6 `epi_archive` with S3 implementation

### LINKED ISSUE #340: Remove R6 interface for `epi_archive`s
From discussion in #181 and off GitHub: in `epi_archive` we are currently paying the costs of maintaining a separate R6 interface and considering details of aliasing and mutation, but none of our operations actually benefit from the ability to use mutation, and we've eliminated or don't use some of the other benefits of R6 (keeping package namespace clean, `print`, cross-package inheritance details).

1. We should eliminate the R6 user interface.  [All `archive$<fn>()` parts of the interface.]
2. We should refactor away all internals that rely on mutation to not do so.  But unless we also move away from `data.table` (see below), we still need to be careful about documenting aliasing of DTs.
3. We should consider changing `epi_archive` and `grouped_epi_archive` to S3 classes, not R6.
  - Should grouped epi archives be of class `c("grouped_epi_archive", "epi_archive")` or just `"grouped_epi_archive"`?  The former is more consistent with dplyr, but the latter is [less error-prone](https://github.com/cmu-delphi/epiprocess/issues/64#issuecomment-1163634934).
    - [x] Let's go with just `"grouped_epi_archive"`.
  - Is R6 providing better error signaling when users try to use functions we haven't thought of on `epi_archive`s that might start "working" and doing wrong things as lists?  (E.g., `c`? `rbind`?)
    - [ ] This needs testing.  If sensible-sounding S3 generics "work" (incorrectly) on the new S3-class list format, we should probably provide methods for them to throw appropriate errors until we actually provide real implementations.  This also would mean there is still usefulness for listing actually-applicable methods like we do currently.
  - Consider technicalities with `is_epi_archive` etc. on saved archives from older versions.  We don't have a format version attribute now, so we'd need to check if it's an R6 object for example.  Maybe we should add an attr now indicating the "archive format version"?
4. [x] Consider whether we should simultaneously try to move to tibbles + dplyr 1.1 joins, polars, or `lazy_dt`, so we don't have to mention aliasing at all, and to try to improve performance.
  - [Initial testing](https://github.com/cmu-delphi/epiprocess/issues/265) indicates that dplyr 1.1 joins are significantly slower, and other alternatives may have comparable or greater baggage than `data.table`.
  - `lazy_dt` might be nice except it would mean simply loading epiprocess can break unrelated user code using dplyr + non-dtplyr data.tables.
  - Seems good to stay with `data.table` for now, but we will need to check that (i) none of our `epix_` methods mutate DT without copying it first (or doing sneaky things aliasing columns), and (ii) that we warn the user that if they should do the same (minus the sneaky things part at least).

dshemetov:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

- ~~**Attention conservation notice:** probably don't review this yet, still waiting downstream A/B tests~~
- Prior comparison work in #430
- The diffs should be easier to read here

A/B tests TODO (A = epiprocess dev, B = this branch):
- [x] make sure vignettes are identical
- [x] make sure the output from a subset of the exploration-tooling forecasters is identical 

- Resolves #340


R/archive.R
```diff
@@ -196,572 +187,755 @@ next_after.Date <- function(x) x + 1L
...
+      fromLast = TRUE
+    ) %>%
+    tibble::as_tibble() %>%
+    dplyr::select(-"version") %>%
+    as_epi_df(
+      geo_type = epi_archive$geo_type,
+      time_type = epi_archive$time_type,
+      as_of = max_version,
+      additional_metadata = c(epi_archive$additional_metadata,
+        other_keys = other_keys
```
brookslogan, line 517:
```suggestion
        list(other_keys = other_keys)
```
This is hopefully a simple fix for a pre-existing thing that stood out here.  Filed off another one [here](https://github.com/cmu-delphi/epiprocess/issues/437).

`c` should basically never be used on a combination of list/NULL and non-list objects.  Example error:
```r
c(NULL, other_keys = c("sex", "age_group"))
#> other_keys1 other_keys2 
#>       "sex" "age_group" 
```

R/archive.R
```diff
@@ -196,572 +187,755 @@ next_after.Date <- function(x) x + 1L
...
+    )
+    new_versions_end <- fill_versions_end
+    # Update `epi_archive` all at once with simple, error-free operations +
+    # return below:
+    epi_archive$DT <- new_dt
+    epi_archive$versions_end <- new_versions_end
+  } else {
+    # Already sufficiently up to date; nothing to do.
+  }
+  return(invisible(epi_archive))
```
brookslogan, line 598: todo: return visibly, not invisibly.  Same deal with the rest of the S3 methods besides `print`.  These other methods are no longer used primarily for mutation or another side effect, but rather are primarily for a return value, so return value should be visible.

R/archive.R
```diff
@@ -196,572 +187,755 @@ next_after.Date <- function(x) x + 1L
...
-            "Columns `geo_value`, `time_value`, and `version` must be present in `x`."
-          )
-        }
-        if (anyMissing(x$version)) {
-          cli_abort("Column `version` must not contain missing values.")
-        }
+#' @name epi_archive
+epi_archive <- function(...) {
+  new_epi_archive(...)
+}
```
brookslogan, line 198: suggest: delete this, but make sure `?epi_archive` still works.  People may expect that you can feed in column data in a manner similar to `data.frame`, `data.table`, `tibble`, etc., especially with the dots delegation being used.

R/archive.R
```diff
@@ -196,572 +187,755 @@ next_after.Date <- function(x) x + 1L
...
+
+  structure(
+    list(
+      DT = DT,
+      geo_type = geo_type,
+      time_type = time_type,
+      additional_metadata = additional_metadata,
+      clobberable_versions_start = clobberable_versions_start,
+      versions_end = versions_end,
+      private = list()
```
brookslogan, line 402: thought: not sure we need `private` right now since it's empty, unless you're sure this is the way to go & want to set up `print` for it ahead of time.

brookslogan (changes requested):
Some minor todos & larger suggestions from a partial first pass; thought I'd just post them now since some probably will be slightly annoying to implement & finishing reviewing's going to take a while.

For future review passes:
- I need to learn the `new_` etc. pattern --- I think Daniel opened an issue about us using this for `epi_df`s & Hadley has a chapter about it & related functions.
- Check that we're achieving the intended non-mutation interface.
- Finish looking over files; rest of `archive.R`, all of `grouped_epi_archive.R`, ...
- Check `archive_cases_dv_subset.R`.

tests/testthat/test-epix_fill_through_version.R
```diff
@@ -70,7 +70,7 @@ test_that("epix_fill_through_version does not mutate x", {
     as_epi_archive(tibble::tibble(geo_value = 1L, time_value = 1L, version = 1L, value = 10L))
   )) {
     # We want to perform a strict comparison of the contents of `ea_orig` before
-    # and `ea_orig` after. `clone` + `expect_identical` based on waldo would
+    # and `ea_orig` after. `copy` + `expect_identical` based on waldo would
```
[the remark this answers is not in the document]
dshemetov: Thanks, I didn't look into the comments in this file. I'm noticing a lot of stuff around using `testthat::local_edition(3)`, which wanted to enforce content comparison instead of memory address. I'm thinking we could try removing that too, what do you think?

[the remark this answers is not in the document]
dshemetov: Just tried cleaning a bit of this up and it seems to work (though I'm not fully following where the copies are happening here). Gonna push the commit, but I'll need to remember to come back here and look a little deeper.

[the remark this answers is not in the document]
brookslogan: There's a lot of `testthat::local_edition(3)` to enforce consistency, so if there's another edition-like update, we don't suddenly get different behavior of these tests, and so that theoretically we could piecewise update the tests to the new edition in that case.  We could maybe do that with testthat version pinning or some other mechanism, though AFAIK the DESCRIPTION file doesn't allow for comments that make clear how big a breaking change a testthat version defaulting to a new edition could be.

R/archive.R
```diff
@@ -196,572 +187,755 @@ next_after.Date <- function(x) x + 1L
...
+
+#' @export
+fill_through_version <- function(x, ...) {
+  UseMethod("fill_through_version")
+}
+
+
+#' Fill through version
+#' @description Fill in unobserved history using requested scheme by mutating
+#'   the given object and potentially reseating its fields. See
```
brookslogan, line 533: issue: notes here & probably elsewhere about mutating, mutation, and reseating need to be removed or updated.  We may still want some notes about potentially aliasing the old DT, or maybe we could just make a global warning about that in the "caution" section.

tests/testthat/test-epix_fill_through_version.R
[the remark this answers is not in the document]
dshemetov: Isn't testthat too embedded in the R ecosystem to make breaking changes? To me, it seems unlikely they'll move away from the behaviors of the current edition (3).

R/archive.R
```diff
@@ -196,572 +187,755 @@ next_after.Date <- function(x) x + 1L
...
+  if (anyMissing(x$version)) {
+    cli_abort("Column `version` must not contain missing values.")
+  }
 
-        # If time type is missing, then try to guess it
-        if (missing(time_type)) {
-          time_type <- guess_time_type(x$time_value)
-        }
+  # If geo type is missing, then try to guess it
+  if (missing(geo_type) || is.null(geo_type)) {
```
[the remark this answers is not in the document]
dshemetov: I'm thinking get rid of the `missing(geo_type)`? I think it's good enough to have NULL (and have NULL be default) trigger the guess.

R/archive.R
```diff
@@ -196,572 +187,755 @@ next_after.Date <- function(x) x + 1L
...
 
-        # Filter by version and return
-        if (all_versions) {
-          result <- epix_truncate_versions_after(self, max_version)
-          # `self` has already been `clone`d in `epix_truncate_versions_after`
-          # so we can modify the new archive's DT directly.
-          result$DT <- result$DT[time_value >= min_time_value, ]
-          return(result)
-        }
+  return(invisible(epi_archive$DT %>% print()))
```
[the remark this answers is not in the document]
dshemetov: Made these fixes along with others to `print.epi_archive`. 

Currently the mixing works fine. Let's wait until we have a concrete problem to fix before filing an issue.

tests/testthat/test-epix_fill_through_version.R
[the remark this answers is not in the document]
brookslogan: It already was with 2 I felt, but they did do 3.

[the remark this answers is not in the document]
brookslogan: "(We don’t anticipate creating new editions very often, and they’ll always be matched with major version, i.e. if there’s another edition, it’ll be the fourth edition and will come with testthat 4.0.0.)" from [here](https://testthat.r-lib.org/articles/third-edition.html) --- they at least keep the possibility open.  And Hadley/RStudio philosophy is different from R Core philosophy, more willing to make breaking changes.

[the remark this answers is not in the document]
brookslogan: Makes sense.  I think having them scattered made more sense for 2 -> 3 so it showed where we actually relied on that behavior, but who knows what would change in 3 -> 4.

dshemetov:
Force pushed some changes, just a heads up @brookslogan

R/archive.R
```diff
@@ -196,572 +187,755 @@ next_after.Date <- function(x) x + 1L
...
+  )
+
+  if (length(x$private_fields) != 0L) {
+    cli_abort("expected no private fields in x",
+      internal = TRUE
     )
+  }
+
+  # Mutate fields all at once, trying to avoid any potential errors:
+  for (field_name in names(x$public_fields)) {
```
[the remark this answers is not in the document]
dshemetov: removed this function, no more references to these anymore

R/archive.R
dshemetov: Reading through the code now, I see this isn't used. We do have a `private` field in grouped_epi_archive and that _is_ used, so I'll keep that one, but remove this one.

R/archive.R
dshemetov: So I've removed most of the warnings about mutation and aliasing. Two questions:
- do we actually need to warn the user about aliasing if there's no risk of mutation? seems like information we could omit and reduce mental load on the reader
- if we do decide that we want to store this somewhere, where should it go? maybe a developer doc? i added a separate note on "compactification" for archives, so we can keep the compactification arg description short and link to it from other places, and it's under a "Advanced" topic section. maybe a note on aliasing there?

brookslogan: - There's a minor risk of mutation if (i) we don't encapsulate, and (ii) the user mutates.
- Advanced section sounds good.
- Eventually, post-encapsulation, and maybe if we have better interface/naming for a few things (construction from snapshots, construction from update data, extracting update data), we won't even need to mention compactification and DT and aliasing in user docs at all.  But this is like two steps down the line from this PR.

brookslogan: It might also help to reduce mental load to prefix it by "For data.table users:" or something like that.

R/archive.R
```diff
@@ -878,17 +296,356 @@ epi_archive <-
...
-as_epi_archive <- function(x, geo_type, time_type, other_keys,
+#'
+new_epi_archive <- function(
+    x,
+    geo_type = NULL,
+    time_type = NULL,
+    other_keys = NULL,
+    additional_metadata = NULL,
+    compactify = NULL,
+    clobberable_versions_start = NA,
```
brookslogan, line 307: suggest: eliminate `new_epi_archive`, keep it as `as_epi_archive`, or introduce a `new_` and `validate_` that are more minimal, and have `as_epi_archive` be the "helper" as described [here](https://adv-r.hadley.nz/s3.html).

suggest: either (a) make more defaults non-NULL, or (b) make `clobberable_versions_start` default also NULL (& replace with NA) if it's possible.  I'm not sure about (a) vs. (b).  (a) gives more type info, but also could hide other arg names in limited-width autocomplete windows.  With (a) approach, there can sometimes also be some issues with confusing defaults --- imagine we had `geo_type = guess_geo_type(x)` default and rebound `x` before `geo_type` was evaluated --- but it doesn't look like we rebind before assigning any of the defaults, so I don't _think_ we're in such a situation             .

brookslogan:
Core refactor changes look pretty good!  Please:
- [ ] todo: Review & if good, merge changes from #443.  Some things of note:
  - `grouped_epi_archive`s aren't `epi_archive`s (e.g., we don't want to try to `x$DT` on a `grouped_epi_archive` since that will be `NULL`.  And we don't want to try to mirror the `epi_archive` structure with `DT` at the top level because that's asking for bugs --- (which I have encountered/made due to `grouped_df`s also being `tibble`s; I don't think we want to mirror this).)
  - `clone()` wasn't needed in a lot of places, and we want to avoid deep copies when we can
- [ ] Look at above/below github code comments.
- Perform vignette snapshot tests, then purge the vignette snapshots from history?
  - [ ] purged
  - [ ] unnecessary

brookslogan (approved):
Please see separate checklist.

R/methods-epi_archive.R
```diff
@@ -941,26 +785,52 @@ group_by.epi_archive <- function(.data, ..., .add = FALSE, .drop = dplyr::group_
...
+    before,
+    ref_time_values,
+    time_step,
+    new_col_name = "slide_value",
+    as_list_col = FALSE,
+    names_sep = "_",
+    all_versions = FALSE) {
   if (!is_epi_archive(x, grouped_okay = TRUE)) {
     cli_abort("`x` must be of class `epi_archive` or `grouped_epi_archive`.")
   }
```
brookslogan, line 802: note: this type check is unnecessary as S3 dispatch will take care of it for us.  Is this for a clearer error message than S3 failed dispatch?

R/methods-epi_archive.R
```diff
@@ -171,8 +256,6 @@ epix_fill_through_version <- function(x, fill_versions_end,
 #'   as_epi_archive(compactify = TRUE)
 #' # merge results stored in a third object:
 #' xy <- epix_merge(x, y)
-#' # vs. mutating x to hold the merge result:
-#' x$merge(y)
 #'
 #' @importFrom data.table key set setkeyv
 #' @export
```
brookslogan, line 261: question: would we want this as an S3 generic as well, or as an implementation of the `merge` generic?  Then people using `methods(, "epi_archive")` might get better results.

R/growth_rate.R
```diff
@@ -226,11 +226,11 @@ growth_rate <- function(x = seq_along(y), y, x0 = x,
...
-      if (is.null(ord)) ord <- 3
-      if (is.null(maxsteps)) maxsteps <- 1000
-      if (is.null(cv)) cv <- TRUE
-      if (is.null(df)) df <- "min"
-      if (is.null(k)) k <- 3
+      ord <- ord %||% 3
+      maxsteps <- maxsteps %||% 1000
+      cv <- cv %||% TRUE
+      df <- df %||% "min"
+      k <- df %||% 3
```
brookslogan, line 233:
```suggestion
      k <- k %||% 3
```

R/methods-epi_archive.R
dshemetov: Nah, just left-over from when epix_slide did not S3 dispatch https://github.com/cmu-delphi/epiprocess/blob/4e65e51bb56ab70cc98fa2d37dd35a4ab2336620/R/methods-epi_archive.R#L946-L962. Removing this.

R/methods-epi_archive.R
dshemetov: Not sure. I've never used `methods(, "<class>")` to explore documentation. Let's just see how `epix_*` works out for now?

dshemetov:
I've purged the vignette snapshots from this branch by rebasing and force pushing, FYI @brookslogan

R/methods-epi_archive.R
```diff
@@ -22,36 +22,17 @@
...
-#'
-#' Mutation and aliasing: `epix_as_of` and `$as_of` will not mutate the input
-#' archives, but may in some edge cases alias parts of the inputs, so copy the
-#' outputs if needed before using mutating operations like `data.table`'s `:=`
-#' operator. Currently, the only situation where there is potentially aliasing
-#' is of the `DT` in edge cases with `all_versions = TRUE`, but this may change
-#' in the future.
-#'
 #' @examples
 #' # warning message of data latency shown
```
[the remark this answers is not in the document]
brookslogan: That makes sense.  Though I might sort of soft-request the partial matching thing in `$` and at least name validation in `$<-` and `[[<-` also to catch errors we make in development.  Not part of this PR though.

R/methods-epi_archive.R
brookslogan: Sure.  Side note: `methods(, "<class>")` or `sloop::{s3,s4}_methods_class()` (better than the former when s4 is involved in ways that confuse `methods()` or its users) is useful for exploring undocumented stuff --- often S3 implementations won't actually be exported, documented, etc.  Of course it's not comprehensive if there are regular functions that can also operate on the class.  I guess that was one benefit of R6, potentially having a sort of full list of functionality at your fingertips.

R/archive.R
dshemetov: Alright, I added a validate_epi_archive and moved a bulk of the validation from new_epi_archive in there. Now as_epi_archive calls validate and then new. new_epi_archive is only used there, so this change should be fine for now, but in the future we can refactor some safe internal calls to as_epi_archive with new_epi_archive. There is still a bunch of validation and construction of the data.table object in new_epi_archive, so that's another thing that could be fixed.

I also made `clobberable_versions_start` default to NULL and then I set it to NA in that case. Just went with the consistent option for now, though we can easily switch it later.

dshemetov: See #445

dshemetov:
I think this PR is ready. @nmdefries (thank you!) reproduced my run of exploration-tooling with this branch and found that the forecaster outputs did not change.

I found a few things that need to change in https://github.com/cmu-delphi/delphi-tooling-book, so I'm working on a PR there. We can probably merge this and I can have that one ready tomorrow.

R/methods-epi_archive.R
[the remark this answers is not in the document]
brookslogan: -> #449

PR #433, opened 2024-03-25
Generalize optimized `epi_slide` fn 

nmdefries:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

Make `epi_slide_mean` more general. Let the user pass a `data.table` rolling fn (`frollmean`, `frollsum`, `frollapply`) or `slider` rolling function (`slide_sum`, etc) to expand optimized slide functionality.


tests/testthat/test-epi_slide.R
```diff
@@ -1386,3 +1386,33 @@ test_that("`epi_slide_mean` errors when passed `col_names` as list", {
...
+  result2 <- epi_slide_opt(small_x, value,
+    f = data.table::frollmean,
+    before = 50, names_sep = NULL, na.rm = TRUE
+  )
+  expect_identical(result1, result2)
+
+  # # Theoretically should work, but getting "Error: C stack usage  7973092 is
+  # too close to the limit" (recursion is too deep)
+  # result3 <- epi_slide_opt(small_x, value, f = slider::slide_mean,
+  #   before = 50, names_sep = NULL, na_rm = TRUE)
```
nmdefries, line 1401: question: Not sure what to do about this. I do think the `slider` functions are useful to support, but this seems like an issue with their code.

`data.table` functions work fine, and I believe both it and `slider` are implemented in C++. [Other tests I've done](https://delphi-org.slack.com/archives/C028AUR6MV1/p1707760507375069) didn't have the recursion issue, so I suppose it is the result of `slider` inside of a `dplyr` call.

R/slide.R
```diff
@@ -731,6 +788,134 @@ epi_slide_mean <- function(x, col_names, ..., before, after, ref_time_values,
   return(result)
 }
 
+#' Optimized slide function for performing rolling averages on an `epi_df` object
```
dsweber2, line 791: The diff is grabbing things in a weird order; I'm guessing this is just the previous docs for `epi_slide_mean`?

dsweber2: actually, the man page suggests something different. I'm a bit confused

tests/testthat/test-epi_slide.R
dsweber2: I can reproduce this problem on my machine, though the actual stack size is different. Doesn't look like they have a corresponding error [here](https://github.com/r-lib/slider/issues)?

dsweber2: actually, throwing a browser on the pile, looks like the problem is happening when trying to identify which function this is?

dsweber2: [just sort of blindly following this advice](https://stackoverflow.com/questions/72219118/how-can-i-check-in-r-if-a-function-is-in-a-list-of-functions) seems to have worked. I'll add the suggestion above

R/slide.R
```diff
@@ -550,6 +566,39 @@ epi_slide_mean <- function(x, col_names, ..., before, after, ref_time_values,
...
   }
 
+  # Check that slide function `f` is one of those short-listed from
+  # `data.table` and `slider` (or a function that has the exact same
+  # definition, e.g. if the function has been reexported or defined
+  # locally).
+  if (any(sapply(
+    c(frollmean, frollsum, frollapply),
+    function(roll_fn) {
+      isTRUE(all.equal(f, roll_fn))
```
dsweber2, line 576:
```suggestion
      isTRUE(identical(f, roll_fn))
```

R/slide.R
```diff
@@ -550,6 +566,39 @@ epi_slide_mean <- function(x, col_names, ..., before, after, ref_time_values,
...
+    }
+  ))) {
+    f_from_package <- "data.table"
+  } else if (any(sapply(
+    c(slide_sum, slide_prod, slide_mean, slide_min, slide_max, slide_all, slide_any),
+    function(roll_fn) {
+      isTRUE(all.equal(f, roll_fn))
```
dsweber2, line 583:
```suggestion
      isTRUE(identical(f, roll_fn))
```

tests/testthat/test-epi_slide.R
dsweber2, line 1: Might want tests for anonymous functions that are identical. otoh, maybe not, since you do use `...` to pass on arguments

dsweber2 (commented):
Other than this bit with slider seems good

tests/testthat/test-epi_slide.R
nmdefries: Oh, interesting. I was trying to use `all.equal` in case the user-passed function had been reexported (namespace wasn't `slider` or `data.table` even though the actual function behavior is the same). Odd that that causes problems. Thanks for finding a fix.

R/slide.R
nmdefries: Diff and function documentation look fine to me. I put most of the args documentation in `epi_slide_opt` and inherited those fields here and for `epi_slide_sum` using the `inherit` and `inheritParams` roxygen tags. So the man pages look quite similar between `epi_slide_opt`, `epi_slide_mean`, and `epi_slide_sum`.

tests/testthat/test-epi_slide.R
nmdefries: `identical` won't recognize an anonymous function (or even a reexported function) as being the same. `all.equal` is a [bit more flexible](https://stackoverflow.com/questions/32266433/in-r-how-do-i-test-that-two-functions-have-the-same-definition), but (apparently) causes the recursion-too-deep error.

dsweber2:
Looks like the linter and styler disagree about the spacing for this:
```
  if (inherits(x$time_value, c("yearquarter", "yearweek", "yearmonth")) ||
    is.numeric(x$time_value)) {
```
so I guess the linter is going to fail indefinitely. Disappointing and weird.

dsweber2 (approved):
Should we wait for Logan to get back? Wouldn't put it past him to have some arcane solution to the `identical` vs `all.equal` problems. I'd be fine with leaving that as an open issue for a potential future feature though. Actually using anons is mostly covered by passing args.

nmdefries:
David opened an issue RE styler [here](https://github.com/r-lib/styler/issues/1188)

Will have Logan take a look at this when he gets back, but if it needs changes, I'll do those in a separate PR. I don't want this to be hanging as a PR for a long time.

nmdefries:
- [ ] demonstrate use of new functions in vignettes

PR #441, opened 2024-04-26
Guess YYYYww numeric as "custom", fix incidental typo

brookslogan:
- Previously we guessed YYYYww numeric as "year", almost surely wrong. But we can't easily/reliably tell what type of yearweek it is, so just use "custom" to avoid various/bad guesses in time_type-reliant code.
- Don't use <int> <= <chr>; seemed nonimpactful in this case, at least in tested locale.

### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

- Previously we guessed YYYYww numerics as "year"s.  Probably wrong.
- Don't use '<= "10"' --- this didn't seem to have an impact, but still seems cleaner.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -403,7 +403,7 @@ guess_geo_type <- function(geo_value) {
 guess_time_type <- function(time_value) {
   # Convert character time values to Date or POSIXct
   if (is.character(time_value)) {
-    if (nchar(time_value[1]) <= "10") {
+    if (nchar(time_value[1]) <= 10L) {
       new_time_value <- tryCatch(
         {
           as.Date(time_value)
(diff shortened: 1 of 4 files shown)


R/utils.R
```diff
@@ -403,7 +403,7 @@ guess_geo_type <- function(geo_value) {
 guess_time_type <- function(time_value) {
   # Convert character time values to Date or POSIXct
   if (is.character(time_value)) {
-    if (nchar(time_value[1]) <= "10") {
+    if (nchar(time_value[1]) <= 10L) {
```
nmdefries, line 406: :clap:

nmdefries (approved):

Issue #250, opened 2022-12-02
Add a `complete` method for `epi_df`s


brookslogan:
Currently, we expect users to appropriately `tidyr::complete` their data appropriately before using with some `epiprocess` functionality, but `complete` on `epi_df`s drops `epi_df`-ness; we should stop that from happening.  (If we were to enforce key uniqueness, this might need to be done carefully or with some additional checks, but since we don't, it's probably just unconditionally `reclass`ing or `dplyr_reconstruct`ing).

brookslogan:
Adding a `complete` method may not be the "right" approach.  It might naturally fall out of fixing more core functionality like joins.  Related: #223.

brookslogan:
Migrating from #138.

> This might be a tool to use in #256; if we want to complete slide inputs, we may need a complete / gap-filling method for that works across both `epi_df`s and tibbles (depending on whether our slide inputs have been decayed into tibbles or are still `epi_df`s). `tidyr::complete` may or may not work for this already, but it's probably what we want to use. We're trying to avoid heavy tidyverts dependencies now.

dshemetov:
Added in #488

PR #471, opened 2024-06-21
Djm/as tibble attr

### LINKED ISSUE #467: Do we need `as_tibble()` to drop the class?
https://github.com/cmu-delphi/epiprocess/blob/7fd2119d5d116069297e06b70dbf3ff5c08c07bf/R/methods-epi_df.R#L11

* This makes epipredict internals painful. I can see that someone may want this behaviour, but it means any time someone calls a function (from another package) that _anywhere_ uses `as_tibble()`, it drops the class.

dajmcdon:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

- Resolves #467

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: epiprocess
 Title: Tools for basic signal processing in epidemiology
-Version: 0.7.11
+Version: 0.7.12
 Authors@R: c(
     person("Jacob", "Bien", role = "ctb"),
     person("Logan", "Brooks", email = "lcbrooks@andrew.cmu.edu", role = c("aut", "cre")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -34,6 +34,10 @@ Pre-1.0.0 numbering scheme: 0.x will indicate releases, while 0.x.y will indicat
   format.
 - Improved documentation web site landing page's introduction.
 
+## Cleanup
+- Added optional `decay_to_tibble` attribute controlling `as_tibble()` behavior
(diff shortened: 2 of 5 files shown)


brookslogan (approved):
Approach looks nice, we get to keep sane behavior for users that take it `as_tibble()` by its name.  I added a bit of roxygen, NEWS.md, and conditional attr printing to make it less mysterious.

Question: how do you want propagation for this attribute to work?  It looks like there is some amount of propagation already, e.g., surviving through `mutate()`, but it does not survive via `|> group_by(geo_value) |> ungroup()` (unlike our metadata).  (And there are still other cases (e.g., #250) where our metadata / `epi_df`ness doesn't survive when we'd want it to.)  If you want it to be hardier, then it probably belongs within `metadata` + a double-check of whether `reclass()` and `dplyr_reconstruct.epi_df()` do indeed propagate it then.

dajmcdon:
My thinking on grouping: since `as_tibble(regular_tibble)` drops the grouping, this should as well for now. I wanted the the deviations from expected behaviour to be as minimal as possible while being sufficient for my needs. I suppose an external user *could* want this functionality, but providing it seems likely to be very low on the pile of issues until someone asks.

brookslogan:
I didn't mean that we shouldn't drop the grouping with `as_tibble()`.  I meant that grouping & ungrouping (and likely various other operations) drops the `decay_to_tibble` attribute.

dajmcdon:
Oh, I see. Yes, I'm hopeful that there's no grouping on internally that will screw me up. But good to have a record here in case it does!

PR #472, opened 2024-06-21
refactor(epi_slide, epix_slide): remove `time_step`, remove `time_type`, unify time handling, require difftimes in `before` and `after`

### LINKED ISSUE #219: Allow `before=Inf` in `epi[x]_slide`

### LINKED ISSUE #246: Consider what to do when `guess_period` and default `time_step` disagree in `epix_slide`
It seems unnatural for the default `ref_time_values` to guess one period, then `time_step` to guess a time step different from this period.  Maybe we should favor one or the other's determination, or maybe we should raise an error.

### LINKED ISSUE #281: Remove, use, or require empty dots in `new_epi_df`
The dots (`...`) in `new_epi_df` are unused.  Misspelling argument names or forgetting the interface can result in swapping in defaults rather than raising an error.  (E.g., `new_epi_df(....., other_keys = "age_group")` --- this is not the current way to set `other_keys` metadata, but will not raise an error, instead continuing on without setting this metadata.)

We should either remove the dots arg altogether, use it in some way, or check that it is empty.  Last part could use `ellipsis::check_dots_empty()` if its messages are all right or `rlang::dots_n(...)` with custom messaging.

dshemetov:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

- Make `before = Inf` default
- Make sure `before` validation functions correctly
- Remove `time_step` arg
- Big refactor cc4a84e:
  - remove the time_type metadata from epi_df, remove time_type as an argument in many functions
  - require before and after to have units in certain time_type cases 

TODO:
- [x] get feedback
  - [x] especially need @nmdefries feedback on the slider-specific change in epi_slide_opt
- [x] fix remaining tests

- Resolves #219 
- Resolves #281 
- Resolves #246


R/grouped_epi_archive.R
```diff
@@ -247,19 +253,9 @@ epix_slide.grouped_epi_archive <- function(x, f, ..., before, ref_time_values,
   }
 
   # Validate and pre-process `before`:
-  if (missing(before)) {
-    cli_abort("`before` is required (and must be passed by name);
-                if you did not want to apply a sliding window but rather
-                to map `epix_as_of` and `f` across various `ref_time_values`,
-                pass a large `before` value (e.g., if time steps are days,
-                `before=365000`).")
+  if (!(anyInfinite(before) || testInt(before, lower = 0L, null.ok = FALSE, na.ok = FALSE))) {
```
brookslogan, line 256:
```suggestion
  if (!(identical(before, Inf) || test_int(before, lower = 0L))) {
```
- No longer accepts -Inf.
- Rest is stylistic.  Suggested here but not completely sure about omitting the redundant args to `test_int()` since it doesn't follow the defaults of the rest of the package (e.g., `test_integerish()`); guess we could accidentally trick ourselves either way.

NAMESPACE
```diff
@@ -95,6 +95,7 @@ importFrom(checkmate,check_atomic)
 importFrom(checkmate,check_data_frame)
 importFrom(checkmate,check_names)
 importFrom(checkmate,expect_class)
+importFrom(checkmate,testInt)
```
brookslogan, line 98:
```suggestion
importFrom(checkmate,test_int)
```

R/epiprocess.R
```diff
@@ -8,6 +8,7 @@
 #'             assert_logical assert_list assert_character assert_class
 #'             assert_int assert_numeric check_data_frame vname check_atomic
 #'             anyInfinite test_subset test_set_equal checkInt expect_class
+#'             testInt
```
brookslogan, line 11:
```suggestion
#'             test_int
```

R/grouped_epi_archive.R
brookslogan: todo: do / open issue to: allow difftimes directly as `before` arguments if the `version` is of a class compatible with difftimes, and maybe disallow int values if `version` is datetime-like.

suggest: also remove `time_step` from `epi_slide*` methods.  Maybe ask Nat first if he foresees gotchas and whether accepting difftimes is required to make some use cases possible.  If not removing in this PR, please open a separate issue.

todo: remove surviving `@param time_step` from `epix_slide` roxygen.

suggest: check over the vignettes (e.g., {slide,archive,advanced}.Rmd, maybe others) and update them if needed.  I briefly looked and didn't see anything that requires an update, but might be worth a second pass.

brookslogan (approved):
See comments; very minor stuff + a couple potential tasks that could be spun off into separate issues.  If you tackle the latter in this PR, might be worth a re-review from me or Nat.

R/grouped_epi_archive.R
dshemetov: - Definitely feeling low prio on the difftimes thing. Never worked with that data type.
- Will check about `time_step`. 
- Removed the docs line.
- Repo-wide search for `time_step` brings up nothing, so I think we're good on vignettes.

brookslogan: One thing that may add to prio of difftimes is that they seem natural for making time windows when the versions and/or time_values are datetimes, and integers don't seem natural or advisable for that.  But there's probably a lot more to think about with datetimes that we've neglected, and it may take some thinking to establish whether difftimes are the best thing to use for `before` in this case or not.

If you do remove difftime support --- and even if you include it with datetime in mind --- maybe double-check that it's not accidentally accepted by the validation code and added in some strange way to non-datetime stuff.

dshemetov: I just think we should stay focused on common use cases before generalizing. I have yet to see a version column that's not Date.

dshemetov: time_step in epi_slide seems functional and non-trivial, see `test-epi_slide.R` tests comparing epi_slide and epi_slide_mean. Doesn't really make sense to remove.

brookslogan (commented):
Overall approach seems good.  Design-wise I had some thoughts on what column we base before validation on in epix_slide().  Plus maybe found a bug.

brookslogan:
issue: local tests fail, with various complaints like
```
Error in `validate_version_bound(fill_versions_end, x$DT, na_ok = FALSE)`: fill_versions_end must have the same `class` vector as x$version, which has a `class` of "Date"
```
as well as
```
Error in `epix_merge(as_epi_archive(tibble::tibble(geo_value = "pa", time_value = d, 
    version = d + 1L, x_value = 1L)), as_epi_archive(tibble::tibble(geo_value = "pa", 
    time_value = tsibble::make_yearweek(2020, 1), version = d + 
        5L, y_value = 2L)))`: `x` and `y` were not equally up to date version-wise: `x$versions_end` was not identical to `y$versions_end`; either ensure that `x` and `y` are equally up to date before merging, or specify how to deal with this using `sync`
```

... and it looks like we're not running checks in CI somehow??

brookslogan:
@nmdefries I would also appreciate if you could take a pass.  I feel like I must have glossed over something given the check results, and was pausing until tomorrow but accidentally hit approve instead of comment.

R/epi_df.R
```diff
@@ -88,13 +86,13 @@ NULL
...
-#' correct metadata for an `epi_df` object (ie. `geo_type`, `time_type`, and `as_of`).
+#' correct metadata for an `epi_df` object (ie. `geo_type` and `as_of`).
 #' Refer to the below info. about the arguments for more details.
 #'
 #' @template epi_df-params
 #'
 #' @export
-new_epi_df <- function(x = tibble::tibble(), geo_type, time_type, as_of,
+new_epi_df <- function(x = tibble::tibble(), geo_type, as_of,
                        additional_metadata = list(), ...) {
```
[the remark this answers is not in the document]
dshemetov: Related #281

R/grouped_epi_archive.R
```diff
@@ -246,20 +252,10 @@ epix_slide.grouped_epi_archive <- function(x, f, ..., before, ref_time_values,
...
+  if (!checkmate::test_scalar(before)) {
+    cli_abort("`before` is a required scalar value.")
   }
-  before <- vctrs::vec_cast(before, integer())
-  assert_int(before, lower = 0L, null.ok = FALSE, na.ok = FALSE)
-
-  # If a custom time step is specified, then redefine units
-
-  if (!missing(time_step)) before <- time_step(before)
+  validate_slide_window_arg(before, guess_time_type(x$private$ungrouped$DT$version))
```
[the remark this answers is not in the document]
dshemetov: Since much of our API data has time_value and issue in the same data type, I'll just require them to be the same for now. We can add some relaxations for these constraints in another PR.

R/utils.R
```diff
@@ -395,57 +395,31 @@ guess_geo_type <- function(geo_value) {
...
+  if (inherits(time_value, "Date")) {
+    # Else, if a Date class, then use "week" or "day" depending on gaps.
     unique_time_gaps <- as.numeric(diff(sort(unique(time_value))))
-    # We need to check the modulus of `unique_time_gaps` in case there are
-    # missing dates. Gaps in a weekly date sequence will cause some diffs to
-    # be larger than 7 days. If we just check if `diffs == 7`, it will fail
-    # unless the weekly date sequence is already complete.
+    # Gaps in a weekly date sequence will cause some diffs to be larger than 7
+    # days, so check modulo 7 equality, rather than equality with 7.
     return(ifelse(all(unique_time_gaps %% 7 == 0), "week", "day"))
```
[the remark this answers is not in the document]
dshemetov: I'm thinking the logic here will be: if all of the gaps are 0 mod 7, then assume week, otherwise assume day, but give a warning if less than half of the unique time gaps are 1.

R/utils.R
```diff
@@ -395,57 +395,31 @@ guess_geo_type <- function(geo_value) {
...
     return("yearweek")
   } else if (inherits(time_value, "yearmonth")) {
     return("yearmonth")
   } else if (inherits(time_value, "yearquarter")) {
     return("yearquarter")
-  } else if (rlang::is_integerish(time_value) &&
-    all(nchar(as.character(time_value)) == 4L)) { # nolint: indentation_linter
+  } else if (rlang::is_integerish(time_value) && all(nchar(as.character(time_value)) == 4L)) {
     return("year")
   }
```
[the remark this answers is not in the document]
dshemetov: Rolling year into allowing generic integers seems reasonable. Hope it doesn't open the floodgates back to unconstrained type hell :)

R/slide.R
```diff
@@ -151,40 +154,13 @@ epi_slide <- function(x, f, ..., before, after, ref_time_values,
...
-    after <- 0L
-  }
-
-  # If a custom time step is specified, then redefine units
-  if (!missing(time_step)) {
-    before <- time_step(before)
-    after <- time_step(after)
+  # Validate and pre-process `before`, `after`
+  if (!checkmate::test_scalar(before) || !checkmate::test_scalar(after)) {
+    cli_abort("Both `before` and `after` are required scalar values.")
```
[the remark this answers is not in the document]
dshemetov: I can probably bring that back, it's convenient and not that bad to code, especially if we let go of the tsibble types.

brookslogan:
Sorry, I believe I said I'd take a look at the failing checks while @nmdefries finishes looking over the rest.
- There's a number of failing tests regarding `epi_df` grouping; this is a breaking change due to an upstream package update; rebase onto dev and you will get #480 to skip these tests for now.
- Some vignettes are failing to build with errors like
```
   Both `before` and `after` are required scalar values.

   Expected before to be a difftime with units in days.

   `...` must be empty.
   ✖ Problematic argument:
   • time_type = "day"
```
- [ ] Some of these might be addressed by loosening the restrictions as noted in some of the discussion above; others might require vignette updates to conform with the new interface.

In particular, along the loosening-restrictions route, I think we discussed the following but maybe only verbally; sorry if these are repeats:
- [ ] suggest: allow integerish time windows when `time_type` is `"day"`.
- [ ] suggest: in `guess_time_type`, don't infer "day" if `length(unique_time_gaps) != 0L` and it doesn't contain any values that are equal to 1.  The intent is to generate errors on monthly and yearly data.  You could also maybe require something like 10% rounded up of `unique_time_gaps` to be equal to 1 or something along these lines.

R/slide.R
```diff
@@ -631,8 +585,13 @@ epi_slide_opt <- function(x, col_names, f, ..., before, after, ref_time_values,
       }
     } else if (f_from_package == "slider") {
       for (i in seq_along(col_names_chr)) {
+        factor <- if (inherits(before, "difftime") && units(before) == "weeks") {
+          7
+        } else {
+          1
+        }
```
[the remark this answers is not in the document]
nmdefries: Yep, this should be good without any adjustment factor. The allowed `slider` functions construct a window by # of elements, not by date/index (`before` and `after` are "the number of values before or after the current element to include in the sliding window" ([docs](https://www.rdocumentation.org/packages/slider/versions/0.3.1/topics/summary-slide))).

nmdefries (commented):
`epi_slide_opt` and related look good aside from the `factor` issue Logan brought up. I did not look at other changes.

praise: This is quite the complex PR! Thanks for robustifying this, and all the associated investigation into time-handling.

dshemetov:
> suggest: in guess_time_type, don't infer "day" if length(unique_time_gaps) != 0L and it doesn't contain any values that are equal to 1. The intent is to generate errors on monthly and yearly data. You could also maybe require something like 10% rounded up of unique_time_gaps to be equal to 1 or something along these lines.

@lcbrooks the >10% equal to 1 thing seems too fuzzy to turn into an error. I think checking if `all(unique_time_gaps) >= 28` and then giving an error is better.

R/slide.R
```diff
@@ -434,8 +406,7 @@ epi_slide <- function(x, f, ..., before, after, ref_time_values,
 #'   # Remove a nonessential var. to ensure new col is printed
 #'   dplyr::select(geo_value, time_value, cases, cases_7dav = slide_value_cases) %>%
 #'   ungroup()
-epi_slide_opt <- function(x, col_names, f, ..., before, after, ref_time_values,
-                          time_step,
+epi_slide_opt <- function(x, col_names, f, ..., before = NULL, after = NULL, ref_time_values = NULL,
```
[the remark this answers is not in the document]
dshemetov: Thanks for catching, will update that page.

brookslogan:
> I think checking if all(unique_time_gaps) >= 28 and then giving an error is better.

Sounds good, but check parens here (`all(unique_time_gaps >= 28)`).

(In "normal" data I'd expect this to be similar to a `length(unique_time_gaps) == 0L || any(unique_time_gaps == 1)` approach.  For non-"normal" data.... not sure, not a priority.)

Issue #162, opened 2022-07-22
Consider dot-prefixing all parameter names in functions where we take in user-controlled named `...`


brookslogan:
Arose from [this discussion](https://github.com/cmu-delphi/epiprocess/issues/146#issuecomment-1192785302).  When the user can specify desired output columns or arguments to their own functions in `...`, there is the danger of ~partial~ [see clarification below; this can be the source of problems, but not our current problems] argument name matching assigning args the user meant to be part of `...` to one of our built-in parameters, and getting incorrect output or cryptic errors.  Ryan has already encountered this once.  The approach of `dplyr` and `purrr` is to put a dot prefix on all their built-in parameters, as users typically won't use dots as the first character in their own column or function parameter names.  We should consider doing the same in `epi_slide`, `epix_slide`, and any other similar functions.

brookslogan:
Actually, partial matching may not be an issue here; it looks like it's only an issue when we have optional parameters before dots:
``` r
dots_first = function(..., abcde=2) {
  cat("abcde:\n")
  print(abcde)
  cat("list(...):\n")
  print(list(...))
}
dots_last = function(abcde=2, ...) {
  cat("abcde:\n")
  print(abcde)
  cat("list(...):\n")
  print(list(...))
}
# `dots_first` avoids partial-match problems of `dots_last`:
dots_first(abc=10)
#> abcde:
#> [1] 2
#> list(...):
#> $abc # <------------- placed `abc` in dots, the more reasonable place
#> [1] 10
dots_last(abc=10)
#> abcde:  # <------------- matched `abc` to `abcde` parameter, less reasonable
#> [1] 10
#> list(...):
#> list()
# but of course doesn't avoid (rarer) exact-match issues. Regardless of where
# `...` appears in the function signature, calling with `abcde=10` will match to
# an `abcde` parameter rather than placing `abcde=10` into `...`. If `abcde` is
# a popular column name or `f` parameter name (e.g., `n`), then these exact
# matches get in the way.
dots_first(abcde=10) 
#> abcde:
#> [1] 10
#> list(...):
#> list()
dots_last(abcde=10)
#> abcde:
#> [1] 10
#> list(...):
#> list()
```

<sup>Created on 2022-07-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1) (then annotated)</sup>

brookslogan:
[...]

PR #477, opened 2024-06-26
Slide improvements pass

### LINKED ISSUE #186: "Promote" `other_keys` to be printed, a constructor parameter, and more clearly documented

### LINKED ISSUE #240: Consider allowing list-type output from `epi[x]_slide` inner computations
Currently, a workaround for the length/nrow requirement for the inner computations is to return a single-row data frame with a single list column containing the actual result, then do a `time_value`-grouped slice and maybe a `select` to drop the nongrouping key columns on the slide output.  This could be slightly less work if one could just say `data = list(<real result>)` for the computation rather than `data = tibble(result = <real result>)`.

However, we might want to watch out for users trying to achieve something like the `%>% mutate(tibble(col1=..., col2=...))` behavior, except with lists instead of tibbles.  Perhaps that's enough to dissuade us from implementing this, or maybe there's a way around it; e.g., accepting only unnamed lists, since the names would probably be dropped anyway.

Implementation might require a tweak of `as_list_col`, not just the validation code.

### LINKED ISSUE #446: [enh] promote `other_keys`
Can we move `other_keys` out of the `additional_metadata` List? This seems like, by far, the most common use case we've seen for this list. And it already has its own special handling in the `new_epi_df()` constructor.

### LINKED ISSUE #510: `epi_slide` and `epix_slide` args
I still find before/after confusing. It takes mental math and careful reading of the documentation. It's more general, but not really how our target users are likely to think.

I suggest we bring back `n` for the window size as well as `align = c("left", "center", "right")`. We can keep `before`/`after` as `NULL` and ensure there aren't any conflicts. Then immediately process `n`+`align` into the appropriate `before`+`after` pair.

brookslogan:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

- Resolves #240
- Resolves #446 
- Resolves #186 
- Resolves #510 
- Resolves (partially?) #162 
- Addresses some (but not all) naming issues here #146


brookslogan:
(Random implementation notes: I also investigated an `as_slide_computations()` approach that would give a length>=1 list of functions to apply in the tidyeval case and length-1 lists in the function and formula cases.  Pros of that way include reducing code duplication/complexity/divergence when validating computation outputs, and having more freedom if we want to try to detect and optimize some slide computations by transforming them to calls to (internals of) `epi_slide_opt()` (as dplyr grouped_df's already do or did at some point).  Cons of that way are that we may introduce some computational overhead in the function and formula cases (setting up, storing, munging `results_env`s --- this is assuming the loop over computations has been hoisted above the group and/or ref time values loops to enable other optimizations; if that's not the other case, then there's the overhead of repeatedly executing a loop over 1 function which might actually be relevant if you're doing an `epi_slide()` with an unrecognized simple function) + need for yet another argument (.data / results, not just .x --- either added user-side or via transformations which also have time/debuggability overhead) or code complexity to avoid this.)

brookslogan:
- [ ] doc parm deprecations

R/utils.R
```diff
@@ -283,22 +283,78 @@ as_slide_computation <- function(f, ...) {
...
+        } else {
+          cli_abort("
+            Problem with output of {.code
+            {rlang::expr_deparse(rlang::quo_get_expr(f[[quosure_i]]))}}; it
+            produced a result that was neither NULL, a data.frame, nor a vector
+            without unnamed entries (as determined by the vctrs package).
+          ", class = "epiprocess__invalid_slide_comp_tidyeval_output")
+        }
+      }
+      validate_tibble(new_tibble(as.list(results_env, all.names = TRUE)[results_names]))
```
brookslogan, line 357: issue?: in rare case that user redefines a col within a computation, we'll request same colname twice here.  Somehow gets de-duped down the line, but probably better to make some comments or fixes here.

brookslogan:
In retrospect, this round of changes isn't as helpful or necessary as I expected.

The main change here seems to be
- Defaulting to something like `names_sep = NULL`.  This requires a second round of changes (incoming) to really work smoothly.  This second pass is probably going to much more impactful.

There are some potential minor conveniences: 
- Multiple slide computations
- `:=` support
- Hiding a bit more easily from `epi_slide` output size checks (`as_list_col = TRUE` wasn't enough, because size checks happened before list conversion)..... but in cases where you want to do this, you still want to slice away duplicate results, still quite awkward.
  - Hiding from `epix_slide()` naming is maybe the same level of convenience as before, though it feels a bit more natural to me now to wrap in `list()` rather than `as_list_col = TRUE`.  Another thing that's actually addressed by round 2 and doesn't really need this round 1 except for the convenience of not adding column prefixes by default.
- Maybe a little more parallelism between outputting data frames out of tidyeval and nontidyeval approaches.  But they are still their own distinct things.
- Better parallelism with dplyr syntax/expectations, at least for advanced users.
- Potentially setting up better for the next item below.

What I should probably add on to this pass or another PR:
- Support for slide computations directly outputting the result of `arx_forecaster()` ("scalar" S3 things).

And what neither of these rounds are going to address:
- Lack of `epi_slide_{summarize,reframe}` functions.  This is what we might want for some geoaggregation and version-unfaithful forecasting applications..... but there are probably better ways requiring less work.

brookslogan:
- [ ] todo: see if can rebase to have a `names_sep = NULL` default commit

dshemetov:
We should probably bump the version to 0.9.0, since that's our way of signaling breaking changes. Should probably also tag the version prior to merging this in, to make it easy to pin.

dsweber2:
I was talking with Dmitry about this, and I would request any functions that are defined in-line inside other functions be moved outside and have their dependencies included as explicit arguments, instead of capturing their value from context. They make it hard to follow the thread of what a function is doing.

brookslogan:
I accidentally pushed some changes that introduced some more dot prefixes to try to prevent some confusion and edge cases.  But I think this also fixed some outdated error messages, so maybe it's a good thing.  (There are also some orthogonal but also renaming-related fixes mixed in as well.)

- [ ] todo: fix roxygen/pkgdown config to solve pkgdown issue

dshemetov:
- [ ] consider a rename of epi_slide / epix_slide (but let's try not to get stuck here)

R/utils.R
brookslogan: should be done

dshemetov (approved):
Good to go. Will merge to main after this. epipredict PR will follow soon after.

PR #480, opened 2024-07-09
test grouping only before tsibble 1.1.5

dsweber2:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

Not bumping the version or adding to NEWS because it only changes tests, and is strictly upstream changes.

I added a skip that only tests for grouping's presence if the version of `tsibble` is before 1.1.5 specifically.

### DIFF
--- a/tests/testthat/test-as_tibble-decay.R
+++ b/tests/testthat/test-as_tibble-decay.R
@@ -8,6 +8,7 @@ test_that("as_tibble checks an attr to avoid decay to tibble", {
 })
 
 test_that("as_tibble ungroups if needed", {
+  skip_if(packageVersion("tsibble") > "1.1.4")
   edf <- jhu_csse_daily_subset %>% group_by(geo_value)
   # removes the grouped_df class
   expect_identical(class(as_tibble(edf)), c("tbl_df", "tbl", "data.frame"))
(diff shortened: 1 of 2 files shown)


dsweber2:
I also pinned `tsibble` to 1.1.4 via `renv::install("tsibble@1.1.4")`, and the tests passed for that

brookslogan (approved):
Thanks, this looks good for repairing the tests.  If you haven't already, can you also open an issue to investigate whether we need to "overoverwrite"/avoid this behavior?

dsweber2:
Unless you have specific instances in mind where we would use this to ungroup, I don't think we need to (in which case, we should be using `ungroup` anyways). If you'd asked me whether `as_tibble` destroyed grouping before I ran into this, I would've assumed it wouldn't, and would in fact just be a null-op, like they overwrote it to be.

brookslogan:
I plan to include a fix for this in #477, returning dplyr's expectations, as `as_tibble` appears to be used within some join or summarize operation used within our slides.  I'm not sure whether it's possibly grouped at this point or not, but this is showing the whatever S3 magic is going on is not completely insulating dplyr from tsibble overwriting `as_tibble.grouped_df`, and I don't want to find out where it matters.  For what it's worth, it's actually my expectation that `as_tibble` would output a plain tibble; that seems to follow base R things I'm familiar with, and it's clearly seems to be what dplyr wants since there's an explicit `dplyr:::as_tibble.grouped_df` doing this.

Issue #391, opened 2024-01-10
Patches for dplyr type conversions


dsweber2:
Various functions will convert `epi_df`'s to `tibble`s. This is to catalog which of those have been fixed/don't need to be fixed

- [x] arrange
- [ ] complete
- [ ] distinct
- [x] filter
- [x] slice*
- [ ] glimpse (doesn't make sense)
- [x] mutate
- [x] pull (doesn't make sense)
- [ ] relocate (some design choices to be made)
- [ ] rename
- [x] select (PR #390)
- [ ] count, tally, (may not make sense)
- [x] add_count, add_tally
- [ ] summarize (doesn't make sense)
- [ ] n, cur_group, cur_group_id etc doesn't really make sense
- [ ] bind_cols (also doesn't work for ungrouped)
- [x] bind_rows
- [x] intersect, union etc
- [ ] inner_join

dsweber2:
we also need something simliar for `epi_archive`'s. `filter` for example doesn't work.

dshemetov:
Related #618

dshemetov:
We now have filter for epi_archives.

JavierMtzRdz:
I reviewed the unresolved functions in this list that are neither marked as nonsensical nor discussed elsewhere, as well as other functions I considered relevant. It is worth noting that `complete.epi_df` is already implemented, so I marked it as completed in the list. 
From this, I observed the following behaviour:
* `epi_df` is preserved in the following cases:
  * `relocate`: seems to work as expected (via `dplyr` generic).
  * `distinct`:  seems to work as expected  (via `dplyr` generic).
  * `fill`:  seems to work as expected  (`tidyr` generic preserves class).
* `epi_df` decays when using:
  * `drop_na`.
  * `pivot_*`.

For cases where `epi_df` is preserved, I do not think further changes are necessary. For the cases that decay, I have added a possible method in the [tidyr-met](https://github.com/cmu-delphi/epiprocess/tree/tidyr-met) branch. They work as follows:
* `drop_na`.  Always returns an epi_df, as it only subsets rows and does not remove essential key columns.
* `pivot_wider`. 
  * Preserves `epi_df` if `geo_value` and `time_value` columns remain in the output. `other_keys` metadata is automatically updated to exclude any keys that were pivoted away or removed.
  * Silently decays if `geo_value` or `time_value` are used in `names_from` and thus removed as columns.
* `pivot_longer`.
[...]

PR #488, opened 2024-07-20
Add `complete.epi_df` method so grouped completions don't drop epi_df…

brookslogan:
…ness

Requesting review from @dshemetov since we talked about S3 stuff recently + to point at #391 where David's already done some of the work we were talking about.  @nmdefries I'm not sure this really needs 2 people to review, but GitHub's not letting me unrequest.

[Actually this isn't ready for review yet.  I should rebase on top of the time type changes and re-run tests first.  And test the time type reinference.]

### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

Adds a `complete.epi_df` method that doesn't drop epi_dfness when grouped.

Too much context:

We've had to basically always recommend `complete()`ing before putting things into an `epi_df`, because of various things that could go wrong down the line.  While `epi_slide_opt()` avoids bad behaviors when running on un`complete()`d data, we still aren't free: there is still cmu-delphi/epipredict#362 to avoid and complete()ion for other purposes, such as sliding functions `epi_slide_opt()` doesn't handle (and `epi_slide()` does bad things on), filling with 0s, making `dplyr::lag` actually work, naturally, maybe fixing some other 3rd-party library stuff, and standardizing time availability across epikeys (though we probably actually work for that last case already).

Grouped completions drop `epi_df`ness; it's annoying to have to carefully place the `complete()` before `as_epi_df()` when tweaking `epi_df` construction (especially if using the new `as_epi_df` renaming conveniences where you need to remember what the old names were when writing the `complete()`), and even more annoying/error-inducing to try to convert back to an `epi_df` after losing `epi_df`ness (since it's easy to forget and annoying to provide metadata). 

- Part of #391; probably the most useful item on the list.


brookslogan:
Okay, think this is probably ready now.  I realize now I'm not taking advantage of the `time_type` attribute to e.g., set a default `period`, and there may be some overlap with the reasoning used to transform `before` for `epi_slide`; we might want to export some helper function that gives an object representing a time difference of 1 period given the time type (such that adding/subtracting it from corresponding `time_value`s "does the right thing").  (or alternatively a function taking an integerish (vector) number of time periods and a `time_type`, and outputs an object (vector) representing time differences).  For the sake of keeping this PR simple, I won't add this unless you think the `period` default or such a helper function would be a very handy feature.

dshemetov (approved):

PR #507, opened 2024-08-09
Remove `select.epi_df`

brookslogan:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
- [-] Makes sure to bump the version number in `DESCRIPTION`. Always increment
      number (the second number).
- [-] Describe changes made in NEWS.md, making sure breaking changes
  process.

- The code for `select.epi_df` looked suspicious/weird without explanatory comments, and I don't want us to start basing other changes on it without in-source comments to explain the details... but now it turns out `select.epi_df` seems to be completely unnecessary now.  (But at some point we needed #390... perhaps older dplyr versions resulted in a weird situation that no longer appears in more recent dplyr versions, but the fact that we needed it at some point makes me a bit nervous so:)
  - I've added a few more tests to try to make sure I'm not breaking things instead.
- While attempting to test on older dplyr versions, other broken tests kept popping up before I got to tests of `select`, so I didn't fully track down the story here.  But in response, I've bumped our dependency a little bit to 1.0.8, and adjusted some failing tests to work on that version (which had a more rigid `if_else`).

- Resolves #{issue number}


brookslogan:
Hi @dshemetov & @dsweber2, it'd be great if one of you could sanity-check this and/or verify this isn't breaking things downstream in exploration-tooling.

dshemetov (approved):
Tested with rolling_mean and rolling_sd functions, which previously had to reconstruct the epi_df after a select, and it works well without it.

brookslogan:
Awesome, thanks for the quick review!

PR #512, opened 2024-08-13
446 promote other_keys as an argument to `as_epi_df()`

### LINKED ISSUE #446: [enh] promote `other_keys`

dajmcdon:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).

- Promotes `other_keys` and deprecates `additional_metadata` as arguments to `as_epi_df()`
- One can still populate `additional_metadata` with `attr(epi_df, "additional_metadata") <- list(...)`.
- Nothing is broken internally, but this will likely break some `{epipredict}` vignettes and the tooling book.

- Resolves #446


R/epi_df.R
```diff
@@ -195,7 +191,7 @@ new_epi_df <- function(x = tibble::tibble(), geo_type, time_type, as_of,
   }
 
   # Apply epi_df class, attach metadata, and return
-  class(x) <- c("epi_df", class(x))
+  class(x) <- c(subclass, "epi_df", class(x))
```
brookslogan, line 194: question: when is `subclass` useful in a constructor for the main class?  Why wouldn't this just be handled in a separate constructor function?  I see this pattern in dplyr/tibble but also massive extensibility issues in dplyr.  It sort of makes sense in `vctrs::new_vctr` but somehow `new_{tibble,epi_df,epi_archive}` seem different.

musing: especially for `epi_archive`s, except for maybe some role-attachment subclass, I think the extensibility mechanism we most often would want is an "abstract class" that derives certain default operations from certain core operations, but doesn't provide the core operations, + various concrete classes (`{ram,parquet,merge,derived}_epi_archive`) implementing the core operations, or the type-constructor/wrapper thing we do in `grouped_epi_archive`.

R/epi_df.R
```diff
@@ -274,29 +270,26 @@ as_epi_df.tbl_df <- function(
...
+  assert_character(other_keys, null.ok = TRUE)
+  new_epi_df(x, geo_type, time_type, as_of, other_keys)
 }
 
 #' @method as_epi_df data.frame
 #' @rdname epi_df
 #' @export
-as_epi_df.data.frame <- function(x, as_of, additional_metadata = list(), ...) {
-  as_epi_df.tbl_df(x = tibble::as_tibble(x), as_of = as_of, additional_metadata = additional_metadata, ...)
+as_epi_df.data.frame <- function(x, as_of, other_keys = NULL, ...) {
```
brookslogan, line 280:
```suggestion
as_epi_df.data.frame <- function(x, as_of, other_keys = character(), ...) {
```

R/epi_df.R
```diff
@@ -274,29 +270,26 @@ as_epi_df.tbl_df <- function(
...
+  as_epi_df.tbl_df(x = tibble::as_tibble(x), as_of = as_of, other_keys = other_keys, ...)
 }
 
 #' @method as_epi_df tbl_ts
 #' @rdname epi_df
 #' @export
-as_epi_df.tbl_ts <- function(x, as_of, additional_metadata = list(), ...) {
+as_epi_df.tbl_ts <- function(x, as_of, other_keys = NULL, ...) {
```
brookslogan, line 287:
```suggestion
as_epi_df.tbl_ts <- function(x, as_of, other_keys = character(), ...) {
```

R/methods-epi_archive.R
```diff
@@ -112,11 +112,11 @@ epix_as_of <- function(x, max_version, min_time_value = -Inf, all_versions = FAL
...
       as_of = max_version,
-      additional_metadata = c(
-        x$additional_metadata,
-        list(other_keys = other_keys)
-      )
+      other_keys = other_keys
     )
+  if (length(x$additional_metadata) != 0L) {
+    attr(as_of_epi_df, "additional_metadata") <- x$additional_metadata
+  }
```
brookslogan, line 119: suggest: get rid of `additional_metadata` + handling for `epi_archive`s as well.  It doesn't serve a purpose there either.

brookslogan (commented):
Thanks for taking this on!  A few haphazard suggestions & a question about `subclass`.

R/archive.R
```diff
@@ -297,7 +300,8 @@ new_epi_archive <- function(
     compactify,
     clobberable_versions_start,
     versions_end,
-    compactify_tol = .Machine$double.eps^0.5) {
+    compactify_tol = .Machine$double.eps^0.5,
+    subclass = character()) {
```
dshemetov, line 304: What is the intended use case for this that can't be solved with `ea <- new_epi_archive(...); class(ea) <- c(subclass, class(ea))`? Also the term "canonical class" was vague to me, had to read the source to understand. Would be nice to document the intent in the function docstring.

R/epi_df.R
dajmcdon: I'm going to remove these as "not really necessary".

R/archive.R
dajmcdon: I'm going to remove these as "not really necessary".

brookslogan:
> On the other_keys = character(0): @brookslogan should the default be NULL or character(0)?
I'd favor `character(0)`; it seems cleaner to reason about type/class-wise.  If it needs to be `NULL` for some reason we might need to regenerate the example data structures.

dajmcdon:
The downside of `character()` is that assigning `metadata$other_keys <- character()` means that this component will always exist, where as `metadata$other_keys <- NULL` gives a list without the `other_keys` component. So I think the options are:

1. Default `character()`; `other_keys` is always present in metadata. Printing could ignore or always print the empty field.
2. Default `character()`; internals process it away so that `other_keys` is not present in metadata when there aren't any.
3. Default `NULL`; `other_keys` not present in metadata when none are specified. Requires less internal control flow than 2 to implement.

It sounds like you prefer 1 or 2?

brookslogan:
I prefer 1. over 3. because 3. required checking how `NULL` behaved with every function we used (it seemed to act right for most/all of our uses, but so would `character(0)`), seemed to me like it actually required more control flow (though maybe not with `checkmate`), and complicated the "type" of `other_keys`.  2. doesn't make sense to me, though I guess having multiple ways to say zero `other_keys` also doesn't sound great.

dajmcdon:
On 2., maybe I should be more explicit. Something like:

```{r}
as_epi_df <- function(..., other_keys <- character(0), ... ) {
  ...
  metadata$other_keys <- if (length(other_keys)) other_keys else NULL
}
```
In this case, no `other_keys` slot will be populated in the `metadata` list.

And you would need similar handling in `print()` to avoid lines like `* other_keys:     `.

brookslogan:
Yeah, for 2., that's what I was thinking of.  I think we had 2. up to some point, and I changed it, I think maybe for a concrete reason, but I'm not sure.  Could try to track it down.

Is the concern about keeping it in the metadata just about RAM space or mindspace?  On the mindspace-part, I feel it's a bit easier to think about than possibly-NULLness.

[Here](https://github.com/cmu-delphi/epiprocess/pull/390)'s where we made the switch from 3. to 1.  Perhaps this was one of the cases where NULL doesn't act like character(0) automatically.

dshemetov:
Playing around a bit, confused why I can't see `other_keys` in jhu_csse_daily_subset:

```r
r$> attributes(jhu_csse_daily_subset)$metadata
$geo_type
[1] "state"

$time_type
[1] "day"

$as_of
[1] "2024-01-26 09:27:32 PST"
```

brookslogan:
I think this might have been a missed chore from #390?  Somehow I thought we already noticed it and regenerated, but guess not.

PR #513, opened 2024-08-14
wip: rework slide window args

### LINKED ISSUE #510: `epi_slide` and `epix_slide` args

dshemetov:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

Prototype of new slide interface. Fixing tests takes a long time, want to make sure we agree on inner logic and deprecation handling before I tackle all that. The core logic may be easier to read outside the diffs, so reproducing here:

```r
  if (!is.null(before) || !is.null(after)) {
    cli_abort("`before` and `after` are deprecated for `epi_slide`. Use `n` instead.")
  }

  # Handle window arguments
  align <- match.arg(align)
  time_type <- attr(x, "metadata")$time_type
  validate_slide_window_arg(n, time_type)
  if (n == Inf) {
    if (align == "left") {
      before <- Inf
      if (time_type %in% c("day", "week")) {
        after <- as.difftime(0, units = glue::glue("{time_type}s"))
      } else {
        after <- 0
      }
    } else {
      cli_abort(
        "`epi_slide`: center and right alignment are not supported with an infinite window size."
      )
    }
  } else {
    if (align == "left") {
      before <- n - 1
      after <- 0
    } else if (align == "center") {
      # For n = 5, before = 2, after = 2. For n = 4, before = 2, after = 1.
      before <- floor(n / 2)
      after <- n - before - 1
    } else if (align == "right") {
      before <- 0
      after <- n - 1
    }
  }

```

- Resolves #510

### DIFF
--- a/R/autoplot.R
+++ b/R/autoplot.R
@@ -47,8 +47,8 @@ autoplot.epi_df <- function(
     .facet_by = c(".response", "other_keys", "all_keys", "geo_value", "all", "none"),
     .base_color = "#3A448F",
     .max_facets = Inf) {
-  .color_by <- rlang::match_arg(.color_by)
-  .facet_by <- rlang::match_arg(.facet_by)
+  .color_by <- rlang::arg_match(.color_by)
+  .facet_by <- rlang::arg_match(.facet_by)
 
   assert(anyInfinite(.max_facets), checkInt(.max_facets), combine = "or")
   assert_character(.base_color, len = 1)
--- a/R/autoplot.R
+++ b/R/autoplot.R
@@ -47,8 +47,8 @@ autoplot.epi_df <- function(
     .facet_by = c(".response", "other_keys", "all_keys", "geo_value", "all", "none"),
     .base_color = "#3A448F",
     .max_facets = Inf) {
-  .color_by <- match.arg(.color_by)
-  .facet_by <- match.arg(.facet_by)
+  .color_by <- rlang::match_arg(.color_by)
+  .facet_by <- rlang::match_arg(.facet_by)
 
   assert(anyInfinite(.max_facets), checkInt(.max_facets), combine = "or")
   assert_character(.base_color, len = 1)
--- a/R/slide.R
+++ b/R/slide.R
@@ -82,8 +82,7 @@ epi_slide <- function(
     .ref_time_values = NULL, .new_col_name = NULL, .all_rows = FALSE) {
   # Argument deprecation handling
   provided_args <- rlang::call_args_names(rlang::call_match())
-  old_arg_names <- c("x", "f", "ref_time_values", "new_col_name", "all_rows")
-  if (any(purrr::map_lgl(provided_args, ~ .x %in% old_arg_names))) {
+  if (any(purrr::map_lgl(provided_args, ~ .x %in% c("x", "f", "ref_time_values", "new_col_name", "all_rows")))) {
     cli::cli_abort(
       "epi_slide: you are using one of the following old argument names: `x`, `f`, `ref_time_values`,
       `new_col_name`, or `all_rows`. Please use the new dot-prefixed names: `.x`, `.f`, `.ref_time_values`,
@@ -430,8 +429,7 @@ epi_slide_opt <- function(
 
   # Argument deprecation handling
   provided_args <- rlang::call_args_names(rlang::call_match())
-  old_arg_names <- c("x", "col_names", "f", "ref_time_values", "all_rows")
-  if (any(purrr::map_lgl(provided_args, ~ .x %in% old_arg_names))) {
+  if (any(purrr::map_lgl(provided_args, ~ .x %in% c("x", "col_names", "f", "ref_time_values", "all_rows")))) {
     cli::cli_abort(
       "epi_slide_opt: you are using one of the following old argument names: `x`, `col_names`, `f`, `ref_time_values`,
       or `all_rows`. Please use the new dot-prefixed names: `.x`, `.col_names`, `.f`,
@@ -578,7 +576,7 @@ epi_slide_opt <- function(
   # using `names(pos)` directly for robustness and in case we later want to
   # allow users to rename fields via tidyselection.
   if (class(quo_get_expr(enquo(.col_names))) == "character") {
-    pos <- eval_select(all_of(.col_names), data = .x, allow_rename = FALSE)
+    pos <- eval_select(dplyr::all_of(.col_names), data = .x, allow_rename = FALSE)
   } else {
     pos <- eval_select(enquo(.col_names), data = .x, allow_rename = FALSE)
   }
@@ -757,8 +755,7 @@ epi_slide_mean <- function(
     .ref_time_values = NULL, .all_rows = FALSE, .na_rm = FALSE) {
   # Argument deprecation handling
   provided_args <- rlang::call_args_names(rlang::call_match())
-  dot_prefixed_args <- c("x", "col_names", "f", "ref_time_values", "all_rows")
-  if (any(purrr::map_lgl(provided_args, ~ .x %in% dot_prefixed_args))) {
+  if (any(purrr::map_lgl(provided_args, ~ .x %in% c("x", "col_names", "f", "ref_time_values", "all_rows")))) {
     cli::cli_abort(
       "epi_slide_mean: you are using one of the following old argument names: `x`, `col_names`, `f`, `ref_time_values`,
       or `all_rows`. Please use the new dot-prefixed names: `.x`, `.col_names`, `.f`,
@@ -843,8 +840,7 @@ epi_slide_sum <- function(
     .ref_time_values = NULL, .all_rows = FALSE, .na_rm = FALSE) {
   # Argument deprecation handling
   provided_args <- rlang::call_args_names(rlang::call_match())
-  dot_prefixed_args <- c("x", "col_names", "f", "ref_time_values", "all_rows")
-  if (any(purrr::map_lgl(provided_args, ~ .x %in% dot_prefixed_args))) {
+  if (any(purrr::map_lgl(provided_args, ~ .x %in% c("x", "col_names", "f", "ref_time_values", "all_rows")))) {
     cli::cli_abort(
       "epi_slide_sum: you are using one of the following old argument names: `x`, `col_names`, `f`, `ref_time_values`,
       or `all_rows`. Please use the new dot-prefixed names: `.x`, `.col_names`, `.f`,
--- a/R/slide.R
+++ b/R/slide.R
@@ -5,7 +5,7 @@
 #' for examples.
 #'
 #' @template basic-slide-params
-#' @param f Function, formula, or missing; together with `...` specifies the
+#' @param .f Function, formula, or missing; together with `...` specifies the
 #'   computation to slide. To "slide" means to apply a computation within a
 #'   sliding (a.k.a. "rolling") time window for each data group. The window is
 #'   determined by the `before` and `after` parameters described below. One time
@@ -21,10 +21,11 @@
 #'   If `f` is missing, then `...` will specify the computation.
 #' @param ... Additional arguments to pass to the function or formula specified
 #'   via `f`. Alternatively, if `f` is missing, then the `...` is interpreted as
-#'   an expression for tidy evaluation; in addition to referring to columns
-#'   directly by name, the expression has access to `.data` and `.env` pronouns
-#'   as in `dplyr` verbs, and can also refer to `.x`, `.group_key`, and
-#'   `.ref_time_value`. See details.
+#'   a ["data-masking"][rlang::args_data_masking] expression or expressions for
+#'   tidy evaluation; in addition to referring columns directly by name, the
+#'   expressions have access to `.data` and `.env` pronouns as in `dplyr` verbs,
+#'   and can also refer to `.x`, `.group_key`, and `.ref_time_value`. See
+#'   details.
 #'
 #' @template basic-slide-details
 #'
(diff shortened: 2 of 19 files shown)


brookslogan:
- issue: I think right & left align cases are backward.  The default in say frollmean is right = trailing.
- note: not currently shown in sketch above, but default align should be trailing (say, `align = c("right", "center", "left")`).
- todo: hear back from dajmcdon and/or search docs for whether we want/need to support non-right/center/left alignments.
- partially separate issue: adding `n` back as an arg would mean going to be back to having a naming conflicts if a user tries to pass an `n` arg to their computation and/or use data-masking tidyeval to straightforwardly define a column with name `n`.  In older epiprocess, ryantibs and I separately ran into one/both of these, and at least one of these cases had a very confusing error message.  I'm not sure if we want to try to incrementally move toward [dot-prefixing](https://github.com/cmu-delphi/epiprocess/issues/162) arg names here, then follow up with dot-prefixing in `epix_slide` separately.  Or see if it's possible to use other approaches to avoid the `n` name conflict; I think that's the main one to be worried about.  (`epi_slide_opt` wouldn't need & wouldn't tidyverse-idiomatically use dot prefixes, as it's not passing tidyeval through `...`.)

misc. suggests:
- `identical` for comparisons with `Inf`, like in existing code
- `rlang::arg_match` instead of `match.arg` for better error messages, if you also agree that partial matches (e.g., "c" for "center") should be forbidden
- maybe comment reminding us what kinds of things n can be somewhere?  e.g., I initially forgot it could be a difftime when reading through the "center" calc (it does seem to work properly with difftimes).

dajmcdon:
Responding to @brookslogan's questions above:

* I can't really think of a reason for non-left/center/right alignment. Though as soon as we drop support, one may appear...
* On the `n` naming issue, the dot-prefix is an option. Also things like `window_size` or `n_timesteps` would be OK. I think that the issue we want to solve is the mental overhead of thinking "7 day trailing window" and then having to put `before = 6`. But maybe someone else really wants `n` specifically?

dshemetov:
I started thinking "left window" rather than "align on window's left side", oops. I think right/center/left names are fine. I thought about suggesting `n` change to `window_size`, but I don't want to bikeshed these details. Adding dot prefixes to slide arguments is also reasonable, that's the suggestion made in the [tidy design guide](https://design.tidyverse.org/dots-prefix.html).

dshemetov:
- Went ahead a converted all match.arg to rlang::arg_match. Big fan of disabling partial matching in both arguments and values.
- Went ahead and prepended all arguments with . and setup a clear error message if the old ones are used.
- Addressed all other points, switched .n to .window_size and added documentation about what values is takes.

brookslogan:
musing: Sorry for all the trouble, I'm not sure dot prefixing / name conflict stuff was actually necessary here after all.  The root cause in the past might(??) have been the fact that we had an `n = 7` default.  But dot-prefixing/rare-naming in this PR does prevent some weird messages if someone tried to use `n`/`window_size` in `...` and forgot that they needed to provide the window size to `epi_slide`.  (I haven't been able to track down what the original thing that actually caused the issues Ryan(?) and I encountered was.  I'm starting to suspect the original issues were maybe in `epix_slide(n =)` which wouldn't be an issue presently anyway.)
- idea: if there are lingering problems from including it in this PR, seems like we can delay it some more since I'm having trouble even remembering/reproducing the old issue.  ("n" happens not to be the best symbol to try to search for ever.)

thoughts: re. naming: ... I don't know if these are bikeshedding or load-bearing signage.  With `epi_slide` no longer closely coupled with `epix_slide` interface, it's more likely to be bikeshedding than before.
- I expect ryantibs would strongly favor `n` and would accept `.n`.
- `n` is quite popular in a small set of packages I tested; this could be good or bad for `n`/`.n` but I don't have the energy to analyze it right now
<details>
```r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
tibble(pkgname = c("dplyr", "tidyr", "slider", "zoo", "xts", "data.table")) %>%
  mutate(member_value = lapply(pkgname, function(pkgname) as.list(asNamespace(pkgname)))) %>%
  mutate(member_name = lapply(member_value, names)) %>%
  tidyr::unchop(c(member_name, member_value)) %>%
  mutate(argname = lapply(member_value, function(x) {
    if (is.function(x)) {
      formalArgs(args(x))
    } else {
      character(0L)
    }
  }), member_value = NULL) %>%
  tidyr::unchop(argname) %>%
  filter(argname %in% c("n", ".n", "window", ".window", "window_size", ".window_size")) %>%
  print(n = 100L)
#> # A tibble: 74 × 3
#>    pkgname    member_name                argname
#>    <chr>      <chr>                      <chr>  
#>  1 dplyr      slice_min.data.frame       n      
#>  2 dplyr      vec_as_location_invert     n      
#>  3 dplyr      check_slice_unnamed_n_prop n      
#>  4 dplyr      sample_int                 n      
#>  5 dplyr      check_size                 n      
#>  6 dplyr      dplyr_new_data_frame       n      
#>  7 dplyr      shift_c                    n      
#>  8 dplyr      shift                      n      
#>  9 dplyr      check_slice_n_prop         n      
#> 10 dplyr      minimal_names              n      
#> 11 dplyr      ntile                      n      
#> 12 dplyr      slice_max                  n      
#> 13 dplyr      slice_tail                 n      
#> 14 dplyr      ntext                      n      
#> 15 dplyr      get_slice_size             n      
#> 16 dplyr      fmt_check_length_val       n      
#> 17 dplyr      progress_estimated         n      
#> 18 dplyr      slice_min                  n      
#> 19 dplyr      slice_tail.data.frame      n      
#> 20 dplyr      random_table_name          n      
#> 21 dplyr      shift_slice                n      
#> 22 dplyr      slice_max.data.frame       n      
#> 23 dplyr      nth                        n      
#> 24 dplyr      slice_sample               n      
#> 25 dplyr      lag                        n      
#> 26 dplyr      lead                       n      
#> 27 dplyr      top_n_rank                 n      
#> 28 dplyr      top_frac                   n      
#> 29 dplyr      slice_head                 n      
#> 30 dplyr      slice_sample.data.frame    n      
#> 31 dplyr      slice_head.data.frame      n      
#> 32 dplyr      top_n                      n      
#> 33 dplyr      check_length_val           n      
#> 34 dplyr      last_dplyr_warnings        n      
#> 35 tidyr      str_split_fixed            n      
#> 36 slider     vec_init_unspecified_list  n      
#> 37 slider     compute_from               n      
#> 38 slider     compute_to                 n      
#> 39 zoo        scale_y_yearqtr            n      
#> 40 zoo        scale_x_yearmon            n      
#> 41 zoo        make.par.list              n      
#> 42 zoo        yearqtr_trans              n      
#> 43 zoo        head.zoo                   n      
#> 44 zoo        scale_y_yearmon            n      
#> 45 zoo        yearmon_trans              n      
#> 46 zoo        scale_x_yearqtr            n      
#> 47 zoo        tail.zoo                   n      
#> 48 xts        shift.time                 n      
#> 49 xts        last.default               n      
#> 50 xts        naCheck                    n      
#> 51 xts        align.time.xts             n      
#> 52 xts        align.time.POSIXct         n      
#> 53 xts        shift.time.xts             n      
#> 54 xts        first.xts                  n      
#> 55 xts        last.xts                   n      
#> 56 xts        align.time.POSIXlt         n      
#> 57 xts        first.default              n      
#> 58 data.table froll                      n      
#> 59 data.table gtail                      n      
#> 60 data.table frollsum                   n      
#> 61 data.table first                      n      
#> 62 data.table g[                         n      
#> 63 data.table gshift                     n      
#> 64 data.table last                       n      
#> 65 data.table cedta                      n      
#> 66 data.table shift                      n      
#> 67 data.table head.data.table            n      
#> 68 data.table setalloccol                n      
#> 69 data.table alloc.col                  n      
#> 70 data.table frollapply                 n      
#> 71 data.table tail.data.table            n      
#> 72 data.table ghead                      n      
#> 73 data.table g[[                        n      
#> 74 data.table frollmean                  n
```

<sup>Created on 2024-08-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>
</details>

- I thought some time ago we found some Python/Rust library with names we liked for this and for `n_recent`?
- tidyeval guidelines probably don't apply to current `epi_slide_opt` since we aren't forwarding or tidyevaluating `...` (and they have `dplyr::slice_head(n =)` for example)... but it seems like it'd be nice for our ease of documentation and maybe also also better for users if we did consider including it in any impending dot prefixing PRs, for a more consistent set of arg names.

- [ ] todo: resolve current mix of `.n` and `.window_size` in current PR (I think `.window_size` is current arg name, but `.n` is in an error message somewhere).

dshemetov:
My personal take is that these names don't matter as much as clear documentation around them and we should spend our very finite research time elsewhere (though I appreciate the thoroughness you're bringing to this question Logan). I'm not going to weigh this discussion down with more opinions about `n` or `window_size`, I think we should pick one and not spend more time here.

This PR fixes the key window arithmetic issues Dan pointed out and incrementally moves this package towards a tidy design best practice. I can address the todos above and I propose we move ahead.

R/slide.R
```diff
@@ -88,62 +88,85 @@
 #'     before = 1, as_list_col = TRUE
 #'   ) %>%
 #'   ungroup()
-epi_slide <- function(x, f, ..., before = NULL, after = NULL, ref_time_values = NULL,
-                      new_col_name = "slide_value", as_list_col = FALSE,
-                      names_sep = "_", all_rows = FALSE) {
+epi_slide <- function(
+    x, f, ...,
```
brookslogan, line 92: suggest: also dot-prefix `x` and `f`, like in `purrr::map`

(maybe we lose some ease in searchability since there's .x into `epi_slide` and `.x` the pronoun-like thing in formulas and tidyeval, but that's shared by `purrr::map`)

R/slide.R
```diff
@@ -88,62 +88,85 @@
...
+    .window_size = 0, .align = c("right", "center", "left"), .ref_time_values = NULL,
+    .new_col_name = "slide_value", .as_list_col = FALSE, .names_sep = "_", .all_rows = FALSE,
+    before = NULL, after = NULL, ref_time_values = NULL, new_col_name = NULL,
+    as_list_col = NULL, names_sep = NULL, all_rows = NULL) {
+  if (any(map(c(n, before, after, ref_time_values, new_col_name, as_list_col, names_sep, all_rows), Negate(is.null)))) {
```
brookslogan, line 97: suggest: use something like `%in% rlang::call_args_names(rlang::call_match())` so if they try to use it as tidyeval they get a better error message.

issue: if not using above suggestion, `map` should be `map_lgl`

man/epi_slide_opt.Rd
```diff
@@ -59,44 +59,44 @@ these args via \code{...} will cause an error. If \code{f} is a \code{slider} fu
...
-depend on the type of the \code{time_value} column:
-\itemize{
-\item if it is a Date and the cadence is daily, then they can be integers
-(which will be interpreted in units of days) or difftimes with units
-"days"
-\item if it is a Date and the cadence is weekly, then they must be difftimes
-with units "weeks"
-\item if it is an integer, then they must be integers
-}
+\item{before, after}{Deprecated. Use \code{.n} instead.}
```
brookslogan, line 62: issue: looks like `@template` usage is making this documentation inaccurate.  Ideas: either
- ensure both `epi_slide` and `epi_slide_{opt,mean,sum}` have appropriate parameter documentation --- this means they shouldn't use the same `@template` to document `before`, `after`, `.n`, `.align`, maybe some others
- change `epi_slide_opt` to use `.n` and `.align`, or
- change `epi_slide_opt` to use `n` and `align` (like e.g., `slice_head`) PLUS fix the `@template` issue

R/autoplot.R
```diff
@@ -47,8 +47,8 @@ autoplot.epi_df <- function(
     .facet_by = c(".response", "other_keys", "all_keys", "geo_value", "all", "none"),
     .base_color = "#3A448F",
     .max_facets = Inf) {
-  .color_by <- match.arg(.color_by)
-  .facet_by <- match.arg(.facet_by)
+  .color_by <- rlang::match_arg(.color_by)
+  .facet_by <- rlang::match_arg(.facet_by)
```
brookslogan, line 51: it's `rlang::arg_match` --- `rlang`, `vctrs`, and `purrr` often try to use a convention of putting a sort of generalized "type" descriptor for the main argument as the first "word" in a function name.

brookslogan (commented):
Browsing review requests and wasn't sure if this was still wip or not.  Just dropped a few comments but realized it's probably still wip.  If it's ready, please re-request!

R/slide.R
dshemetov: sure yea, prefixing them sounds reasonable.

(huh, interesting, well intuitively most of my uses of purrr::map it's piped into, so that takes care of the first argument. using purrr::map with a do.call type construct with list(.x=...) seems really unlikely.)

man/epi_slide_opt.Rd
dshemetov: Updated all the documentation that I could find. I'm sure there will be things I missed. Will fix issues as I spot them on the site in the documentation pass coming up.

dshemetov:
Updated all the documentation that I could find, including vignettes. I'm sure there will be things I missed. Will fix issues as I spot them on the site in the documentation pass coming up.

PR #517, opened 2024-08-23
refactor: simplify ungrouped epix_slide

dshemetov:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

Mostly just copied the existing `epix_slide` code, but removed the group_by concerns. The function is simpler and a bit more readable now.

- Resolves #{issue number}

### DIFF
--- a/R/methods-epi_archive.R
+++ b/R/methods-epi_archive.R
@@ -833,22 +833,116 @@ epix_slide.epi_archive <- function(
     ref_time_values = NULL,
     new_col_name = NULL,
     all_versions = FALSE) {
-  # For an "ungrouped" slide, treat all rows as belonging to one big
-  # group (group by 0 vars), like `dplyr::summarize`, and let the
-  # resulting `grouped_epi_archive` handle the slide:
-  epix_slide(
-    group_by(x),
-    f,
-    ...,
-    before = before, ref_time_values = ref_time_values, new_col_name = new_col_name,
-    all_versions = all_versions
-  ) %>%
-    # We want a slide on ungrouped archives to output something
-    # ungrouped, rather than retaining the trivial (0-variable)
-    # grouping applied above. So we `ungroup()`. However, the current
-    # `dplyr` implementation automatically ignores/drops trivial
-    # groupings, so this is just a no-op for now.
-    ungroup()
+  ### START Copy pasta from grouped_epi_archive ###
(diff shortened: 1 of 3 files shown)


brookslogan:
Sorry, behind on PR reviews.  This sounds nice for thinking about things, but scary for maintenance (code duplication & having to update multiple places).  I would prefer if we maybe tagged this or, if group agrees, removed the grouping features, so we only have to update 1 epix_slide per update.

dshemetov:
FWIW, when it comes to updating code in two or three places, I'm not that against it on principle, as long as the code comments point to similar code. But I'd also be happy to consider removing the grouped features.

dsweber2:
Probably better to build shared functions and then call those twice (e.g. `check_inputs` eating all the things that are... checking inputs... would cover like 90% of the redundancy here with 0 code duplication).

R/methods-epi_archive.R
```diff
@@ -833,22 +833,116 @@ epix_slide.epi_archive <- function(
...
-    ...,
-    before = before, ref_time_values = ref_time_values, new_col_name = new_col_name,
-    all_versions = all_versions
-  ) %>%
-    # We want a slide on ungrouped archives to output something
-    # ungrouped, rather than retaining the trivial (0-variable)
-    # grouping applied above. So we `ungroup()`. However, the current
-    # `dplyr` implementation automatically ignores/drops trivial
-    # groupings, so this is just a no-op for now.
-    ungroup()
```
brookslogan, line 851: note: this comment and `ungroup()` don't make sense now, since we stopped grouping the `epix_slide` output a while ago.

R/methods-epi_archive.R
```diff
@@ -833,22 +833,116 @@ epix_slide.epi_archive <- function(
...
+  ### START Copy pasta from grouped_epi_archive ###
```
brookslogan, line 836: idea: also cross-ref in epix_slide.grouped_epi_archive if not already, to remind we need to update here?  Though it's easy to miss such comments in such a long function. 

Sorry, I've let this PR get out of sync with slide updates, so this copy-pasta would need updated as well, rather than along the way.  I was hoping we'd decide on whether to remove grouping capabilities first, but that got punted.  You feel strongly that the alternate code here is worth the copy-pasta?

PR #519, opened 2024-08-24
wip: enforce epi_slide to work on grouped epi_dfs only

### LINKED ISSUE #154: Warn if a non-`version` key is duplicated in `epi_df` or `epi_archive`
Currently, we do not check for distinct key values. This has some [potential uses](https://github.com/cmu-delphi/epiprocess/issues/89) but also introduces the opportunity to form unexpected or invalid archives. Continue allowing this behavior, but warn the user if duplicate keys exist when creating or adding to an archive or an `epi_df`.

dshemetov:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

- epi_slide requires the epi_df to be grouped on all key_cols (automatically groups too)
- refactored epi_slide for readability 
- epi_slide_one_group is now defined outside epi_slide (as per @dsweber2's [comment](https://github.com/cmu-delphi/epiprocess/pull/477#issuecomment-2307775099))
- epi_slide_one_group completes the time index when sliding (matching epi_slide_opt); grouping also simplifies many of the internals
- huge test coverage improvement for epi_slide - includes all valid parameter combinations
- add `sum_group_epi_df` helper function to help aggregate across geos and other columns (since we don't allow epi_slide to handle that functionality anymore)

- Resolves #154


R/methods-epi_df.R
```diff
@@ -383,3 +383,36 @@ arrange_canonical.epi_df <- function(x, ...) {
...
+    dplyr::summarize(!!(value_col) := sum(!!sym(value_col))) %>%
+    ungroup() %>%
+    {
+      if (!"geo_value" %in% group_cols) {
+        mutate(., geo_value = "total") %>% relocate(geo_value, .before = 1)
+      } else {
+        .
+      }
+    } %>%
+    as_epi_df(as_of = attr(.x, "metadata")$as_of)
```
brookslogan, line 417: issue: this probably aborts if `"time_value"` isn't in the `group_col`s; we should either
- require/auto-add it to the group cols or
- allow the result to be a tibble in this case.

issue: `other_keys` won't be set properly.
- If `group_cols` are a subset of `"geo_value"`, `"time_value"`, and `.x`'s `other_keys`, then the result's `other_keys` should be the non-geo-time grouping cols.
- If `group_cols` include other things, then I'm not sure what we'd do.  Easiest is to just not let them.  I'm trying to think when they might want to:
  - If user is trying to group by season/wave/quarter/wday: this seems plausible.  We might only want an `epi_df` as the result when aggregating by season, though, and then (something of an accepted time type standing in for) the season would become the time_value rather than the season becoming one of the `other_keys`.

R/methods-epi_df.R
```diff
@@ -383,3 +383,36 @@ arrange_canonical.epi_df <- function(x, ...) {
...
+#' `value` column, and returning an `epi_df`. If aggregating over `geo_value`,
+#' the resulting `epi_df` will have `geo_value` set to `"total"`.
+#'
+#' @param .x an `epi_df`
+#' @param value_col character name of the column to aggregate
+#' @param group_cols character vector of column names to group by
+#' @return an `epi_df` object
+#'
+#' @export
+aggregate_epi_df <- function(.x, value_col = "value", group_cols = "time_value") {
```
brookslogan, line 399: question: why `aggregate_epi_df` instead of making an `aggregate.epi_df` implementation?

suggest: rename `value_col` to `counts_col` or something like that, or rename this function `sum_epi_df` or something like that

suggest: support multiple count columns and make it `counts_cols` or something like that

R/methods-epi_df.R
brookslogan: musing: it was a bit complicated to think about whether we get the right `geo_type`.  I think we do & this is how, but it seems convoluted.  Perhaps worth a source comment.
- if grouping by at least `geo_value` and `time_value`, then we should have the same `geo_type` as `.x`; we feed an `epi_df` with the right `geo_type` into `as_epi_df` and it trusts our metadata.
- if not grouping by `geo_value`, then the `geo_type` should be `"custom"`; we get this by `summarize` decaying to a tibble &`as_epi_df` setting the `geo_type` as "custom".

R/slide.R
```diff
@@ -111,8 +121,23 @@ epi_slide <- function(
...
+  if (checkmate::test_class(.x, "grouped_df")) {
+    expected_group_keys <- .x %>%
+      key_colnames() %>%
+      kill_time_value() %>%
+      sort()
+    if (.x %>% groups() %>% as.character() %>% sort() != expected_group_keys) {
+      cli_abort(
+        "epi_slide: `.x` must be either ungrouped or grouped by {expected_group_keys}. You may need to aggregate
+        your data first, see aggregate_epi_df().",
+        class = "epi_slide__invalid_grouping"
```
brookslogan, line 135:
```suggestion
        class = "epiprocess__epi_slide__invalid_grouping"
```
iirc rlang/tidyverse suggests including package name in custom error classes

R/epi_df.R
```diff
@@ -309,3 +323,7 @@ as_epi_df.tbl_ts <- function(x, as_of, other_keys = character(), ...) {
 is_epi_df <- function(x) {
   inherits(x, "epi_df")
 }
+
+group_epi_df <- function(x) {
+  x %>% group_by(group_by(across(all_of(kill_time_value(key_colnames(.))))))
```
brookslogan, line 328:
```suggestion
  x %>% group_by(across(all_of(kill_time_value(key_colnames(.)))))
```
suggest: more descriptive name; if still trying to be short, maybe `epigroup_epi_df`

R/epi_df.R
```diff
@@ -277,6 +277,20 @@ as_epi_df.tbl_df <- function(
   }
 
   assert_character(other_keys)
+
+  # Check one time_value per group
+  duplicated_time_values <- x %>%
+    group_by(across(all_of(c("geo_value", "time_value", other_keys)))) %>%
+    dplyr::summarize(n = dplyr::n(), .groups = "drop") %>%
+    filter(n > 1)
```
brookslogan, line 285:
```suggestion
  if ("n" %in% other_keys) {
    cli_abort('`other_keys` can't include "n"') 
  }
  duplicated_time_values <- x %>%
    count(across(all_of(c("geo_value", "time_value", other_keys)))) %>%
    filter(n > 1)
```
or
```suggestion
  if ("n" %in% other_keys) {
    cli_abort('`other_keys` can't include "n"') 
  }
  duplicated_time_values <- x %>%
    group_by(across(all_of(c("geo_value", "time_value", other_keys)))) %>%
    filter(dplyr::n() > 1) %>%
    ungroup()
```
if you want to actually show data

brookslogan: todo: we need to error or decay to tibble when we violate this in other operations.  That means updating `dplyr_reconstruct.epi_df`.

R/grouped_epi_archive.R
```diff
@@ -270,11 +270,14 @@ epix_slide.grouped_epi_archive <- function(
...
   }
 
-  validate_slide_window_arg(.before, .x$private$ungrouped$time_type)
+  validate_slide_window_arg(.before, .x$private$ungrouped$time_type, lower = 0) # nolint: object_usage_linter
 
   checkmate::assert_string(.new_col_name, null.ok = TRUE)
   if (identical(.new_col_name, "time_value")) {
-    cli_abort('`new_col_name` must not be `"time_value"`; `epix_slide()` uses that column name to attach the `ref_time_value` associated with each slide computation') # nolint: line_length_linter
+    cli_abort(
+      '`new_col_name` must not be `"time_value"`; `epix_slide()` uses that column name
```
brookslogan, line 278:
```suggestion
      '`.new_col_name` must not be `"time_value"`; `epix_slide()` uses that column name
```
(maybe don't hunt around for other things like this not touched by diffs in this PR; I've already done some of that in #521)

R/slide.R
```diff
@@ -111,8 +121,23 @@ epi_slide <- function(
...
 
-  # Function body starts
+  # Validate arguments
   assert_class(.x, "epi_df")
+  if (checkmate::test_class(.x, "grouped_df")) {
+    expected_group_keys <- .x %>%
+      key_colnames() %>%
+      kill_time_value() %>%
+      sort()
+    if (.x %>% groups() %>% as.character() %>% sort() != expected_group_keys) {
```
brookslogan, line 131:
```suggestion
    if (!identical(.x %>% group_vars() %>% sort(), expected_group_keys)) {
```

R/slide.R
```diff
@@ -111,8 +121,23 @@ epi_slide <- function(
...
+      cli_abort(
+        "epi_slide: `.x` must be either ungrouped or grouped by {expected_group_keys}. You may need to aggregate
+        your data first, see aggregate_epi_df().",
+        class = "epi_slide__invalid_grouping"
+      )
+    }
+  } else {
+    .x <- group_epi_df(.x)
```
brookslogan, line 139: issue: this auto-grouping could cause confusion while `epi_slide` and `epix_slide` still have such similar names, since `epix_slide` doesn't have (and shouldn't have) auto-grouping.

R/slide.R
```diff
@@ -208,129 +206,198 @@ epi_slide <- function(
...
+  # epi_slide_one_group, then we end up in a situation where the result as no
+  # new columns at all. This is a very fragile solution, I might even just
+  # remove it and error instead.
+  result <- group_modify(.x, epi_slide_one_group_partial, ..., .keep = FALSE)
+  if (ncol(result) == ncol(.x)) {
+    cli_warn(
+      "epi_slide: no new columns were created. This can happen if every group has no available ref_time_values.
+      In this case we return your epi_df but with an all-NA column 'slide_value' or what you provided in
+      `.new_col_name`. If your computation returned data.frames you may not get the expected column names.",
+      class = "epiprocess__epi_slide_no_new_columns"
```
brookslogan, line 237: question: what happens if user tries `epi_slide(.window_size = 7, ~ sum(.x$value), .new_col_name = "value")`?  Do we warn + overwrite their overwriting with NAs?

R/slide.R
```diff
@@ -121,60 +146,27 @@ epi_slide <- function(
...
-        "`ref_time_values` must be a unique subset of the time values in `x`.",
-        class = "epi_slide__invalid_ref_time_values"
-      )
-    }
-
-    if (anyDuplicated(.ref_time_values) != 0L) {
-      cli_abort(
-        "`.ref_time_values` must not contain any duplicates; use `unique` if appropriate.",
+        "epi_slide: `ref_time_values` must be a unique subset of the time values in `x`.",
         class = "epi_slide__invalid_ref_time_values"
```
brookslogan, line 153:
```suggestion
        class = "epiprocess__epi_slide__invalid_ref_time_values"
```

R/grouped_epi_archive.R
dshemetov: Sounds good. My plan is to make multiple passes through the docs for details like this in the coming weeks.

R/slide.R
```diff
@@ -208,129 +206,198 @@ epi_slide <- function(
...
+      before = window_args$before,
+      after = window_args$after,
+      ref_time_values = .ref_time_values,
+      all_rows = .all_rows,
+      new_col_name = .new_col_name,
+      used_data_masking = used_data_masking,
+      time_type = time_type,
+      complete_only = .complete_only
+    )
+  }
```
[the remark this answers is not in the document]
dshemetov: just a minor personal preference for making my own partial functions rather than letting some downstream code handle it. minor distrust of other code xD.

i prefer to keep epi_slide_one_group outside - one thing i don't like about keeping a giant work function inside another function is needing to scroll down through the whole body of `epi_slide_one_group` to see what the end of `epi_slide` is doing. i also just find it more natural to read: the broad structure of epi_slide as (1) validation, (2) group_modify with group-specific code, (3) stitching together is much clearer to me this way.

R/slide.R
dshemetov: hm, good point, yea this will probably overwrite their old values. i'll add a check for that.

R/methods-epi_df.R
dshemetov: issue 1: fixed. 
issue 2: i'm going to assume group_cols are a subset of key_colnames(.x) for now and make sure other_keys are set properly. the main goal of this function is to have a built-in workaround for people that wanted to use the old epi_slide as a way to aggregate geos while sliding (with this they can aggregate by non-geo_value and then slide). don't want to lose focus by getting into up/down scaling time resolution and up/down scaling geo resolution here.
musing: good point, will make sure we inherit the geo_type if geo_value is in group_cols

dshemetov: side-note, we should figure out if we want to maintain this geo_type metadata inside an epi_df versus just writing some helper functions that can up/down scale the geo resolution and recognize geo_type, etc.

R/methods-epi_df.R
dshemetov: > why aggregate_epi_df instead of making an aggregate.epi_df implementation
just a placeholder name, didn't think about it too much. i tend to think about dispatch when i have 2 or more types i want to dispatch over and here i just have one. just `aggregate` seems too generic, though i admit `aggregate_epi_df` isn't much better. maybe `group_by_aggregate`?

dshemetov: question: how do i generalize this to allow a character vector in sum_cols?
```R
dplyr::summarize(!!(sum_cols) := sum(!!sym(sum_cols)))
```

R/epi_df.R
dshemetov: I'll add the guard, but I also renamed the `n` to something that will be less likely  `.time_value_counts`. Not too worried about performance rn, but we can easily refactor this section if it ever comes to that.

dshemetov: > todo: we should error or decay to tibble when we violate this in other operations

we should talk through the ways this can get violated in other operations. a bit wary of decaying to tibble - silently losing epi_df-ness seems like it could lead to unpleasant surprises. will come back to this when i address other points.

dshemetov:
A thing to think through: if .all_rows = TRUE and .complete_only = TRUE, do we want to return the time_values on the left and right boundaries that don't have complete windows? Currently we don't and _I think_ I did that to match slider behavior, but I'll need to check.

R/slide.R
dshemetov: Hm, I'm thinking: the user intention here is clearly to overwrite their original values, so overwriting is fine. 

I do however think that we should validate `.new_col_name` a little more than we do now. I'm leaning towards "error if `any(.new_col_name %in% c("geo_value", "time_value"))`". This way, we make sure epi_slide returns an epi_df, but allow people to overwrite their own columns if they wish (since we don't modify their input object, we're not at risk of them losing their values completely here). While it would be nice to warn them if they overwrite, those warnings would get very annoying for users that intend that, so I don't think we should go down the road of warning.

dshemetov:
A tricky issue with data.frame returns in `.f`:
- how do we validate the return schema of `.f` to make sure it doesn't collide with `geo_value, time_value`?
  - we're forced to make the validation inside `epi_slide_one_group`, which is undesirable, but fine, but this leads into the next issue
- how do we get the return schema from `.f` when `epi_slide_one_group` doesn't have any `available_ref_time_values` (i.e. for whatever reason, the data provided doesn't have a sufficient window to run `.f`)?
  - if at least one group has enough data, then the validation can happen there
  - but what to do when none of the groups have enough data? I don't see a general way to introspect into the output of `.f` here, so I think we're forced to error
  - our current solution of detecting this case (happens right after the group_modify call) will just return an epi_df with the slide_value column, ignoring the schema that the data.frame from `.f` would have created

R/epi_df.R
dshemetov: hm, getting a call stack overflow cause i use dplyr functions in the duplicated time values check above. do i really have to implement my very own group_by for this?

brookslogan (commented):
Making a lot of progress here!  Does testing overhaul need much review?  I've glossed over it for now.

This looks good but I'd like to check any changes made to preserve the `epi_df` unique key invariant & any changes to decaying behavior.

tests/testthat/test-epi_slide.R
```diff
@@ -1,357 +1,500 @@
...
+  )
 }
-compute_slide_external_g <- function(.window_size) {
-  toy_edf_g %>%
-    mutate(slide_value = slider::slide_sum(.data$value, before = .window_size - 1)) %>%
-    dplyr::arrange(geo_value, time_value) %>%
-    as_epi_df(as_of = test_date + 100)
+get_test_units <- function(time_type = "day") {
+  switch(time_type,
+    day = as.difftime(1, units = "days"),
```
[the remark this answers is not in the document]
dshemetov: hm, good point, I can add that in.

R/slide.R
```diff
@@ -8,27 +8,37 @@
...
 #'   for tidy evaluation; in addition to referring columns directly by name, the
 #'   expressions have access to `.data` and `.env` pronouns as in `dplyr` verbs,
-#'   and can also refer to `.x`, `.group_key`, and `.ref_time_value`. See
-#'   details.
+#'   and can also refer to `.x` (not the same as the input epi_df),
+#'   `.group_key`, and `.ref_time_value`. See details.
 #' @param .new_col_name String indicating the name of the new column that will
 #'   contain the derivative values. Default is "slide_value"; note that setting
 #'   `new_col_name` equal to an existing column name will overwrite this column.
+#' @param .complete_only Logical; if `TRUE`, only slide values that have a
```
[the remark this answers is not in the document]
dshemetov: > different from what we get out of epi_slide_sum, which outputs rows with explicit NAs by default

have some TODOs above the consistency tests for `epi_slide/epi_slide_opt` (line ~300 in test-epi_slide.R):

```
  # These are the consistency tests between epi_slide and epi_slide_opt
  # functions. Only the specific case of .complete_only = FALSE and the opt
  # functions using na.rm = TRUE is testsed (the two options are equivalent for
  # our purposes here).
  # TODO: See if we can include the .complete_only = TRUE case in the future.
  # TODO: Add a case where the data contains NA values (not just gaps in time_value).
```

agree that it would be nice to get something consistent between the two.

> todo: note the default value of .complete_only in docs here.

done

> todo: also make the grouping demands consistent between epi_slide and epi_slide_opt; e.g., the former currently will auto-group an ungrouped epi_df, while the latter refuses to continue

will do. haven't propagate changes from epi_slide to opt yet, but definitely intended.

> note: .complete_only = TRUE seems significantly slower than FALSE

hm, we should definitely spend some time to address this (though let's do it after we settle the questions about what behaviors from epi_slide_opt we want to imitate; our decisions there might make things easier). `slider::slider_hop_index` doesn't help us unfortunately, because its `complete = TRUE` argument only applies to windows on the boundary (iirc), so I had to implement my own (and unfortunately slow) check for window completeness. one option is just to imitate epi_slide_opt and create a contiguous index by default and then we don't need to check for completeness (users can forward na.rm to their `.f` if they desire).

R/methods-epi_df.R
```diff
@@ -383,3 +383,55 @@ arrange_canonical.epi_df <- function(x, ...) {
...
+    group_by(across(all_of(group_cols))) %>%
+    dplyr::summarize(across(sum_cols, sum), .groups = "drop")
+  out <- if (!"geo_value" %in% group_cols) {
+    out %>%
+      mutate(geo_value = "total") %>%
+      relocate(geo_value, .before = 1)
+  } else {
+    out
+  }
+  out <- if ("time_value" %in% group_cols) {
```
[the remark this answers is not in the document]
dshemetov: Sorry, yea, I took this out as you were reviewing. Left overs from when I thought this might decay to tibble.

R/slide.R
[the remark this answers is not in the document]
dshemetov: Hm I see your point, so I inlined it as a lambda again, but kept the variables in. I mostly just disliked using `group_modify` to send arguments to the lambda `.f` because with `.keep = FALSE` there, we're mixing arguments the lambda and group_modify itself. Let me know if that feels better.

R/slide.R
```diff
@@ -208,129 +206,198 @@ epi_slide <- function(
...
 
-    slide_values <- vctrs::list_unchop(slide_values_list)
+  # Get stateful function that tracks ref_time_value per group and sends it to
+  # `f` when called.
+  f <- f_factory(available_ref_time_values)
 
+  if (time_type == "yearmonth" && identical(before, Inf)) {
+    starts <- rep(-Inf, length(available_ref_time_values))
+    stops <- available_ref_time_values + after
+  } else {
```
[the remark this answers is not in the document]
dshemetov: Ah yea, good comment. I got this to work and didn't even realize that yearweek ended up casting Inf to yearweek here. Seems like we're on the rough edges of their code, but it works thankfully.

R/slide.R
```diff
@@ -208,129 +206,198 @@ epi_slide <- function(
...
-      # (Loose) check on number of rows:
-      if (vctrs::vec_size(slide_values) != num_ref_rows) {
-        cli_abort(
-          "The slide computations must either (a) output a single element/row each, or
-          (b) one element/row per appearance of the reference time value in the local window."
-        )
-      }
+      return("other")
     }
+  }) %>% unique()
```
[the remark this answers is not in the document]
dshemetov: sounds good, I'll put perf comparison on the todo list below.

dshemetov:
TODO:

- [ ] perf comparison this branch and base branch (profile to find bottlenecks; likely `.complete_only = TRUE` and possibly other parts of `epi_slide_one_group`)

R/slide.R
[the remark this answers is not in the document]
dshemetov: thoughts on this particular line: my thinking is that uniqueness is O(n) in the length of slide_values_list. if we don't pay that cost here, we will have to pay it below when checking `"other" %in% return_types` and then later when checking `length(return_types) != 1L`. i suppose the alternative is to do a single pass and abort as soon as we encounter "other" or a different return_type, but I'm concerned that that for loop would take place in R land, rather than R handling `unique()` in C (i'm guessing). furthermore, if `unique` does take place in C land, it might make sense to output integers here rather than strings (enum optimization), but this is getting into the weeds.

R/slide.R
```diff
@@ -208,129 +206,198 @@ epi_slide <- function(
...
+      .data_group, .group_key, ...,
+      f_factory = f_wrapper_factory,
+      before = window_args$before,
+      after = window_args$after,
+      ref_time_values = .ref_time_values,
+      all_rows = .all_rows,
+      new_col_name = .new_col_name,
+      used_data_masking = used_data_masking,
+      time_type = time_type,
+      complete_only = .complete_only
```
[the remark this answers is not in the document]
dshemetov: not a big deal, I'm happy to rename these. I wasn't happy with mixing all these dotted names and undotted names, it just sorta happened.

R/slide.R
```diff
@@ -186,20 +153,63 @@ epi_slide <- function(
...
 
+  align <- rlang::arg_match(.align)
+  time_type <- attr(.x, "metadata")$time_type
+  if (is.null(.window_size)) {
+    if (time_type == "week") {
+      .window_size <- as.difftime(1, units = "weeks")
+    } else {
+      .window_size <- 1
+    }
+  }
```
brookslogan, line 172: issue: this default doesn't make sense, and we had problems with a default of 7 in the past.

I think issues with 7 were: we're not always using daily data, and, when we were using `n` instead of `.n`, there were some weird messages that could be produced when someone wanted to count things.

R/slide.R
```diff
@@ -208,129 +206,198 @@ epi_slide <- function(
...
+  if (identical(window_size, Inf)) {
+    if (align == "right") {
+      before <- Inf
+      if (time_type %in% c("day", "week")) {
+        after <- as.difftime(0, units = glue::glue("{time_type}s"))
       } else {
-        # vector or packed data.frame-type column:
-        mutate(.data_group, !!new_col_name := slide_values)
+        after <- 0
       }
```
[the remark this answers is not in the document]
dshemetov: til you can leave entries blank in `switch`. looks good to me 👍

brookslogan:
I think we can potentially avoid the size checks on elements of `slide_values_list` by using `slider::hop_index_vec` instead of `slider::hop_index`; `slider::hop_index_vec` expects size-1 vectors in the vctrs sense, and will combine them for us (hopefully more efficiently than we can with a `list_unchop` after the fact).

R/slide.R
```diff
@@ -209,128 +219,213 @@ epi_slide <- function(
...
+    )
+  }
+  # Returned values must all be the same type.
+  if (length(return_types) != 1L) {
+    cli_abort(
+      "epi_slide: slide computations must always return either a data.frame or a vector (as determined by the
+      vctrs package), but not a mix of the two.",
+      class = "epiprocess__invalid_slide_comp_value"
+    )
+  }
```
brookslogan, line 356: I think we can potentially avoid the size checks on elements of `slide_values_list` + the `list_unchop` by using `slider::hop_index_vec` instead of `slider::hop_index`; `slider::hop_index_vec` expects size-1 vectors in the vctrs sense, and will combine them for us (hopefully more efficiently than we can with a `list_unchop` after the fact).

brookslogan: I think I may just push a commit for this because it may simplify #509 implementation.

brookslogan: Reconsidering here.  I guess `hop_index_vec` is basically doing `hop_index` +  unnaming + unchopping; no fancy efficiency stuff.  We can get better validation by name-checking + unchopping, or name-checking + other-checks + unchopping, or name-checking + unchopping&wrapping-error-messages-from-unchopping.

R/epi_df.R
dshemetov: still unclear how to resolve this.

R/slide.R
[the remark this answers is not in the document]
dshemetov: update: given recent conversations, going to remove .complete_only and go with the work Nat has done in #415 that autocompletes the epi_df.

R/epi_df.R
brookslogan: Is this still an issue?  Do you have a reprex?

dshemetov: I don't, but the gist is that if you group_by an epi_df inside dplyr_reconstruct, you end up in an infinite recursion.

R/slide.R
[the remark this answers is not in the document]
brookslogan: `map_chr` is already a for loop calling R function inside.  So this actually seems like a potentially promising approach to me, if benchmarking shows performance of this part actually matters.

Previously we had <check if all are data frames> || <check if all are vectors> or something like that, no intermediate map_chr.  Seems like it might involve fewer ops on the inside and have more opportunity for an early exit vs. an intermediate map_chr.  So going back to that might also be a competitor if perf matters.

R/epi_df.R
brookslogan: I don't understand how that happens.  Aren't we working off of a tibble input & haven't converted it yet?

brookslogan: Is it failing in the current state of the branch?  Or do you have some local work on `dplyr_reconstruct` that's triggering the error?

I need to re-read `dplyr_extending` and see if we actually need to touch `dplyr_reconstruct`.  We may only need it on `dplyr_row_slice` and maybe `[`.

brookslogan: Caching in some stuff from memory:
- We probably don't want to do a full key validity check in dplyr_reconstruct; we run that all the time and it could be expensive.
- We might be able to get by with:
  - In `dplyr_row_slice.epi_df`: check that they have not requested the same row index twice.
  - In `[` when it selects columns: if any of the other_keys columns has been dropped, we need to do a full uniqueness check (or some optimization of it).  If it's also selecting rows... not sure if we also need that row slice check here or if row slice is already being called somehow.
  - ?dplyr_extending seems to suggest we don't need a `dplyr_col_modify` implementation for this.

dshemetov: Hm, I had some local changes but I backed them out after the recursion issues. I'm going to rebase this and probably merge and we can iterate on it in the joint PR?

R/slide.R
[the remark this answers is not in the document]
dshemetov: But early exits only happen on error, in which case performance isn't really a concern. Like in the happy case, you're scanning the full list and concluding there were no issues. 

I do see your point that map_chr followed by unique is two passes through the array. We'll revisit when we profile the joint PR.

PR #528, opened 2024-09-26
Time utils

dsweber2:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

This adds a couple of functions that were enough of a headache that having them as utils with tests would be nice. Some of these are adapted from work that Richard did. These are:
- `daily_to_weekly` an archive to archive transformation that converts daily data to weekly data, with the `time_value` being a single day of the week.
- `convert_to_period_upsample` goes from e.g. weekly data to daily data, either splitting data equally between days (so dividing by 7 in the case of weekly-> daily) or not (useful for e.g. % of cases).

Then there's a family of epiweek/season utilities. Hopefully those are all clear from the names.

@lcbrooks @dshemetov @nmdefries before I do things like add tests and docs, I figured getting input on if/how these should change.

Things to do:
- [ ] add tests
- [ ] add docs


brookslogan:
Dmitry added a basic function `sum_groups_epi_df` for epi_dfs in #477 that overlaps in functionality, though I'm not sure if either would be completely subsumed.  I think maybe his could have theoretically allowed coarsening of epikeys, which presents more interface issues (if we have both count and rate&percent cols, we don't want same aggregation method, so we'd need something like count_cols and rate_cols selectors rather than agg_cols + agg_method.  Plus we'd want pop data for aggregating rate&percent cols).

before = 99999 isn't needed anymore for lg window, Inf is default

wday() in lubridate does not follow the same numbering scheme as base R as.POSIXlt()$wday!  There might be bugs here due to this.  In writing week-related code, we should always be using naming conventions or types to clarify which definition we're using.  (For reference, I had some scheme like wday0, wday1, wday7 in brookslogan/epicalendar though maybe I just imagined adding the comments about which of the various time offering "wday"s these correspond to.  But I think maybe there were like 3 different conventions in common packages.)

`convert_to_period_upsample` seems suspiciously general... Allowing selection of `groups` at same time as periods seems like it might run into the same problematic edge cases as epi_slide did (& in epi_slide we're moving to just forbid things that can potentially allow those edge cases).  We should probably be more rigid, perhaps not allowing `groups` parm, or ensuring that it's also upsampling epikeys?

optimization for daily -> weekly: two options:
- we could use `epi_slide_opt` within the `epix_slide`, and also set `before` according to something like `max(version - time_value) + 6L`
- we can probably do even better directly using `data.table` ops, especially if there is significant variance in reporting latency from version to version

brookslogan:
We're also missing 7dav-ing of archives, but that has more interface concerns like new col names.  But maybe we should think a little about that so that if/when we add that the interfaces mesh well.

dsweber2:
Glad to see that `sum_groups_epi_df` exists! The two should have fundamentally different approaches to rate and % data: aggregating by time, the first is a simple sum and the second is a mean, while aggregating by `geo_value` or some other group we need the associated weights (e.g. population). Given this, they should probably be 2 separate functions or family of functions.

> before = 99999 isn't needed anymore for lg window, Inf is default

This didn't work on dev w/out it, but given that this will probably be merging after the slide stuff I'll keep that in mind.

> wday() in lubridate

if I understand correctly, by default Sunday = 1, which I *think* is the only choice needed to specify the day of the week convention uniquely. Week number is a whole different can of worms where I'm still confused if lubridate's epiweek actually matches MMWRweek matches the more maintained [epiweek](https://pypi.org/project/epiweeks/) package. 

> we should always be using naming conventions or types to clarify which definition we're using.

This is tricky, because cross language that's actually not the same; I ended up helping someone debug some results from scoringutils that came down to the fact that their system was using Hebrew instead of English. I suspect prototype days are probably the best way to go about this; "2020-09-27" is unambiguously Sunday, for example.

> `convert_to_period_upsample`  seems suspiciously general

yeah its scope should probably be narrowed. I developed it on the weekly->daily case; the longer the gaps though, the more questionable it gets as a method. Though I think you may be misinterpreting the `groups` parameter; it's just intended to make sure values aren't leaking across group boundaries. It should probably be inferred from the `epi_archive`. I wasn't planning on getting into imputing values cross-groups with this, fill is way too crude for that.

> maybe bug: `fill`

I'll make sure to add a test for this; iirc because `completed` was grouped, after the join it was still grouped, but that could easily be wrong.


> grouped summarize...

I'm not sure how this would work, because for each time value at a given ref_time_value, we need to return a value?  Summarize would only return one value per group; are you suggesting adding a week indicator first, then grouping by that? I'm using `epi_slide_sum` or `epi_slide_mean`, both of which are actually using `epi_slide_opt`, so I suspect that's not actually the expensive part. I think the reason this ends up slow is that it generates ~5m rows, but can compress it back down to 300k. If we had an archive to archive transform that compressed as we went that *might* speed this up, but making all that memory is expensive. 

I considered uncompressing the epi_archive and then doing the computation, but did this version because that was ~5m rows; I didn't know at the time that this method was also going to generate a lot of rows anyway. It would probably be faster to generate once rather than repeatedly append.

> we can probably do even better directly using data.table ops

Not sure I follow really. Do you mean e.g. frollmean and friends?

brookslogan:
> if I understand correctly, by default Sunday = 1, which I think is the only choice needed to specify the day of the week convention uniquely.

- `lubridate::wday()` w/o `week_start =`: who knows.  Depends on user's `lubridate.week.start` option.  Default default is indeed Sunday = 1.  If we use this definition we should think carefully about implications of the option setting.
- `as.POSIXlt()$wday`, `format(, "%w")`: Sunday = 0 / "0"
- `format(, "%u")`: Sunday = "7"

> Week number is a whole different can of worms where I'm still confused if lubridate's epiweek actually matches MMWRweek matches the more maintained [epiweek](https://pypi.org/project/epiweeks/) package.

```r
dates <- seq(as.Date("1900-01-01"), as.Date("2500-01-01"), by = "day"); all.equal(MMWRweek::MMWRweek(dates)$MMWRweek, lubridate::epiweek(dates))
[1] TRUE
```
oh, apparently there's also a `lubridate::epiyear()` now!

> This is tricky, because cross language that's actually not the same

Agreed, I'm not suggesting using "Sunday".  But rather, instead of `wday` we should use something that indicates what numbering/whatever scheme we're using.  Using something other than `character` and checking for that type could be a cleaner way, though I haven't seen that ever done.

> Though I think you may be misinterpreting the groups parameter; it's just intended to make sure values aren't leaking across group boundaries. It should probably be inferred from the epi_archive

Inferring sounds good.

>  I'm using epi_slide_sum or epi_slide_mean

I didn't see this, sorry.

> Summarize would only return one value per group; are you suggesting adding a week indicator first, then grouping by that?

Yeah, week column, group by epikey x week column, summarize.

> I think the reason this ends up slow is that it generates ~5m rows, but can compress it back down to 300k.

`before = <right non-huge value>` might help here (though before = non-Inf has overhead not in before = Inf), or:

> > we can probably do even better directly using data.table ops

> Not sure I follow really. Do you mean e.g. frollmean and friends?

Probably, but the main thing is to try to avoid avoiding epix_slide overhead, and to adapt to the version / version-epikey-specific latency.  Example of the latter: if there is a 1000d-latency day but most are 5d latency then using the before = <right non-huge value> trick is still going to have a bunch of extra munging done (before = 1006 or whatever even though before = 11 would almost work).  (It could also avoid redundant calculations on some no-diff versions in certain edge cases with epix_slide_default_ref_time_values.)

might look something like
```r
ranges <- DT[,list(mintv = min(time_value), maxtv = max(time_value)), by = setdiff(key(DT), "time_value")]
requests <- ranges %>% rowwise() %>% reframe(geo_value, time = seq(mintv - 6L, maxtv, by = "day"), version) %>% setDT()
# ^ or something fully data.table-based, and maybe non-`Date`-based.
# ^ note this probably breaks if you have dtplyr loaded, we would need to do a better way.  With dtplyr unloaded, it's temporarily breaking DT ownership expectations and keeping around some attributes which MIGHT make `setDT` acknowledge it's still sorted according to the same key as the origiinal & set that as the key without checking/re-sorting. 
DT[requests][, .... frollsum/mean(.....), by = .......]
```

R/time_converters.R
```diff
@@ -0,0 +1,127 @@
...
+#'   Represents the last day, so the week consists of data summed to this day.
+#'   The default value `6` means that the week is summed from Sunday through
+#'   Saturday.
+daily_to_weekly <- function(epi_arch,
+                            agg_columns,
+                            agg_method = c("sum", "mean"),
+                            day_of_week = 4L,
+                            day_of_week_end = 6L) {
+  agg_method <- arg_match(agg_method)
+  if (agg_method == "total") {
```
dshemetov, line 19:
```suggestion
  if (agg_method == "sum") {
```

R/time_converters.R
```diff
@@ -0,0 +1,127 @@
...
+    agg_fun <- epi_slide_sum
+  } else if (agg_method == "mean") {
+    agg_fun <- epi_slide_mean
+  }
+  keys <- grep("time_value", key_colnames(epi_arch), invert = TRUE, value = TRUE)
+  too_many_tibbles <- epix_slide(
+    epi_arch,
+    before = 99999999L,
```
dshemetov, line 27: Inf is default
```suggestion
```

R/time_converters.R
```diff
@@ -0,0 +1,127 @@
...
+    ref_time_values = ref_time_values,
+    function(x, group, ref_time) {
+      x %>%
+        group_by(across(all_of(keys))) %>%
```
dshemetov, line 31: after new updates you can just do and get rid of the keys assignment above
```suggestion
        group_epi_df(exclude = "time_value") %>%
```

R/time_converters.R
```diff
@@ -0,0 +1,127 @@
...
+    to_complete_datatable %>%
+    group_by(across(all_of(groups))) %>%
+    reframe(
+      time_value = seq(from = time_value, to = time_value + source_period - 1, by = target_period)
+    ) %>%
+    unique()
+  completed <- to_complete_datatable %>%
+    full_join(
+      completed_time_values,
+      by = join_by(season, geo_value, version, time_value)
```
dshemetov, line 71: `season` seems particular to whatever you were working on. this probably needs to relate somehow to `groups` above.

R/time_converters.R
```diff
@@ -0,0 +1,127 @@
...
+  }
+  keys <- grep("time_value", key_colnames(epi_arch), invert = TRUE, value = TRUE)
+  too_many_tibbles <- epix_slide(
+    epi_arch,
+    before = 99999999L,
+    ref_time_values = ref_time_values,
+    function(x, group, ref_time) {
+      x %>%
+        group_by(across(all_of(keys))) %>%
+        agg_fun(agg_columns, before = 6L) %>%
```
dshemetov, line 32: agg_fun doesn't exist yet? i'd recommend looking at `sum_groups_epi_df` in epiprocess 0.9.0.

dshemetov (commented):
overall seems sensible and handy to have, but we should think a bit about design - like Logan says, we can either extend sum_groups_epi_df or make a complement to it (expand_groups_epi_df?). i suggest we do a use case analysis here: prototype example workflows we want and see if we can make a common interface.

R/time_converters.R
brookslogan: (this should also give some perf boost)

dshemetov:
> Glad to see that sum_groups_epi_df exists! The two should have fundamentally different approaches to rate and % data: aggregating by time, the first is a simple sum and the second is a mean, while aggregating by geo_value or some other group we need the associated weights (e.g. population). Given this, they should probably be 2 separate functions or family of functions.

Time resolution upsampling and downsampling definitely felt like a separate feature, so we punted on it in #519. 

Mostly what we're doing in `sum_groups_epi_df` is allowing users to contract data along a grouping key index. The intention there was to give an easy workaround to someone who used to use ungrouped epi_slides to aggregate to national at the same time as sliding - now they call sum_groups_epi_df and then slide. 

Logan's comments about disallowing user-supplied groups seem wise. I would suggest we first try to group epi_dfs by default by `key_colnames(epi_df, exclude = "time_value")` (this should be geo_value + other_keys) and handling all upsampling/downsampling there. Since our restriction on time_types, we might want to have the user explicitly mark their "from time_type" and "to time_type" (starting to feel like geo_coding) to avoid the user up and downsampling to types that we don't understand how to handle. One of your functions required a "period" length and it's set to 7 to weekly, but going up to month or yearly, this approach won't work, so we'll need special code paths for those time_types if we plan to support them.

brookslogan:
As I thought about optimizations, I'm also realizing that our logic here depends on whether user is using Sun--Sat weeks, Mon--Sun weeks, or something more esoteric (more esoteric things have been used, I believe, in some source mirrored in covidcast).  We could start off with Sun--Sat for US epiweeks but might want this to be a parameter.  Also there's the question of what day of the week should represent the week in the output, or if we add support for tsibble::yearweek to represent; `day_of_week_end` sounds like a decent default at least given how CDC/Hubverse stuff looks.

dshemetov:
As for concerns about different conventions for wday(), it's good to be aware of these issues, but as long as we describe our convention in the docs and are consistent with whatever date-time library we use (whether as.POSIX or lubridate or other), I don't see what else there is to do.

> idea[non-blocking]: As I thought about optimizations, I'm also realizing that our logic here depends on whether user is using Sun--Sat weeks, Mon--Sun weeks, or something more esoteric (more esoteric things have been used, I believe, in some source mirrored in covidcast). We could start off with Sun--Sat for US epiweeks but might want this to be a parameter. Also there's the question of what day of the week should represent the week in the output, or if we add support for tsibble::yearweek to represent; day_of_week_end sounds like a decent default at least given how CDC/Hubverse stuff looks.

The fully general solution here would require a general week format library - `epi_week_sequence(start, end, wday_start, at_least_n_days_in_week)` - and I don't think we're prepared to write or support that. Given that the general solution is infeasbile, we should make some common choices and just support those until we get another request. Probably the CDC epiweek and just a basic "sum to every Monday" (without the epiweek rounding) seem good enough to me.

brookslogan:
I realize David already has all the options I was talking about but forgot about while writing data.tableish appr: `day_of_week` and `day_of_week_end`.  (For week numbering you may also need `owning_wday` if you don't take it to be `day_of_week`.)  Guess that could be cut out if it's too complicated.

R/time_converters.R
```diff
@@ -0,0 +1,127 @@
...
+        select(-all_of(agg_columns)) %>%
+        rename_with(~ gsub("slide_value_", "", .x)) %>%
+        # only keep 1/week
+        filter(wday(time_value) == day_of_week_end) %>%
```
brookslogan, line 36:
```suggestion
        filter(as.POSIXlt(time_value)$wday == day_of_week_end) %>%
```
If you want to ensure 0 = Sun, 6 = Sat.  Could maybe call this `lt_day_of_week_end`.

OR

```suggestion
        filter(wday(time_value, week_start = 1) == day_of_week_end) %>%
```
If you want to ensure 1 = Mon, 6 = Sat, 7 = Sun.

OR

```suggestion
        filter(wday(time_value, week_start = 7) == day_of_week_end) %>%
```
If you want to ensure 1 = Sun, 7 = Sat.

R/time_converters.R
dsweber2: glad to hear it! I was getting very tired of recreating `keys` all over the place

R/time_converters.R
dsweber2: yeah, should probably just be `keys`. Relatedly, we probably need a function to create `keys`

R/time_converters.R
dsweber2: it's defined above? If there's a general function that wasn't what I was invoking

R/time_converters.R
dsweber2: Based on you pointing out the default is customizable, I was just planning on doing the 3rd one. Most of our collaborators/expected users are in the US, so defaulting to Sunday being the first day of the week is probably safest.

R/time_converters.R
dshemetov: there is epiprocess::key_colnames. unless you mean something like epiprocess::add_key which would probably require a join and recreating the epi_df under the hood.

R/time_converters.R
dshemetov: I swear i CTRL + F for it and only found this

R/time_converters.R
dsweber2: `key_colnames` includes `time_value`, unless there's an option I missed. So every time I have to remove `time_value`; Dan even added a `kill_time_value` function in `epipredict` because it's such a recurring pattern.

R/time_converters.R
```diff
@@ -0,0 +1,127 @@
...
+        # switch time_value to the designated day of the week
+        mutate(time_value = time_value - 7L + day_of_week) %>%
```
brookslogan, line 38: issue?: this might need to be
```suggestion
        mutate(time_value = time_value - 7L + (day_of_week - day_of_week_end) %% 7L) %>%
```
Probably easiest to construct a test case with `day_of_week_end` not Saturday & see if anything gives the right result.

brookslogan: Did a little test, seems to back up current suggestion above.
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(tidyr)
crossing(end_date = as.Date("2020-01-01") + 0:6, day_of_week = 1:7) %>%
  mutate(day_of_week_end = lubridate::wday(end_date, week_start = 7L)) %>%
  mutate(
    method1 = end_date - 7L + day_of_week,
    method2 = end_date - 7L + (day_of_week - day_of_week_end) %% 7L,
    method3 = end_date - (day_of_week_end - day_of_week) %% 7L
  ) %>%
  pivot_longer(method1:method3, names_to = "method", values_to = "date") %>%
  mutate(day_of_week_actual = lubridate::wday(date, week_start = 7L),
         right_day_of_week = day_of_week == day_of_week_actual,
         right_week = as.integer(end_date - date) %in% 0:6) %>%
  summarize(.by = "method", issues = sum(!right_day_of_week | !right_week))
#> # A tibble: 3 × 2
#>   method  issues
#>   <chr>    <int>
#> 1 method1     42
#> 2 method2      7
#> 3 method3      0
```

<sup>Created on 2024-09-28 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

dsweber2: yeah this definite seems right, thanks for the catch

brookslogan:
Here's the fast-geo-aggregation code I was referring to.  This might be useful on the road to a `sum_groups_epi_df` analogue, though this code is doing means rather than sums.  To make it applicable to time aggregation, we would need to first generate an old-time-to-new-time crosswalk; e.g., for daily-to-weekly, a tibble with 2 or 3 columns: `old_time_value`, `new_time_value`, and maybe a `weight` column.  For each `new_time_value` (week), there should be 7 rows (1 per constituent day); the `old_time_value`s should be a superset of `unique(archive$DT$time_value)`.  The weights would all be 1.

```r
# XXX perf: the following vs. something with epiprocess:::tbl_patch sequence within each new epikey?

# XXX perf: try to prevent actual joins on character vectors?  convert to integers?  chop?

# Geo-interaction data:
contribs =
  county_archive$DT |>
  as.data.frame() |>
  as_tibble() |>
  inner_join(
    by = c("geo_value" = "county22_fips"),
    county22_phrru22_pop20_crosswalk |>
      select(county22_fips, phrru22_code, county22_pop20_prop_in_phrru22),
    # LHS: counties can have many time_value x versions; RHS: each county only
    # appears once (each county is only part of one PHRRU and has only one
    # population measurement; these pop20 proportions should all be 1 but the
    # weighting is included to try to build towards generality)
    relationship = "many-to-one",
    unmatched = c("error", "drop") # lookup LHS in RHS
  ) |>
  # mutate(signal_contrib = signal * county22_pop20_prop_in_phrru22,
  #        .keep = "unused") |>
  identity()

# For each new epikeytime, which versions update it?
phrru22_update_ukeys =
  contribs |>
  distinct(phrru22_code, time_value, version)

# To calculate the update for new-epikeytime n for version v, we need to look up
# all old-epikeytime-new-epikeytime-interaction data *as of* version v; this is
# a rolling join rather than an ordinary join; e.g., consider county -> state
# when only one county in the state had an update/diff.
contrib_requests =
  phrru22_update_ukeys |>
  inner_join(
    by = "phrru22_code",
    county22_phrru22_pop20_crosswalk |>
      select(county22_fips, phrru22_code, county22_pop20_prop_in_phrru22),
    relationship = "many-to-many",
    unmatched = c("error", "drop")
  ) %>%
  select(phrru22_code, geo_value = county22_fips, time_value, version)

phrru22_archive =
  contribs %>%
  as.data.table(key = c("phrru22_code", "geo_value", "time_value", "version")) %>%
  `[`(contrib_requests, on = c("phrru22_code", "geo_value", "time_value", "version"), roll = TRUE) %>%
  # XXX NAs vs. 0s... may want an option to exclude things based on a ".real"
  # indicator .... but if summing need to renormalize; need to keep weights around
  as.data.frame() %>%
  as_tibble() %>%
  summarize(.by = c("phrru22_code", "time_value", "version"),
            signal = weighted.mean(signal, county22_pop20_prop_in_phrru22)) %>%
  rename(geo_value = phrru22_code) %>%
  as_epi_archive(
    other_keys = character(),
    clobberable_versions_start = county_archive$clobberable_versions_start,
    versions_end = county_archive$versions_end
  ) %>%
  {}
```

PR #532, opened 2024-10-01
Fix validate_version_bound typo

brookslogan:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
- [x] Make sure to bump the version number in `DESCRIPTION`. Always increment
      number (the second number).
- [-] Describe changes made in NEWS.md, making sure breaking changes
  process.

This fixes a typo that appears to have no impact, but might confuse us in the future.  The impacted `x_arg` never actually seems to be used, so maybe something else should be done but I'm not looking too much into it.

- Resolves #{issue number}

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: epiprocess
 Title: Tools for basic signal processing in epidemiology
-Version: 0.9.0
+Version: 0.10.0
 Authors@R: c(
     person("Jacob", "Bien", role = "ctb"),
     person("Logan", "Brooks", , "lcbrooks@andrew.cmu.edu", role = c("aut", "cre")),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: epiprocess
 Title: Tools for basic signal processing in epidemiology
-Version: 0.10.0
+Version: 0.9.1
 Authors@R: c(
     person("Jacob", "Bien", role = "ctb"),
     person("Logan", "Brooks", , "lcbrooks@andrew.cmu.edu", role = c("aut", "cre")),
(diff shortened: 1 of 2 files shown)


DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Type: Package
 Package: epiprocess
 Title: Tools for basic signal processing in epidemiology
-Version: 0.9.0
+Version: 0.10.0
```
dshemetov, line 4: Increment patch number of a patch?
```suggestion
Version: 0.9.1
```

R/archive.R
```diff
@@ -25,7 +25,7 @@
 #' @noRd
 validate_version_bound <- function(version_bound, x, na_ok = FALSE,
                                    version_bound_arg = rlang::caller_arg(version_bound),
-                                   x_arg = rlang::caller_arg(version_bound)) {
+                                   x_arg = rlang::caller_arg(x)) {
```
dshemetov, line 28: This argument appears unused in this function?

dshemetov (approved):
Good catch, but a few minor comments. Approving so you can merge after whatever fixes you decide on.

R/archive.R
brookslogan: Right, I'm not sure why; haven't looked enough at function to know if it should be.  But that's also why I felt confident enough to just make this typo fix in case we do use/read it in the future.

brookslogan: If you know whether it should be used maybe we could replace this PR with an actual fix.

DESCRIPTION
brookslogan: I'm getting mixed up looking at changelog.  Are we building up changes for 0.10 release under 0.9.x?

dshemetov: Think of the 0.9 heading like an append log of all the changes in the 0.9.x versions. So we should put the changes from here there.

R/archive.R
dshemetov: I don't know, let's just go ahead with the small fix for now.

DESCRIPTION
brookslogan: Still confused here.  Going to follow the 0.7.y and 0.8.y pattern of 0.x.y, y > 0, of adding NEWS.md entries for 0.{x+1}, since we can fix up easily later.

PR #533, opened 2024-10-01
feat(epi[x]_slide): hint on forgotten syntax specifying comp

### LINKED ISSUE #529: Force `.f` argument usage in epix_slide
- [ ] issue?: unclear error message here
```
archive_cases_dv_subset %>% group_by(geo_value) %>% epix_slide(.before = 6, tibble(slide_value = mean(.x$percent_cli)))
Error: object '.x' not found
```
We should probably capture + generate a better error message by forcing `.f` at the beginning; a fix is to insert an extra comma so this is sent to `...` rather than `.f` and can be considered for tidyeval unnamed tibble column unpacking.  Seems like probably an issue that already existed.  It'd be nice to just let this expression through in `.f`, but I think I already have a dedicated separate issue for this and it is not as simple as it sounds. 

_Originally posted by @brookslogan in https://github.com/cmu-delphi/epiprocess/issues/477#issuecomment-2328442625_

brookslogan:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

Tries to provide some hints when a user has forgotten part of the syntax when specifying their slide computation, e.g., the tilde in a formula.  It's just a hint so we don't false-diagnose e.g.
```r
epix_slide(archive, my_failing_forecaster_function_factory(my_forecaster_args))
```
Perfectly distinguishing between forgotten syntax and the above type of error may allow us to just fix it up, but likely requires significantly more complex slide code, so we could try these hints first.

- Resolves #529.


DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Type: Package
 Package: epiprocess
 Title: Tools for basic signal processing in epidemiology
-Version: 0.9.0
+Version: 0.10.1
```
dshemetov, line 4: I think we should be in the 0.9.x territory, since we just incremented to 0.9.0 from 0.8.x in the slide improvements.
```suggestion
Version: 0.9.1
```

dshemetov (approved):
Looks great to me, thanks! Didn't test it, since the tests look good enough to me 👍

DESCRIPTION
brookslogan: Should changelog notes still be under 0.10?

dshemetov: Hm, my thinking was we append to 0.x whenever we bump 0.x.n to 0.x.n+1 and when we make a transition to 0.x+1, we write the changes that went along with that change under 0.x+1. So here we'd just be appending to the list of changes in 0.9.

brookslogan: I commented on other PR about this but I'm just going to temporarily match 0.7.y and 0.8.y, y>0 behavior here of adding changes under next 0.x heading.

brookslogan:
While addressing some lints & testing annoyances, I realized existing error messages may not hint enough for someone who is expecting `epi_slide_opt` interface.  I tweaked things a bit and added a bit of a lazy check for whether they are trying to use `.col_names` (only detecting whether they pass it by name).

@dshemetov If you have time, could you please take a look?  Think it's ready to merge (DESCRIPTION, NEWS.md etc. up to date) if you think it looks good.

dshemetov (approved):
For some reason seeing diffs from #535  and #532 here, I thought those were already in dev. Strange. Oh well, this looks good to me!

PR #566, opened 2024-11-12
update workflow, no more special color handling

dajmcdon:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

- Resolves #{issue number}

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -42,7 +42,22 @@ jobs:
           needs: website
 
       - name: Build site
-        run: pkgdown::build_site_github_pages(new_process = FALSE, install = FALSE)
+        # our versioning system+dev branch doesn't match the requirements for
+        # develop mode = auto
+        run: |
+          target_ref <- "${{ github.event_name == 'pull_request' && github.base_ref || github.ref }}"
+          override <- if (target_ref == "main" || target_ref == "refs/heads/main") {
+            list(development = list(mode = "release"))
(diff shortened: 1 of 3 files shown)


.github/workflows/pkgdown.yaml
```diff
@@ -42,32 +42,12 @@ jobs:
...
-            list(development = list(mode = "devel", version_label = "success"))
-          } else {
-            stop("Unexpected target_ref: ", target_ref)
-          }
-          pkg <- pkgdown::as_pkgdown(".", override = override)
-          cli::cli_rule("Cleaning files from old site...")
-          pkgdown::clean_site(pkg)
-          pkgdown::build_site(pkg, preview = FALSE, install = FALSE, new_process = FALSE)
-          pkgdown:::build_github_pages(pkg)
+        run: pkgdown::build_site_github_pages(new_process = FALSE, install = FALSE)
```
dshemetov, line 45: So all the previous code handled made custom dev docs for the `dev` branch, rather than basing it on `x.y.z.9000` version numbers. I honestly would prefer to move to the latter, default model, but we'll need to adjust our workflows a bit.

dajmcdon: I'll bring this back. I don't think there's a way to change with the `dev` branch unfortunately. But hopefully we don't need the color handling any more.

brookslogan:
I let this sit too long and am going to just release dev -> main so we should get the updates from that.

PR #577, opened 2024-12-06
Autoload `epidatasets` as a dependency rather than reexporting epidatasets data

nmdefries:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

- Resolves #{issue number}


brookslogan:
Could you give an overview of the pros/cons of this change?  I think I ended up adjusting things trying to make checks run over in https://github.com/cmu-delphi/epiprocess/pull/564, but I remember feeling forced into some approach to satisfy both R CMD CHECK and pkgdown checks simultaneously, and don't remember how that approach ended up "feeling".  Does CRAN suggest against `Depends:`?

nmdefries:
Pros for this change:

- matches the [approach used in `epipredict`](https://github.com/cmu-delphi/epipredict/pull/382)
- from experience with `epipredict`, speeds up `document`ing
- don't need to re`document` this package whenever `epidatasets` changes

Cons:

- Adds a `Depends:`
- "if your package relies on a package A which itself "Depends" on another package B, your package will likely need to attach A with a "Depends" directive. This is because the functions in package A were written with the expectation that package B and its functions would be attached to the search() path" [source](https://stackoverflow.com/a/8638902) ("likely"?). This is old and not mentioned in the [R Packages book](https://r-pkgs.org/dependencies-mindset-background.html#sec-dependencies-imports-vs-depends), so unclear if this is a theoretical or perhaps resolved issue. I haven't found additional information yet.

Neutrals:

- `Depends` are considered less safe than `Imports` because of possible name collisions in the search path. This is very unlikely to impact us; the data objects we are making available have fairly long and specific names.

No specific guidelines about Depends vs Imports in the [CRAN policies](https://cran.r-project.org/web/packages/policies.html).

brookslogan:
Thanks for recording this!

Notes:
- The approach in #564 also allows for not having to re-document() when epidatasets is updated, unless object names change etc.  IIRC it did have a sort of useless/annoying re-export doc page though.

> "if your package relies on a package A which itself "Depends" on another package B, your package will likely need to attach A with a "Depends" directive. This is because the functions in package A were written with the expectation that package B and its functions would be attached to the search() path" [source](https://stackoverflow.com/a/8638902) ("likely"?). This is old and not mentioned in the [R Packages book](https://r-pkgs.org/dependencies-mindset-background.html#sec-dependencies-imports-vs-depends), so unclear if this is a theoretical or perhaps resolved issue. I haven't found additional information yet.

- I'd be very surprised if this is the case.  They say that '''An "Imports" directive will load but not attach package A and will neither load nor attach package B.''', but the whole thing with Depends: is that simply loading (not attaching) a package with attach its Depends: packages; I don't know why the behavior would be changed simply because it was transitively loaded via Imports: of another package.

> Depends are considered less safe than Imports because of possible name collisions in the search path. This is very unlikely to impact us; the data objects we are making available have fairly long and specific names.

Right, I doubt we are going to have name collisions.  Just maybe some annoying junk in completion options if only loading epiprocess and not attaching.  Seems rare enough.

brookslogan (approved):
- [x] todo: resolve merge conflicts.  Probably at least version number and where the NEWS entry goes because we did a 0.10.0 release for the workshop.

PR #578, opened 2024-12-07
Rn/na-notebook

rnayebi21:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

Added the reviewed na-notebook. Key differences from last time: added an example and explained `complete()`, as well as added the LOCF in version example.

- Resolves #{issue number}


vignettes/na-notebook.Rmd
```diff
@@ -0,0 +1,510 @@
...
+
+To deal with these smaller NA gaps we will impute in time (as opposed to in version). Two of the most common and more simpler imputation schemes are the last observation carried forward (LOCF) method and a moving average imputation method. As suggested in the name of the method, to do LOCF imputation you simply use the previous available value for the missing value. Here is the corresponding code for LOCF.
+
+```{r locf_fn, warning=FALSE}
+impute_locf <- function(data) {
+  data %>%
+    group_by(geo_value) %>%
+    mutate(across(where(is.numeric), ~ zoo::na.locf(.x, na.rm = FALSE), .names = "{.col}_locf")) %>%
```
brookslogan, line 291: todo: use `fill` from `tidyr` to do this instead, to focus on a smaller set of packages.  I guess we should probably include an `arrange` to make sure things are arranged by `time_value` within each `geo_value` as well.

vignettes/na-notebook.Rmd
```diff
@@ -0,0 +1,510 @@
...
+  library(ggplot2)
+})
+utils::globalVariables(c("is_na", "signal", "issue"))
+```
+
+# How Different NAs Arise and Strategies for Handling Them
+
+### NAs from merging
+
+First let's start with discussing the most common type of missing values that appeared in the context of my auxiliary signal project. When working with multiple signals each signal will likely begin recording at different times. In other words each signal's first data point $t_0$ will differ on the absolute time scale. As a result, when calling `epix_merge()` to combine multiple signals, the signals that started recording at a later point in time will have missing values for the time periods where the other signals were already recording. Here's a quick example
```
brookslogan, line 24: todo: let's present this as more instructional.  Here, that means just tweaking language, getting rid of "that appeared in the context of my auxiliary signal project", and then softening/rewording "the most common" since it may not be the most common in general.

(But in other parts we actually need to make it the content more instructional/relevant (use real data not buggy/artificial).)

brookslogan: issue: `epix_merge()` can introduce NAs both from differing min `time_value`s and differing min `version`s, but
- the description here is a little ambiguous about which it's referring to
- the table is showing something more like an `epi_df`

Possible fix: instead of mentioning `epix_merge()` at this point, we could say this is an issue with some types of joins, and then transition into the `epix_merge()` example with some more discussion.

brookslogan (commented):
Comments so far.  Will revisit.

vignettes/na-notebook.Rmd
```diff
@@ -0,0 +1,510 @@
...
+) %>%
+  select(geo_value, time_value, version = issue, out_visits_smoothed_dr = value) %>%
+  as_epi_archive()
+```
+
+Let's verify that there aren't any NAs in the signals prior to merging. *Note we will be looking into the latest version of the data first*
+
+```{r latest_fn}
+latest <- function(x) {
+  epix_as_of(x, max_version = max(x$versions_end))
```
brookslogan, line 79:
```suggestion
  epix_as_of(x, x$versions_end)
```
we renamed `max_version`, and `versions_end` is a scalar

vignettes/na-notebook.Rmd
```diff
@@ -0,0 +1,510 @@
...
+
+# s05 is Anosmia, Dysgeusia, Ageusia
+s05_smoothed_search <- pub_covidcast(
+  source = "google-symptoms",
+  signals = "s05_smoothed_search",
+  time_type = "day",
+  geo_type = "state",
+  time_values = epirange(20200220, today),
+  geo_values = states,
+  issues = epirange(20201130, today)
```
brookslogan, line 57: suggest:
move to `time_values = "*", issues = epirange(12340101, today)` or some other absurdly early start issue

brookslogan:
Going to make sure we have some other main todos here:
- [ ] @rnayebi21 / etc. to find some alternative to the artificial/buggy example data
- [ ] @brookslogan / etc. to figure out where to put this; vignette vs. book chapter; plus figure out how to store aux data if necessary

PR #591, opened 2025-01-10
Fix age group rate aggregation example

brookslogan:
### Checklist

This is part 1 of a 3-part PR.  (Checks and lints connect it to the other parts.)  A recombined PR will be made to dev.  Some of these process things will be handled there.  I don't want to do more git surgery.  Probably should have just had a couple reviewers for the original PR and pointed to different files.

Please:

- [-] Make sure this PR is against "dev", not "main" (unless this is a release
      PR).
- [-] Request a review from one of the current main reviewers:
      brookslogan, nmdefries.
- [-] Makes sure to bump the version number in `DESCRIPTION`. Always increment
      the patch version number (the third number), unless you are making a
      release PR from dev to main, in which case increment the minor version
      number (the second number).
- [-] Describe changes made in NEWS.md, making sure breaking changes
      (backwards-incompatible changes to the documented interface) are noted.
      Collect the changes under the next release number (e.g. if you are on
      1.7.2, then write your changes under the 1.8 heading).
- See [DEVELOPMENT.md](DEVELOPMENT.md) for more information on the development
  process.

### Change explanations for reviewer

- See #587

### Magic GitHub syntax to mark associated Issue(s) as resolved when this is merged into the default branch

- Resolves #587

PR #592, opened 2025-01-10
Rework `key_colnames`

### LINKED ISSUE #565: `key_colnames` returns wrong values for archive data
See 
```
archive_cases_dv_subset$DT %>% mutate(some_extra_key = geo_value) %>% as_epi_archive(compactify = TRUE, other_keys = "some_extra_key") %>% key_colnames
```
only returns `geo_value` and `time_value`. The result should match `data.table::key`

brookslogan:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

- Make `key_colnames.epi_archive` output its unique key colnames (including version).
- Make `key_colnames.data.frame` require `other_keys` to be passed.
- Remove `key_colnames.default`.
- Require `exclude =` to be passed by name in `key_colnames`.
- Require dots to be empty in `key_colnames`.
  - Note: this triggers some errors in epipredict tests.  See below.

#### `epipredict` errors

Originally, I misread and thought these were from passing `other_keys =` hopefully-redundantly into `key_colnames.epi_df` and having them be ignored.  I was planning to do something like
```r
  expected_other_keys <- attr(x, "metadata")$other_keys
  if (is.null(other_keys)) {
    other_keys <- expected_other_keys
  } else {
    if (!identical(other_keys, expected_other_keys)) {
      cli_abort(c(
        "`other_keys` was provided, but didn't match expectations from inspecting `x`",
        "*" = "`other_keys` was {format_chr_with_quotes(other_keys)}",
        "*" = "`expected_other_keys` was {format_chr_with_quotes(expected_other_keys)}",
        "i" = "If you resolve this discrepancy by adjusting the metadata of `x`, you
               shouldn't have to pass `other_keys =` here anymore unless you want to
               continue to perform this check."
      ))
    }
  }
```
However, the error is actually from passing `extra_keys =` (which was also previously ignored, but it sounds like it has a different meaning).  I am looking a bit more into this.

- [x] todo: decide what to do re. `extra_keys =`
  - idea for now: forbid + PR to epipredict to specify `other_keys =` instead.  If current behavior of `other_keys =` with `epi_df`s breaks later, then decide whether to change things here or adjust any offending steps/layers.
- [x] todo: check whether epipredict#410 is caught... requiring `other_keys =` in `key_colnames.data.frame` should be flagging this, right?
  - It is.  Also need to include fix in epipredict PR before merging this.
- [x] todo: prepare epipredict PR and get it merged

- Resolves #565


brookslogan:
- [ ] after merging this + merging #591: open PR [here](https://github.com/cmu-delphi/epiprocess/compare/dev...lcb/key_colnames-revision_summary-age_agg-updates) to hold the recombined PR with the intended changes and various CHECK+lint+adjacent fixes that were split off into dependent PRs

R/key_colnames.R
```diff
@@ -1,47 +1,123 @@
...
+  setdiff(keys, exclude)
 }
 
 #' @rdname key_colnames
 #' @method key_colnames epi_df
 #' @export
-key_colnames.epi_df <- function(x, exclude = character(0L), ...) {
+key_colnames.epi_df <- function(x, ...,
+                                geo_keys = "geo_value",
+                                other_keys = NULL,
```
dshemetov, line 64: Maybe default to `attr(x, "metadata")$other_keys`?

R/key_colnames.R
```diff
@@ -1,47 +1,123 @@
...
-#' @rdname key_colnames
+#' @importFrom rlang check_dots_empty0
 #' @method key_colnames data.frame
 #' @export
-key_colnames.data.frame <- function(x, other_keys = character(0L), exclude = character(0L), ...) {
+key_colnames.data.frame <- function(x, ...,
+                                    geo_keys = intersect("geo_value", names(x)),
+                                    other_keys,
+                                    time_keys = intersect("time_value", names(x)),
+                                    exclude = character()) {
```
dajmcdon, line 41: On first look, this order seems odd.
1. Dots must be empty, so I feel like the should be last (or potentially before exclude but after the rest).
2. `other_keys` with no default. I tend to prefer args with no default to precede those with defaults. Here, it seems like it should come after `time_keys` based on relative importance, and probably have a default (`NULL` to match the `.epi_df` case?).

dajmcdon (approved):

R/key_colnames.R
brookslogan: It is odd... here's the debate I think I was having:
- We have been moving toward using the order geo_value, !!!other_keys, time_value[, version] more consistently throughout epiprocess and maybe epipredict.
- I was trying to compromise on how demanding to be.  `key_colnames.data.frame`'s main use case is in epipredict.
  - Not providing `other_keys` is very likely an epipredict bug; we should always be feeding it in from roles.
  - Not providing `geo_keys` or `time_keys` is very likely also an epipredict bug; we should be feeding it in from roles.  But I was trying to match some of the defaulty-ness of current epipredict/helpers.

Two potential alternatives:
- Make them all mandatory.  This probably requires epipredict updates to forward geo & time roles, but would be (maybe only partially) fixing bugs.
- Make them all optional; `other_keys` defaulting to `character(0)`.

brookslogan: Hmm, maybe no updates required to do option 1?  Based on running epipredict tests and as much of the examples as devtools would (it errors with "Can't create a layer without a stat"), I get the following counts:
```r
  <chr>                                       <int>
1 epi_df, tbl_df, tbl, data.frame              3009
2 epi_recipe, recipe                           1567
3 epi_workflow, epi_workflow, workflow          364
4 epi_df, grouped_df, tbl_df, tbl, data.frame    20
5 epi_workflow, workflow                          9
6 recipe                                          4
```
I thought I already checked that we were using the data.frame method somewhere in epipredict, but maybe that went away with the earlier updates avoiding key_colnames I made or other updates.

brookslogan: I'm going with approach 1 for now.  This leaves us some flexibility to start using the class-hedging approach if we need to in epipredict. Should be easy enough to dump the extra code and move to approach 4 later if we feel like it's not needed.

R/key_colnames.R
brookslogan: Fixed in #540 which will be merged in here.

PR #607, opened 2025-02-18
Minor performance improvements and enabling fixes

brookslogan:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

These are a handful of minor performance improvements that helped improve an archive -> archive 7dav "epix_epi_slide_opt" that I'm still packaging up, but these seem like they can stand alone.

An overview of changes:
- Avoid unnecessary group "reconstruction" after a lot of grouped epi_df operations.  The root of this is our wonky `?dplyr_extending` implementation which, in a bunch of epi_df dplyr operations, ran `dplyr_reconstruct.epi_df`, which for grouped epi_dfs (i) ran `dplyr_reconstruct.grouped_df`, and (ii) did actual epi_df-related checks&adjustments.  (i) recomputes the grouping, but the basic operations already all forwarded to grouped_df methods that already calculated the grouping (likely more efficiently).
- Avoid functions that use tidyeval (data masking or tidyselect) within tight loops.
- Replace some other operations (duplicate key detection via locf vs. just vec_duplicate_any, Date %in% vs. vctrs alternative, bind_rows vs. vec_rbind when the output classes/ptypes are compatible or can be made compatible) based on profiling on some of our example data.

- Resolves #{issue number}


R/epi_df.R
```diff
@@ -189,7 +189,10 @@ new_epi_df <- function(x = tibble::tibble(geo_value = character(), time_value =
 
   # Reorder columns (geo_value, time_value, ...)
   if (nrow(x) > 0) {
-    x <- x %>% relocate(all_of(c("geo_value", other_keys, "time_value")), .before = 1)
+    all_names <- names(x)
+    ukey_names <- c("geo_value", other_keys, "time_value")
+    value_names <- all_names[!all_names %in% ukey_names]
+    x <- x[c(ukey_names, value_names)]
```
dshemetov, line 195: Was this noticeably faster? Good to know, if so.

R/methods-epi_df.R
```diff
@@ -153,7 +164,30 @@ dplyr_reconstruct.epi_df <- function(data, template) {
...
+#' [`dplyr::dplyr_extending`] for `epi_df`s, we moved a lot of checks in
+#' `dplyr_reconstruct` and used it instead of [`reclass()`] in various
+#' operations to prevent operations from outputting invalid metadata/classes,
+#' instead of more careful tailored and relevant checks. However, this actually
+#' introduced extra overhead due to `dplyr_reconstruct.epi_df()` passing off to
+#' `dplyr_reconstruct.grouped_df()` when grouped, which assumes that it will
+#' need to / should for safety recompute the groups, even when it'd be safe for
+#' it not to do so. In many operations, we're using `NextMethod()` to dispatch
+#' to `grouped_df` behavior if needed, and it should output something with valid
+#' groupings.
```
dshemetov, line 182: Thanks for this description, this is v handy. Also good catch on finding this intervention!

R/methods-epi_df.R
```diff
@@ -198,19 +232,40 @@ dplyr_reconstruct.epi_df <- function(data, template) {
...
+
+  reconstruct_light_edf(res, x)
+}
+
+#' @export
+`$<-.epi_df` <- function(x, ...) {
+  res <- NextMethod()
+
+  reconstruct_light_edf(res, x)
 }
```
dshemetov, line 257: basic question: Is this a way to intercept dplyr_reconstruct? Where does res <- NextMethod() go and does it really avoid dplyr_reconstruct?

R/methods-epi_df.R
```diff
@@ -404,7 +459,7 @@ arrange_row_canonical.default <- function(x, ...) {
 arrange_row_canonical.epi_df <- function(x, ...) {
   rlang::check_dots_empty()
   cols <- key_colnames(x)
-  x %>% dplyr::arrange(dplyr::across(dplyr::all_of(cols)))
+  x[vctrs::vec_order(x[cols]), ]
```
dshemetov, line 462: side-note: would be nice to get a "top 10 vctrs tricks" walkthrough from you at some point, might be nice for the rest of us to benefit from your learnings!

R/slide.R
```diff
@@ -309,15 +309,16 @@ epi_slide <- function(
     ...,
     .keep = TRUE
   ) %>%
-    bind_rows() %>%
-    filter(.real) %>%
-    select(-.real) %>%
+    {
+      vec_rbind(!!!.)
```
dshemetov, line 313: This incantation looks so strange. Is it much faster than bind_rows? Or does it have other benefits?

R/slide.R
```diff
@@ -309,15 +309,16 @@ epi_slide <- function(
...
+    } %>%
+    `[`(.$.real, names(.) != ".real") %>%
```
dshemetov, line 315: Is this another instance of avoiding tidyselect for performance?

R/epi_df.R
brookslogan: This particular change seemed to help a lot, but that was likely because it was avoiding group reconstruction from us calling dplyr_reconstruct.  Another change avoids dplyr_reconstruct which avoids some of that same overhead.  (I re-ordered some commits to batch together all these inner loop tidyeval replacements, so history probably doesn't reflect the original order/benefit.)

Looking at this change atop the (current history of) previous commits:
- This individual change makes `arrange_col_canonical` take 20% of the old time.
- This individual change makes the epix_epi_slide_opt WIP take 97% of its old time.
- The tidyeval replacements together make the epix_epi_slide_opt WIP take 60% of the old time.

R/methods-epi_df.R
brookslogan: The `reconstruct_light_edf` is a way to avoid a full `dplyr_reconstruct`.  But these methods probably weren't calling `dplyr_reconstruct` anyway.  I added these `epi_df` implementations because without them, these methods drop epi_df-ness when called on grouped epi_dfs, and output grouped tibbles instead.

> Where does res <- NextMethod() go

If we're grouped, to `$<-.grouped_df`, which in turn uses `NextMethod()` to call `$<-.tbl_df`.  If we're ungrouped, then it just goes to `$<-.tbl_df`.

That's also where these methods originally dispatched to before these `$<-.epi_df` was added:
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(epiprocess)
#> Loading required package: epidatasets
#> Registered S3 method overwritten by 'tsibble':
#>   method               from 
#>   as_tibble.grouped_df dplyr
#> 
#> 
#>     filter
edf <- covid_case_death_rates_extended
gedf <- edf %>% group_by(geo_value)
sloop::s3_dispatch(`$<-`(edf, bumped_case_rate, edf$case_rate + 1e-12))
#>    $<-.epi_df
#> => $<-.tbl_df
#>    $<-.tbl
#>  * $<-.data.frame
#>    $<-.default
#>  * $<- (internal)
sloop::s3_dispatch(`$<-`(gedf, bumped_case_rate, edf$case_rate + 1e-12))
#>    $<-.epi_df
#> => $<-.grouped_df
#> -> $<-.tbl_df
#>    $<-.tbl
#>  * $<-.data.frame
#>    $<-.default
#>  * $<- (internal)
```

<sup>Created on 2025-02-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

From `?sloop::s3_dispatch`:
```
        • '=>' method exists and is found by 'UseMethod()'.
        • '->' method exists and is used by 'NextMethod()'.
        • '*' method exists but is not used.
        • Nothing (and greyed out in console): method does not exist.
```

With the `epi_df` impls:
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(epiprocess)
#> Loading required package: epidatasets
#> Registered S3 method overwritten by 'tsibble':
#>   method               from 
#>   as_tibble.grouped_df dplyr
#> 
#> 
#>     filter
edf <- covid_case_death_rates_extended
gedf <- edf %>% group_by(geo_value)
sloop::s3_dispatch(`$<-`(edf, bumped_case_rate, edf$case_rate + 1e-12))
#>    $<-.epi_df
#> => $<-.tbl_df
#>    $<-.tbl
#>  * $<-.data.frame
#>    $<-.default
#>  * $<- (internal)
sloop::s3_dispatch(`$<-`(gedf, bumped_case_rate, edf$case_rate + 1e-12))
#>    $<-.epi_df
#> => $<-.grouped_df
#> -> $<-.tbl_df
#>    $<-.tbl
#>  * $<-.data.frame
#>    $<-.default
#>  * $<- (internal)
```

<sup>Created on 2025-02-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

If one of the upstream methods calls `dplyr_reconstruct` then we can't avoid it.  But I think they probably have a saner implementation upstream and only do group reconstruction if the object is grouped and the grouping columns are changed.  And we avoid triggering it ourselves with the "light" edf reconstruction.

R/slide.R
brookslogan: At some point it seemed to give a substantial performance benefit, but now it seems like it might be a wash.  `vec_rbind` seems a bit cleaner in that it determines the output ptype by looking at all the arguments and finding something compatible, rather than `dplyr_reconstruct`-ing with the first argument as a template.  I do wish that they had a `list_rbind` (oh hey, there's one in `purrr`! though might need to check overhead).

The incantation is the combination of these two:
- `x %>% { f(g(.)) }` --- call f(g(x)), not, f(x, g(x))
- `vec_rbind(!!! list_of_data_frames)` (`vec_rbind` expects things like `vec_rbind(df1, df2, df3)`, but supports splatting)

brookslogan: Side note: `vec_cbind` also seems to try to find a compatible ptype it could cast all the entries to... which doesn't make any sense.  I forget exactly why I moved to it; I think it seemed like `bind_cols` had some overhead and/or output `epi_df`s when working with tibbles was faster, and I didn't want to make things look visually even messier with `[<-`, though operation-wise that'd probably be cleanest.

brookslogan: Immediate issue with `[<-` was actually recycling of the group key.  I found another approach that looks a little bit cleaner than `vec_cbind()` or `bind_cols()` IMO.  Also I moved to using `list_rbind()` above.

R/slide.R
brookslogan: Yep.  It's also probably a bit more efficient just from combining row & col slicing when you have only a few columns.

Issue #173, opened 2022-07-27
Add special constructor function for an epi_archive based on a list/data.table of snapshots


brookslogan:
See #172 for background.  This would work around some of the caveats noted there by:
- Detecting row deletions, probably just representing them as an update row with all non-key columns NA.
- For each explicit version, check whether a snapshot of `next_after(version)` was provided.  If not, insert a snapshot revising all observations to NA with this version tag (`next_after(version)`).  ---- This is pretty inefficient space-wise.  Maybe we would want to do something with the design of epi_archive to make more efficient options available.

brookslogan:
We can be a lot more memory-friendly than an in-memory list/df of snapshots, by only reading in one / a few snapshots at a time, and compactifying along the way (plus suggesting to save the compactified version on disk if we compactified a lot, or more complicated things regarding proposed "updating" archives that are out of scope for this initial feature).  [This could also be useful for getting lower-latency starts to interactive nowcast flipbook visualizations.]

brookslogan:
Status: I have the iterator stuff ready in a separate repo + a version of a `map_ea`.  Documenting, testing, and refactoring to reduce duplication with existing stuff means it will still take some work to incorporate.

brookslogan:
Going to punt this down the road in favor of doc updates.  See `map_ea` [here](https://github.com/brookslogan/scratchpads/blob/43e2bbafb148629df8f2455d7a4045bfcfd30021/archive-agg/archive-agg-pack.R#L359), iterators [here](https://github.com/brookslogan/scratchpads/blob/main/archive-agg/sized-iterators.R) (iterb approach seems probably preferable), and `tbl_{diff2,patch}` in #611.

PR #611, opened 2025-02-26
Add specialized epix_slide for epi_slide_opt

### LINKED ISSUE #609: Implement smart sparse archive -> archive slide

brookslogan:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

- Resolves #609.

### TODO

- [ ] Address review comments
- [ ] Address change in `epi_slide_opt(.before = Inf)` behavior from refactor (CHECK check not running at the moment despite some others running?)


brookslogan:
- [x] handle all possible .f, before
- [x] finish docs, tests
- [-] port map_ea at least, maybe base on iterators
  - handle in #173 or beyond
- [x] in-source todos
- [ ] check whether frollapply on function that errors on ref time values not in `ref_time_values` will error in current scheme; consider adding skipping layer for it or documenting

nmdefries (commented):
Various nits and some questions. The logic looks fine to me, but given the complexity of this change, someone else should review as well.

R/patch.R
```diff
@@ -0,0 +1,348 @@
...
+  combined_n <- nrow(combined_tbl)
+
+  # We'll also need epikeytimes and value columns separately:
+  combined_ukeys <- combined_tbl[ukey_names]
+  combined_vals <- combined_tbl[val_names]
+
+  # We have five types of rows in combined_tbl:
+  # 1. From earlier_snapshot, no matching ukey in later_tbl (deletion; turn vals to
+  #    NAs to match epi_archive format)
+  # 2. From earlier_snapshot, with matching ukey in later_tbl (context; exclude from
```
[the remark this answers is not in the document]
brookslogan: It gives context for what to do with the corresponding row in `later_tbl`, which will be category 3/4.  `combined_tbl` has earlier followed by later, so ` vec_duplicate_id(combined_ukeys)` will point from the category 3/4 row back to the corresponding category 2 row; we can then look at the measurement cols to decide whether we're in case 3. or in case 4.  The category 2. row should be thrown out in either case; if the later row's category 3 then we should not have this ukey in the diff result at all, and if the later row's category 4, then we should only have that later row in the diff result.

R/patch.R
```diff
@@ -0,0 +1,348 @@
...
+  # More input validation:
+  if (!identical(tbl_names, names(later_tbl))) {
+    cli_abort(c("`earlier_snapshot` and `later_tbl` should have identical column
+                 names and ordering.",
+      "*" = "`earlier_snapshot` colnames: {format_chr_deparse(tbl_names)}",
+      "*" = "`later_tbl` colnames: {format_chr_deparse(names(later_tbl))}"
+    ))
+  }
+
+  combined_tbl <- vec_rbind(earlier_snapshot, later_tbl)
```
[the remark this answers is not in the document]
brookslogan: Performance was one, with a `full_join`-based approach taking about 4x the time of this approach.  (That `full_join` approach didn't consider compactification tol, which means considering it would probably make it a bit slower still.)

I had a pretty fast, pretty legible approach in terms of anti joins that used a simpler form of this duplicate-logic to make a faster anti join.  Unfortunately, I had troubles implementing tolerance in an efficient way there, but could take another look if this is too much gibberish.

R/epi_slide_opt_archive.R
```diff
@@ -0,0 +1,221 @@
...
+      }
+      # We need to use the entire input snapshot range, filling in time gaps. We
+      # shouldn't pad the ends.
+      slide_t_min <- min(inp_snapshot$time_value)
+      slide_t_max <- max(inp_snapshot$time_value)
+    } else {
+      # If the input had updates in the range t1..t2, this could produce changes
+      # in slide outputs in the range t1-after..t2+before, and to compute those
+      # slide values, we need to look at the input snapshot from
+      # t1-after-before..t2+before+after. nolint: commented_code_linter
```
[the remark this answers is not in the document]
brookslogan: Suppose `before = 6, after = 0` and there's an update to the 1d value at time t alone.  We need to compute the trailing 7dav for t..t+6.  But in order to compute the trailing 7dav for time t, we need 1d data from t-6..t.  So we end up needing the union of the ranges t-6..t, t-5..t+1, t-4..t+2, ..., t..t+6, which gives us t-6..t+6.

R/epi_slide_opt_archive.R
```diff
@@ -0,0 +1,221 @@
...
+      for (col_i in seq_along(in_colnames)) {
+        if (before == Inf) {
+          slide[[out_colnames[[col_i]]]] <-
+            f_dots_baked(slide[[in_colnames[[col_i]]]], seq_len(slide_nrow), adaptive = TRUE)
+        } else {
+          out_col <- f_dots_baked(slide[[in_colnames[[col_i]]]], before + after + 1L)
+          if (after != 0L) {
+            # data.table always puts NAs at tails, even with na.rm = TRUE; chop
+            # off extra NAs from beginning and place missing NAs at end:
+            out_col <- c(out_col[seq(after + 1L, slide_nrow)], rep(NA, after))
```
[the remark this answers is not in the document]
brookslogan: Nice catch!  I think I was worried about some garbage-rather-than-NA being in the tails which is why I didn't just copy that approach, but on second look that worry seems unfounded.

R/epi_slide_opt_archive.R
```diff
@@ -0,0 +1,221 @@
...
+          # Only include time_values that appeared in the input snapshot:
+          !is.na(slide_inp_backrefs)
+      }
+    out_update <- slide[rows_should_keep, ]
+    out_diff <- tbl_diff2(prev_out_snapshot, out_update, "time_value", "update")
+    out_snapshot <- tbl_patch(prev_out_snapshot, out_diff, "time_value")
+    prev_inp_snapshot <<- inp_snapshot
+    prev_out_snapshot <<- out_snapshot
+    out_diff$version <- version
+    out_diff
```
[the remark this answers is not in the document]
brookslogan: Perhaps; some thoughts:
- Recording all full `out_snapshot`s seems like it will potentially give a very hefty result, and they could get that same result by something like `epix_slide(~ .x)` on the result or a helper function doing something like that.
- Recording just `out_snapshot` restricted to times that were in `inp_update` plus any times that are required for validity (e.g., an change in just the time t 1d value producing an update to the trailing 7dav for t+1..t+6 as well as t) seems maybe more useful.  I'm not sure how hard/slow this would be; I had some issues trying to selectively apply `compactify_tol` to the 7dav/etc. cols and set aside that idea.

R/patch.R
```diff
@@ -0,0 +1,348 @@
...
+#'
+#' @keywords internal
+tbl_diff2 <- function(earlier_snapshot, later_tbl,
+                      ukey_names,
+                      later_format = c("snapshot", "update"),
+                      compactify_abs_tol = 0) {
+  # Most input validation + handle NULL earlier_snapshot. This is a small function so
+  # use faster validation variants:
+  if (!is_tibble(later_tbl)) {
+    cli_abort("`later_tbl` must be a tibble")
```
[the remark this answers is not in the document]
brookslogan: Added!

R/epi_slide_opt_archive.R
```diff
@@ -0,0 +1,221 @@
...
+      as_tibble(.name_repair = "minimal") %>%
+      # 0 rows input -> 0 rows output, so we can just say drop = TRUE:
+      grouped_df(epikey_names, TRUE)
+    if (use_progress) progress_bar_id <- cli::cli_progress_bar(.progress, total = n_groups(updates_grouped))
+    result <- updates_grouped %>%
+      group_modify(function(group_values, group_key) {
+        group_updates <- group_values %>%
+          nest(.by = version, .key = "subtbl") %>%
+          arrange(version)
+        # TODO move nesting inside the helper?
```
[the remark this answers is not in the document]
brookslogan: Should be addressed now.

R/epi_slide_opt_archive.R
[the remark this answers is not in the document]
brookslogan: Hmm, actually moving the first `after` values to the end of the result seems like it might not impact the output at all, at least for `epi_slide_opt.epi_df`, as these appear to all be non-`.real` (corresponding to padding `time_value`s) and excluded from the final result anyway.  We may need to intercept & manually handle `fill =`, or disallow it.

R/epi_slide_opt_archive.R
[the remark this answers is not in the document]
brookslogan: Just pushing something to disallow `fill =` altogether.  Refiling to #646.

R/epi_slide_opt_archive.R
```diff
@@ -0,0 +1,221 @@
...
+          slide[[out_colnames[[col_i]]]] <- out_col
+        }
+      }
+    } else if (f_from_package == "slider") {
+      for (col_i in seq_along(in_colnames)) {
+        slide[[out_colnames[[col_i]]]] <- f_dots_baked(slide[[in_colnames[[col_i]]]], before = before, after = after)
+      }
+    } else {
+      cli_abort("epiprocess internal error: `f_from_package` was {format_chr_deparse(f_from_package)}")
+    }
```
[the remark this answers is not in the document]
brookslogan: Incorporated!  (Diff was outdated so I needed to manually fiddle with it.)

dsweber2 (commented):
Not done with my review by any means, but I think it probably makes more sense to do a comment at a time instead of bundling them all.

R/epi_slide_opt_archive.R
```diff
@@ -0,0 +1,230 @@
...
+    slide_nrow <- time_delta_to_n_steps(slide_t_max - slide_t_min, time_type) + 1L
+    slide_time_values <- slide_t_min + 0L:(slide_nrow - 1L) * unit_step
+    slide_inp_backrefs <- vec_match(slide_time_values, inp_snapshot$time_value)
+    # Get additional values needed from inp_snapshot + perform any NA
+    # tail-padding needed to make slider results a fixed window size rather than
+    # adaptive at tails + perform any NA gap-filling needed:
+    slide <- inp_snapshot[slide_inp_backrefs, ]
+    slide$time_value <- slide_time_values
+    if (f_from_package == "data.table") {
```
dsweber2, line 90: This is the only branch that's actually tested right?

R/epi_slide_opt_archive.R
```diff
@@ -0,0 +1,230 @@
...
+        # computation outputs:
+        vec_rep_each(c(FALSE, TRUE, FALSE), c(before, slide_nrow - before - after, after)) &
+          # Only include time_values that appeared in the input snapshot:
+          !is.na(slide_inp_backrefs)
+      }
+    out_update <- slide[rows_should_keep, ]
+    out_diff <- tbl_diff2(prev_out_snapshot, out_update, "time_value", "update")
+    out_snapshot <- tbl_patch(prev_out_snapshot, out_diff, "time_value")
+    prev_inp_snapshot <<- inp_snapshot
+    prev_out_snapshot <<- out_snapshot
```
dsweber2, line 132:
```suggestion
    prev_inp_snapshot <<- inp_snapshot
    prev_out_snapshot <<-  tbl_patch(prev_out_snapshot, out_diff, "time_value")
```

dsweber2: `out_snapshot` isn't used anywhere else?

R/patch.R
```diff
@@ -0,0 +1,375 @@
...
+#'   non-`ukey_names` columns to NA values (using `{vctrs}`). If "update", then
+#'   it's assumed that any deletions have already been represented this way in
+#'   `later_tbl` and any ukeys not in `later_tbl` are simply unchanged; we are
+#'   just ensuring that the update is fully compact for the given
+#'   `compactify_abs_tol`.
+#' @param compactify_abs_tol compactification tolerance; see `apply_compactify`
+#' @return a tibble in compact "update" (diff) format
+#'
+#' @keywords internal
+tbl_diff2 <- function(earlier_snapshot, later_tbl,
```
dsweber2, line 204: not really important but why isn't this just `tbl_diff`?

R/patch.R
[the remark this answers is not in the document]
dsweber2: If we're keeping this in this format, it might be a good idea to directly test it, e.g. against an implementation that just uses `full_join`s on some randomly generated `earlier_snapshot` `later_tbl` pairs.

Alternatively, since we'll be compactifying later, the anti-join version that only tests direct equality might work fine. Occasionally including redundant values is less of a problem than excluding unique ones.

I do think you did a pretty good job of documenting it!

R/patch.R
```diff
@@ -0,0 +1,375 @@
...
+      inds1 = combined_ukey_is_repeat,
+      inds2 = ukey_repeat_first_i
+    )
+
+  # Which rows from combined are in cases 3., 4., or 5.?
+  combined_from_later <- vec_rep_each(c(FALSE, TRUE), c(earlier_n, later_n))
+
+  if (later_format == "update") {
+    # Cases 4. and 5.:
+    combined_tbl <- combined_tbl[combined_from_later & !combined_compactify_away, ]
```
dsweber2, line 307: Case 1 isn't relevant here because it's indistinguishable from unchanged rows when we're just updating, right?

R/epi_slide_opt_archive.R
```diff
@@ -0,0 +1,230 @@
...
+  result <- list_rbind(result)
+  result
+}
+
+#' @method epi_slide_opt grouped_epi_archive
+#' @export
+epi_slide_opt.grouped_epi_archive <- function(.x, ...) {
+  assert_set_equal(
+    group_vars(.x),
+    key_colnames(.x$private$ungrouped, exclude = c("time_value", "version"))
```
dsweber2, line 145: I guess we ban grouping by non-keys? That does make sense, first time I think I've actually run into it though.

dsweber2: actually would this ban grouping by a subset of key_colnames? that seems like something that we would want to support

R/epi_slide_opt_archive.R
```diff
@@ -0,0 +1,230 @@
...
+        res <- epi_slide_opt_archive_one_epikey(
+          group_values,
+          names_info$input_col_names,
+          .f_dots_baked, .f_info$from_package, window_args$before, window_args$after, time_type,
+          names_info$output_col_names
+        )
+        if (use_progress) cli::cli_progress_update(id = progress_bar_id)
+        res
+      }) %>%
+      as.data.frame() %>% # data.table#6859
```
dsweber2, line 220: This issue? looks like they fixed it. https://github.com/Rdatatable/data.table/issues/6859

I guess you're trying to avoid having to pin data.table to a recent version?

R/epi_slide_opt_edf.R
```diff
@@ -0,0 +1,705 @@
...
+    if (identical(.window_size, Inf)) {
+      n <- "running_"
+      time_unit_abbr <- ""
+      align_abbr <- ""
+    } else {
+      n <- time_delta_to_n_steps(.window_size, time_type)
+      time_unit_abbr <- time_type_unit_abbr(time_type)
+      align_abbr <- c(right = "", center = "c", left = "l")[[.align]]
+    }
+    glue_env <- rlang::env(
```
dsweber2, line 134: This whole section seems overly indirect, but I guess it's so that `glue` can allow the user to set the prefix to e.g. `"{n}^{time_unit_abbr}"`  themselves or something like that?

R/patch.R
```diff
@@ -0,0 +1,375 @@
...
+    # Matching vec_equal, we ignore names and other attributes.
+    if (!is.null(inds1)) vec1 <- vec_slice(vec1, inds1)
+    if (!is.null(inds2)) vec2 <- vec_slice(vec2, inds2)
+    na_or_nan1 <- is.na(vec1)
+    na_or_nan2 <- is.na(vec2)
+    # Since above are bare logical vectors, we can use `fifelse`
+    res <- fifelse(
+      !na_or_nan1 & !na_or_nan2,
+      abs(vec1 - vec2) <= abs_tol,
+      if (na_equal) {
```
dsweber2, line 127: doesn't this break the entire point of `fifelse` by forcing it to use an R for loop?

R/patch.R
```diff
@@ -0,0 +1,375 @@
...
+        na_or_nan1 & na_or_nan2 & (is.nan(vec1) == is.nan(vec2))
+      } else {
+        # Like `==` and `vec_equal`, we consider NaN == {NA, NaN, anything else}
+        # to be NA.
+        NA
+      }
+    )
```
dsweber2, line 134:
```suggestion
    if (na_equal) {
      res <- fifelse(
        !na_or_nan1 & !na_or_nan2,
        abs(vec1 - vec2) <= abs_tol,
        na_or_nan1 & na_or_nan2 & (is.nan(vec1) == is.nan(vec2))
      )
    } else {
      # Like `==` and `vec_equal`, we consider NaN == {NA, NaN, anything else}
      # to be NA.
      res <- fifelse(
        !na_or_nan1 & !na_or_nan2,
        abs(vec1 - vec2) <= abs_tol,
        NA
      )
    }
```

dsweber2: doesn't this break the entire point of fifelse by forcing it to use an R for loop?

R/patch.R
```diff
@@ -0,0 +1,375 @@
...
+    # (we only need to manually recurse if we potentially have columns that would
+    # be specially processed by the above)
+    if (ncol(vec1) == 0) {
+      rep(TRUE, nrow(vec1))
+    } else {
+      Reduce(`&`, lapply(seq_len(ncol(vec1)), function(col_i) {
+        vec_approx_equal0(vec1[[col_i]], vec2[[col_i]], na_equal, abs_tol, inds1, inds2)
+      }))
+    }
+  } else if (is_bare_list(vec1) && abs_tol != 0) {
```
dsweber2, line 152: I don't think this branch is getting tested at all.

tests/testthat/test-archive.R
```diff
@@ -92,7 +92,7 @@ test_that("epi_archives are correctly instantiated with a variety of data types"
   expect_equal(key(ea1$DT), c("geo_value", "time_value", "version"))
 
   ea2 <- as_epi_archive(df, other_keys = "value", compactify = FALSE)
-  expect_equal(key(ea2$DT), c("geo_value", "time_value", "value", "version"))
+  expect_equal(key(ea2$DT), c("geo_value", "value", "time_value", "version"))
```
dsweber2, line 95: where did this order change come from? It does seem like an odd order; why isn't it geo_value, time_value, version, value?

tests/testthat/test-compactify.R
```diff
@@ -124,7 +124,7 @@ quantile_pred_once <- function(estimates_vec, levels_vec) {
   hardhat::quantile_pred(t(as.matrix(estimates_vec)), levels_vec)
 }
 test_that("compactify works on distributions", {
-  skip("Until #611 is merged or hardhat/epipredict is patched")
+  skip("See #631.")
```
dsweber2, line 127: weird. Dan wrote both, so you may just want to talk to him about it

tests/testthat/test-epi_slide_opt_archive.R
```diff
@@ -0,0 +1,219 @@
...
+    } %>%
+    filter(geo_value %in% head(unique(geo_value), 4L)) %>%
+    as_epi_archive()
+
+  mini_case_death_rate_archive_time_opt <- system.time(
+    mini_case_death_rate_archive_result <- mini_case_death_rate_archive %>%
+      epi_slide_opt(case_rate, frollmean, .window_size = 7)
+  )
+
+  mini_case_death_rate_archive_time_gen <- system.time(
```
dsweber2, line 155: So comparing these times, it's ~100x faster? Is this representative of other experiments you've done? That's certainly pretty nice if it is.

tests/testthat/test-epi_slide_opt_archive.R
```diff
@@ -0,0 +1,219 @@
...
+  expect_equal(
+    mini_case_death_rate_archive_result2,
+    mini_case_death_rate_archive_result %>%
+      group_by(geo_value)
+  )
+
+  mini_case_death_rate_archive_b <- mini_case_death_rate_archive %>%
+    {
+      as_tibble(as.data.frame(.$DT))
+    } %>%
```
dsweber2, line 179:
```suggestion
  mini_case_death_rate_archive_b <-
    mini_case_death_rate_archive$DT %>%
    as.data.frame() %>%
    as_tibble()

```

tests/testthat/test-epi_slide_opt_archive.R
```diff
@@ -0,0 +1,219 @@
...
+      group_by(age_group) %>%
+      epi_slide_opt(case_rate, frollmean, .window_size = 7)
+  )
+
+  archive_cases_dv_subset_time_opt <- system.time(
+    archive_cases_dv_subset_result <- archive_cases_dv_subset %>%
+      epi_slide_opt(percent_cli, frollmean, .window_size = 7)
+  )
+
+  archive_cases_dv_subset_time_gen <- system.time(
```
dsweber2, line 203: so this pair is a factor of ~11

dsweber2 (changes requested):
So I've got some qualms about the complexity and the relatively narrow set of tests, but it does seem to be working, and would be useful for porting a faster `daily_to_weekly` transform.

There's a number of nits which I'd appreciate you fixing, and I've made a PR with some notes and some of the nits changed

R/archive.R
```diff
@@ -464,56 +462,38 @@ update_is_locf <- function(arranged_updates_df, ukey_names, abs_tol) {
...
-  lag_vec <- lag(vec)
-  if (is.vector(vec, mode = "numeric") && !is_key) {
-    # (integer or double vector, no class (& no dims); maybe names, which we'll
-    # ignore like `vec_equal`); not a key column
-    unname(if_else(
-      !is.na(vec) & !is.na(lag_vec),
-      abs(vec - lag_vec) <= abs_tol,
-      is.na(vec) & is.na(lag_vec)
-    ))
+  n_updates <- nrow(arranged_updates_df)
```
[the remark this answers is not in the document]
brookslogan: Nice catch; this is outdated.  Before, we did use it within a `filter`, and we used `if_all`, which required dplyr-data-masking.

todo: remove

R/archive.R
```diff
@@ -536,11 +516,32 @@ as_epi_archive <- function(
...
     versions_end = .versions_end) {
   assert_data_frame(x)
+  # Convert first to data.frame to guard against data.table#6859 and potentially
+  # other things epiprocess#618:
+  x_already_copied <- identical(class(x), c("data.table", "data.frame"))
+  x <- as.data.frame(x)
   x <- rename(x, ...)
-  x <- guess_column_name(x, "time_value", time_column_names())
   x <- guess_column_name(x, "geo_value", geo_column_names())
+  if (!all(other_keys %in% names(x))) {
```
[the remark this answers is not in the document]
brookslogan: No behavior-driven reason.  We previously canonically ordered columns `geo_value, time_value, other_keys, [version,] <values>`; but have moved toward `geo_value, other_keys, time_value, [version,] <values>` in various places; I was just trying to match that.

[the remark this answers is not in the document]
dsweber2: Consistency is more important than the specific order. Good to know

R/archive.R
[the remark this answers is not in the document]
dsweber2: it's removed in the PR on this

R/epi_slide_opt_archive.R
```diff
@@ -0,0 +1,230 @@
...
+    in_colnames,
+    f_dots_baked, f_from_package, before, after, time_type,
+    out_colnames) {
+  grp_updates_by_version <- grp_updates %>%
+    nest(.by = version, .key = "subtbl") %>%
+    arrange(version)
+  unit_step <- unit_time_delta(time_type)
+  prev_inp_snapshot <- NULL
+  prev_out_snapshot <- NULL
+  result <- map(seq_len(nrow(grp_updates_by_version)), function(version_i) {
```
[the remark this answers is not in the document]
brookslogan: Right, that's the idea.  We also get the time range that was updated, allowing us to limit slide computation time ranges.  Comparing progressive patching vs. `epix_as_of`... I thought I did at some point, but now I'm questioning whether I did it in the right way, plus I've sped up some other internals.  Trying to do a quick comparison for 7dav-ing now:

- Splitting into tibbles for each epikey vs. operating on combined epi_dfs has 3.8x the speed in `archive_cases_dv_subset`, 0.8x on `case_death_rate_archive`.

The rest of this comparison is based on combined epi_dfs, so it may not be completely comparable.
- The special time logic actually slows things down on these two test cases, ... 0.82x / 0.93x speed of without.  However, this definitely didn't seem to be the case during development, so either some "incidental" difference or other improved internal made a difference.  These data sets are also probably not the best setup for the special time logic, as they actually have pretty large lag ranges recorded in updates.
- The progressive-patching to get inputs is 1.32x / 1.12x faster than `epix_as_of` querying, given that the time range logic has been removed.
- The progressive-diffing of the output is comparable to rbinding + compactifying all at once in terms of speed, but should avoid some potential RAM issues when dealing with larger archives; given all features above removed.
- Using dedicated code here vs. delegating to `epix_slide` is 1.14x / 31x (!) faster than `epix_slide`.

So... potential responses here: I could try
- A. removing the special time range logic and see the impact on current code.  (Potentially simpler & faster!  Or could point to something yet more complicated but yet faster.)
- B. removing the patching logic (might slow things down non-negligibly based on the above).
- C. locating the `epix_slide` overhead.
- D. rewriting `epix_slide` with progressive-patching logic.
- E. rewriting this code and eventually `epix_slide` with iterator code that separates the progressive-patching, sliding, and progressive-diffing code into more readable, re-usable chunks.
- F. try again to simplify `tbl_diff2`... this is likely very hard to do with `compactify_abs_tol` involved, though I guess I never actually used that feature in `epi_slide_opt`.  If we keep that feature out, then I think I probably have something faster & simpler sitting around.

Right now, I'm thinking maybe I should take a look at A here, and we should decide whether `compactify_abs_tol` should be supported here and whether you're willing to review a reimplementation from F.  The others... I'm not sure B. is that much of a simplification if progressive diffing is kept.  I'd like to do C. and/or E.+D. at some point, but separately.

Thoughts?

R/epi_slide_opt_archive.R
brookslogan: Good catch!  I've done some ad-hoc testing with other functions but actual tests are looking pretty homogeneous...  Should probably look into code coverage tools again.

R/patch.R
brookslogan: To attempt to clarify that it's a binary operation, unlike `diff`.

R/epi_slide_opt_archive.R
brookslogan: Like with the current `epi_slide_opt.epi_df`, we require the grouping to be by the epikeys.  There are some frustrating situations to think about when grouping by other things, such as scattered location drop-in & drop-out & how we should do completion.

R/epi_slide_opt_archive.R
brookslogan: I filed this issue as I was writing this PR.  We could do a version range requirement I guess; has this hit CRAN?

R/epi_slide_opt_edf.R
brookslogan: Something like that (e.g., `"_{.n}{.time_unit_abbr}_mean"`).  Idea was that if you're copy-pasting & adjusting epi_slide_opts, or tweaking epi_slide_opt parameters over time, it might be more conveniently correct.  Also originally it was intended to prevent "magical defaults" but not sure that was actually a problem.

R/patch.R
brookslogan: Don't think there's an R loop here.

Is suggestion for clarity or intended for performance?  (For clarity, what would you think about 
```r
      if (na_equal) {
        na_comparisons <- na_or_nan1 & na_or_nan2 & (is.nan(vec1) == is.nan(vec2))
      } else {
        # Like `==` and `vec_equal`, we consider NaN == {NA, NaN, anything else}
        # to be NA.
        na_comparisons <- NA
      }
      fifelse(
        [...],
        na_comparisons
      )
```
?)

tests/testthat/test-archive.R
brookslogan: This probably comes from us moving from geo_value, time_value, other_keys[, version] canonical ordering to geo_value, other_keys, time_value[, version].  It looks pretty weird here probably moreso due to using the column name "value" rather than "age_group" or something like that.

tests/testthat/test-compactify.R
brookslogan: Chatted with him a bit.  Submitted a PR to hardhat to fix this + some other issues & then we can remove a partial fix sitting in epipredict.

R/epi_slide_opt_edf.R
brookslogan: (This naming code was part of an earlier PR; it appears here because I put it inside of a helper function & in a new file.)

R/epi_slide_opt_archive.R
dshemetov: It's not in the 1.17.0 release, will need to wait a bit.

R/patch.R
dsweber2: right this is a thing that e.g. purrr uses

R/patch.R
dsweber2: For performance primarily. `fifelse` is inherently mapping across a vector of values. We're interrupting that mapping using a base R `if`. The alternative you're suggesting also seems fine to me.

R/epi_slide_opt_archive.R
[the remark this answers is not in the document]
dsweber2: > Trying to do a quick comparison for 7dav-ing now:

Could you share what you're comparing to in some way? If it is the case that we're only getting a 1.3x speed-up relative to e.g. a for loop+`epix_as_of`, I don't think the level of complexity we're adding in this PR is worth it. The 10-100x that show up in the tests comparing vs `epix_slide(epi_slide_opt())` definitely seems worthwhile, but if that's coming from potential improvements to `epix_slide`, it'd be better to re-write this to use `epix_slide` and improve that in a different PR.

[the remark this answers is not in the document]
brookslogan: The 1.3x is comparing [this](https://github.com/brookslogan/scratchpads/blob/bf0c1dae3ca8ef8368f26f8fda13e10b61799e9c/archive-agg/archive-agg.R#L2816) to [this](https://github.com/brookslogan/scratchpads/blob/bf0c1dae3ca8ef8368f26f8fda13e10b61799e9c/archive-agg/archive-agg.R#L2827).  

Here, `map_ea` is akin to `map_dfr`, but instead of collecting & rbinding the results, it
- Immediately diffs each result with the previous one.  The source of the desired max RAM usage improvement.
- Collects the diffs.
- Rbinds the diffs and converts to an archive.
However, this version of `map_ea` is not backed by `tbl_diff2` but an earlier `epi_diff2`, which is slower.

R/patch.R
brookslogan: I don't think there should be a performance difference.  I think the way this works out is
fifelse(test, yes, no)
- sets up a function evaluation environment with a, b, and c being delayed bindings / promises
- hits `.Call(CfifelseR, test, yes, no, na)`; FFI forces all arguments, causing the `no` promise to be evaluated (into a logical vector).
- FFI continues merrily on its way

But I don't see a problem with writing it this other way either.  Thanks for the edits in your PR!

R/patch.R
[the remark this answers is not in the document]
brookslogan: Good idea on the testing.

Regarding compactifying later... that might be a valid approach, though this actually this doesn't compactify later, since it should be redundant.  I was actually thinking that eventually we could try compactifying by splitting a DT apart by versions, then diffing consecutive versions & recombining.  (Looks more elegant with helper functions.)   That would give us more flexibility to do things like https://github.com/cmu-delphi/epiprocess/issues/642.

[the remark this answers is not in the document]
brookslogan: I've rewritten this in terms of anti joins & put the complex vctrs duplicate approaches inside a fast anti-join-with-tolerance function.  I think it reads better, and this iteration of it seems to actually be faster than the 5-case approach (while an older version of it was slower).  Does this seem less convoluted?

R/epi_slide_opt_archive.R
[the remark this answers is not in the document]
dsweber2: ok, I misunderstood, I thought the 1.3x you were talking about was for the entire set of optimizations, and not just `epix_as_of` vs `tbl_patch`. I'm not sure whether leaving it as is, or trying to pare back to a minimal set of improvements is preferable. On the one hand, maintaining the 2.3k loc in this PR is going to be a lot. On the other it would require yet more re-writes to pare this back. to a minimal set

R/patch.R
dsweber2: So I did some direct testing on this because I'm a bit confused
```r
  n_entries <- 1e8
  na_or_nan1 <- sample(c(TRUE, FALSE), n_entries, TRUE)
  na_or_nan2 <- sample(c(TRUE, FALSE), n_entries, TRUE)
  vec1 <- runif(n_entries)
  vec1[na_or_nan1] <- NA
  vec2 <- runif(n_entries)
  vec2[na_or_nan2] <- NA
  na_equal <- TRUE
  abs_tol <- 1e-8
  system.time(
    res <- fifelse(
      !na_or_nan1 & !na_or_nan2,
      abs(vec1 - vec2) <= abs_tol,
      if (na_equal) {
        na_or_nan1 & na_or_nan2 & (is.nan(vec1) == is.nan(vec2))
      } else {
        NA
      }
    )
  )
  system.time(
    {if (na_equal) {
      na_comparisons <- na_or_nan1 & na_or_nan2 & (is.nan(vec1) == is.nan(vec2))
    } else {
      na_comparisons <- NA
    }
    res <- fifelse(
         !na_or_nan1 & !na_or_nan2,
         abs(vec1 - vec2) <= abs_tol,
         na_comparisons
    )}
  )
```
Gives pretty comparable times of ~2.9s. I'm mostly confused how the `no` case of 
```r
      if (na_equal) {
        na_or_nan1 & na_or_nan2 & (is.nan(vec1) == is.nan(vec2))
      } else {
        NA
      }
```
is actually vectorized here. I guess it's only applying the case once, rather than in parallel.

R/patch.R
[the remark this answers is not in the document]
dsweber2: I definitely think so!

tests/testthat/test-archive.R
brookslogan: Sorry, actually we're not canonically ordering archive cols themselves in `as_epi_archive`, but are at least using the new canonically ordering of the vars in the key.  Change presumably comes from
```r
key_vars <- c("geo_value", other_keys, "time_value", "version")
```
line, but it's mixed up in some changes addressing data.table footguns so it's not quite obvious.

R/patch.R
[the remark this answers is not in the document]
brookslogan: Maybe.  Probably would need some generalization work then.  The `vec_` prefix is, as in `vctrs`, to denote that the first arg is a vctrs-sense vector, and to distance further from `all.equal`, which returns a scalar result.

R/epi_slide_opt_archive.R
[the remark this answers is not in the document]
brookslogan: Trying to break down LOC/doc (GitHub doesn't make this easy...) + collect some thoughts about where redundant implementation can be trimmed while keeping the same functions available + a few incomplete thoughts about paring functions down / obstacles.
```
<=230 epi_slide_opt.epi_archive
>=220 test-epi_slide_opt_archive.R
<=184 vec_approx_equal
<= 96 epi_slide_opt.epi_df (epi_slide_opt_edf.R + slide.R - old slide.R = 707 + 550 - 1161)
   79 tbl_diff2
   59 tbl_patch
>= 57 test-vec_approx_equal
   52 tbl_fast_anti_join

(can maybe get -66-34 on epi_slide_opt.epi_df or -138 epi_archive)
(can maybe get -31 on vec_approx_equal by removing inds{1,2})
(need to add tests)
(can maybe get -184-57 by contributing vec_approx_equal upstream... though likely adding a bunch of work)
(can maybe get -52 by contributing tbl_fast_anti_join upstream... though likely adding a bunch of work) 
(can maybe get very roughly -57 by making apply_compactify use tbl_diff2, if perf good enough)

(minimal slower & RAM-exploding epi_slide_opt.epi_archive likely maybe ~27 lines in epi_slide_opt.epi_archive)
(load-archives-from-snapshot-files will require tbl_diff2, tbl_fast_anti_join, vec_approx_equal... or a backend redesign)
```

- [ ] todo: @return for epi_slide_opt.epi_archive

R/epi_slide_opt_archive.R
```diff
@@ -0,0 +1,221 @@
...
+      cli_abort("epi_slide.epi_archive does not support the `.all_rows` argument")
+    }
+    assert(
+      checkmate::check_logical(.progress, any.missing = FALSE, len = 1L, names = "unnamed"),
+      checkmate::check_string(.progress)
+    )
+    if (isTRUE(.progress)) {
+      .progress <- "Time series processed:"
+    }
+    use_progress <- !isFALSE(.progress)
```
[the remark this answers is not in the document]
brookslogan: This would not use a progress bar if the user passed their own progress bar label string.

- [ ] todo: docs.  Maybe tests if that is possible?

R/patch.R
brookslogan: Let N be the length of the `test` vector.  `fifelse` evaluates `yes` and `no` when feeding it into its native function, expecting each to be either length-N or length-1; in the common `length(yes) == N, length(no) == N` case, it just straight up expects to be provided N+N values of `yes` and `no`, half of which are going to be thrown away due to `test` selecting them away.  (Hypothetically, we could try avoiding calculating values of `yes` and `no` that we know will be thrown away & ditch `fifelse`, but it probably won't look as nice and I don't know whether or not to expect it to actually be faster.)

If `na_equal` is `TRUE`, then we are feeding in a length-N vector for `yes` and a length-N vector for `no`; all the operations used to build the `no` vector are already vectorized (`&`, `is.nan`, `==`).  If `na_equal` is `FALSE`, then we are feeding in a length-N vector for `yes` and a length-1 vector for `no`; `fifelse` recycles the `no` vector (at least conceptually) to match the length of `test`.

R/epi_slide_opt_archive.R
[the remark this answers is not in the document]
nmdefries: I do think `fill` is convenient, but it's both not hard to implement (on our side or by the user) and doesn't change the results very much.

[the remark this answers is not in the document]
nmdefries: See https://github.com/cmu-delphi/epiprocess/pull/611#discussion_r2015320321

PR #650, opened 2025-04-01
ci: replace auto-ci with commands

dshemetov:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

* move style, lint, document, preview-docs all in pr-commands.yaml 
* all are now triggered with PR comments, instead of auto-running
* update PR template to remind user to run these commands

- Resolves #{issue number}


dsweber2:
Do we have these set up as optional in e.g. epipredict? Or is this a first

.github/workflows/pr-commands.yaml
```diff
@@ -0,0 +1,173 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+# Modifications:
+# - Allow more roles to trigger each PR command
+# - Document builds README.md from README.Rmd with devtools::build_readme()
+# - Include a lint command (from https://github.com/r-lib/actions/tree/v2/examples#lint-workflow)
```
brookslogan, line 6: We might still want linting to run automatically.  It occasionally catches bugs, and if we're going to check it before merging anyway, then it also reduces wait time.  Guess it might block merging if we want to ignore it, but I think we have the ability to override that anyway.

PR #651, opened 2025-04-02
Add `filter.epi_archive`

### LINKED ISSUE #231: Add `filter` for `epi_archive`

brookslogan:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

- Resolves #231

### TODO
- [x] Add tests
- [x] Complete cut-off comment
- [x] Consider forbidding use of value columns as well.  E.g., `value >= 10` will remove revisions of things >= 10 down to < 10, when the intent is probably to treat the epikeytime as missing in corresponding versions (e.g., putting an NA there).
- [-] Vignette changes?
  - Deferring to https://github.com/cmu-delphi/forecasting-team-meta/issues/29
- [-] Maybe implement a `select.epi_archive` and use it in the example to focus on the 7dav signal -> weekly res... not sure if taking Saturdays of Gaussian linear smoothed signal makes sense necessarily, though also not sure if we should have these lined up in the same data structure at all.
  - Deferring to separate PR sometime.


brookslogan:
@dshemetov GitHub suggested you for review, if you have time.  Experimented with some checks to try to error on things that might be expected to work but have details; see tests for some examples.  (Probably could map to "uncompactified" logic while remaining scalable & without a bunch of extra logic if we had a duckdb backend.)

R/methods-epi_archive.R
```diff
@@ -984,3 +984,136 @@ dplyr_col_modify.col_modify_recorder_df <- function(data, cols) {
...
+                  "Using `{format_varname(measurement_colname)}`
+                 in `filter.epi_archive` may produce unexpected results.",
+                  ">" = "See `?filter.epi_archive` details for how to proceed."
+                ), class = "epiprocess__filter_archive__used_measurement"), assign.env = e)
+              }
+              break
+            }
+            e <- parent.env(e)
+          }
+          TRUE
```
dshemetov, line 1082: This seems like it will error if you use it inside a function that has "version" or "time_value" defined in its environment? I'm reading this as traversing up the environment chain and stopping short of the globalenv(), which would be most likely to have variables like that defined, but intermediate scopes might still have false positives.

dshemetov: Ah ok, I tested it locally and it seems fine. I guess this works because we hit the data mask environment first and break out before we hit the user's function environment? Seems reasonable.

dshemetov (approved):
Generally, this looks good to me. Hard to anticipate all future issues with the archive metadata after the filtering, but I think you made the right calls in choosing the defaults (only reinferring the time_type in the day case and inheriting otherwise) and explaining the counter-intuitive aspects.

R/methods-epi_archive.R
brookslogan: Yep, the data mask environment chain in current dplyr&rlang looks something like
rlang wrapper env -> data bindings env (/ env chain input to as_data_mask env) -> quosure env (chain)
where:
- rlang wrapper env holds the real data pronoun objects, a `~` override I don't quite understand, and some other internals
- data bindings env is typically just a single env holding column bindings; in other contexts, it could be an env chain we fed into as_data_mask (with its "top" ancestor reassigned to point at:)
- the quosure env

We should stop at the data bindings env and reassign things there.

But I did find an issue along those lines
```r
epidatasets::case_death_rate_archive %>% {.$DT <- copy(.$DT)[, e := 1]; .} %>% filter(e < 2)
#> Error in `filter()`:
#> ℹ In argument: `e < 2`.
#> Caused by error in `e < 2`:
#> ! comparison (<) is possible only for atomic and list types
#> Run `rlang::last_trace()` to see where the error occurred.
```
because I'm leaving around an `e` in the rlang wrapper env.

Also, I fell for a classic lazy eval + env issue
```r
epidatasets::case_death_rate_archive %>% filter(case_rate_7d_av < 2)
#> Error in `filter()`:
#> ℹ In argument: `case_rate_7d_av < 2`.
#> Caused by error:
#> ! Using `death_rate_7d_av` in `filter.epi_archive` may produce unexpected results.
#> → See `?filter.epi_archive` details for how to proceed.
#> Run `rlang::last_trace()` to see where the error occurred.
```

brookslogan:
- [ ] clean up `e` binding etc.
- [ ] fix using presumably the last measurement colname in error for everything

R/methods-epi_archive.R
brookslogan: ... and found some others.  Should be fixed now.

brookslogan:
@dshemetov along with the lazy eval gotcha fixes, I've patched some issues regarding Date typeof strictness in the package code and in the testing code.  Maybe I'll just merge for now, but if the latter part raises flags in your mind, please note in https://github.com/cmu-delphi/epiprocess/issues/662 and up its priority & timeline.

PR #685, opened 2026-01-20
Linelist to archive function

### LINKED ISSUE #620: Add line list -> epi_df, epi_archive utilities

JavierMtzRdz:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
- [x] Styling and documentation checks. Make a PR comment with:
  process.

### Adds `linelist_to_archive()` function

Converts a line list (where each row represents a case) into an `epi_archive` object. This generates a versioned time series of aggregated counts, allowing you to see how the daily cases would have appeared at any specific point in history.

- Resolves #620


JavierMtzRdz:
I removed the lint enforcement for the new pipe and the implicit return, as done in the [`.lintr` doc in `epidatr`](https://github.com/cmu-delphi/epidatr/blob/dev/.lintr). It returned errors even when there were none.  Now checks have passed.

R/linelist.R
```diff
@@ -0,0 +1,320 @@
...
+#'   * `version_recorded`: the time at which the event became known/recorded.
+#'   * `version_deleted`: (optional) the time at which the event was
+#'     removed/deleted. If `NULL` (default), it is assumed no events are
+#'     deleted.
+#'   * `other_keys`: (optional) additional key columns (e.g. age group).
+#' @param is_deleted (optional) <[`tidy-select`][dplyr::dplyr_tidy_select]>
+#'   Column in `x` (or a logical vector/predicate) indicating if the row is a
+#'   deletion (`TRUE`/1) or an entry (`FALSE`/0). Only used for "chart-style"
+#'   linelists (where `version_recorded` and `version_deleted` are the same
+#'   column).
```
brookslogan: suggestion: either of these two alternative interfaces:
- separate function for chart-style
- rather than making `version_recorded` and `version_deleted` the same + provide `is_deleted`, make `version_deleted` and `is_deleted` mutually exclusive

suggestion: either:
- rename `is_deleted` -> `is_deletion`
- or call it `update_type` and require it to be a character/factor with possibilities "recorded", "deleted"

R/linelist.R
```diff
@@ -0,0 +1,320 @@
...
+  if (is_chart_style) {
+    if (is.null(is_del_col)) {
+      cli::cli_abort("If `version_recorded` and `version_deleted` are the same column, `is_deleted` must be provided.")
+    }
+
+    # is_del must be valid
+    raw_is_del <- x[[is_del_col]]
+    if (anyNA(raw_is_del)) cli::cli_abort("`{is_del_col}` must not contain NAs.")
+    is_del <- as.logical(raw_is_del)
+    if (anyNA(is_del)) cli::cli_abort("`{is_del_col}` must be coercible to logical without generating NAs.")
```
brookslogan: suggestion: just require it to be logical to begin with

R/linelist.R
```diff
@@ -0,0 +1,320 @@
...
+    ent_sub <- entries_df[c(id_col, ver_rec_col)]
+    rem_sub <- removals_df[c(id_col, ver_del_col)]
+
+    names(ent_sub) <- c("id", "ver_rec")
+    names(rem_sub) <- c("id", "ver_del")
+
+    common <- dplyr::inner_join(ent_sub, rem_sub, by = "id")
+
+    if (any(common$ver_del < common$ver_rec)) {
+      cli::cli_abort("`{ver_del_col}` (removal) must be >= `{ver_rec_col}` (entry).")
```
brookslogan, line 362: suggestion: also check for time_value matching

R/linelist.R
```diff
@@ -0,0 +1,320 @@
...
+  # Select cols
+  sel_cols <- c(geo_col, time_col, other_cols, v_col)
+  out <- df[sel_cols]
+
+  # Rename to standard
+  names(out) <- c("geo_value", "time_value", other_cols, "version")
+
+  out %>%
+    dplyr::count(
+      geo_value, time_value, dplyr::across(dplyr::all_of(other_cols)), version,
```
brookslogan: request: we're gradually trying to standardize more toward geo, other, time, version ordering; can you please switch column and helper function argument naming to that order?  The archive constructor might already re-order the result so it might not really matter though.

R/linelist.R
brookslogan: I've pushed a change that incorporates this as part of it but it needs checks

brookslogan (commented):
Looks pretty good!  I've suggested a few changes + pushed a few tweaks.  The most notable of the latter are some changes to the id validation.

- [x] Could you please sanity-check this + the other changes, and if you like the id validation approach, add some test cases for it?  This validation is also doing what seems like considerable work toward building the change df... I wonder if you could refactor to avoid double work.
- [x] Could you please construct a benchmark and figure out how to add it to it to a new folder in the repo, e.g. "benches" or "benchmarks"?  This probably requires adding something to the .Rbuildignore so CHECKs don't complain.  And post some results to see if we should try to optimize more or not?  Perhaps take the NHSN data set version history, filter down to a state, convert to line list, then see how quickly we can reconstruct?  (Ideally we'd have both the time taken (perhaps using `bench::mark`) and the max RAM usage.)

brookslogan:
Sorry, I realize my latest local test wasn't in sync.  I'll try to push a fix for the stuff I pushed momentarily...

JavierMtzRdz:
Thanks for all the feedback! I have updated the function based on your comments. I made some changes to the ID validation, but I'd like to confirm whether the approach is planned to account for cases where a single ID can produce more than one record. I also added a benchmark document [here](https://github.com/cmu-delphi/epiprocess/blob/linelist/benchmarks/linelist_benchmark.md) as suggested. In general, it seems to work efficiently, but the validation case is significantly slower. I'll try another approach if my interpretation of the validation is correct. 

Also, I tried to use the NHSN data as suggested, but it had some bugs when converting to a linelist format. So I used the synthetic linelist I used in the tests. Let me know if that is okay with you.

brookslogan:
Partway through re-review:
- [x] todo: Could you please add a README to the bench folder saying whether we need to manually rebuild the benchmark file, and what command to use to rebuild them?
- [x] issue: I'm getting " ! Could not automatically select column for `version_recorded`; please specify it manually. " on the benchmark examples if I exclude all the manual parameter settings, even though there don't seem to be extra column names lying around to conflict with the defaults.
- [x] todo: fix the lint I broke, probably from `count` in RHS of `dplyr::rename(change = count) %>%`

- musing: we're a bit slow and sometimes cause out-of-memory issues on something like USA nationwide hospital admissions for covid for a few years, ~100M.  It may be conceivable that users might have such large data even for more limited geographical locations if they are working with outpatient or longer-term data sets (haven't checked the math here).  And this means we probably couldn't handle e.g., the USA nationwide covid case&hosp line list.  Already the 100M row line list might be on the edge of what can be handled by data frames in RAM anyway, though.  It may be that further scaling might be tied to larger projects allowing us to refer to data on disk.

brookslogan (approved):
Looks good.  Please see todo items above, plus:
- [x] todo: add test case for `version_recorded` being recognized
- [x] issue: similarly, `version_deleted` isn't being recognized automatically, and I assume `is_deletion` as well.  We should auto-recognize those and check for mutual exclusivity.  Plus test cases.

E.g., one test case: make this work properly rather than ignoring the column:
```r
  tibble::tibble(
    state = "ca",
    date = as.Date("2022-01-01"),
    version_recorded = as.Date("2022-01-02"),
    version_deleted = as.Date("2022-01-05"),
    value = 1
  ) %>%
    linelist_to_archive()
```

- [x] todo: add a multi-state test case
- [x] todo: convert `expect_equal(ea$DT$n, c(2, 1, 1))` etc. to test all columns, e.g., via `expect_equal(ea$DT, data.table(......, key = ....))`, or probably better if it works, `expect_equal(ea, as_epi_archive(tibble(.......)))` so we check `time_type` etc.
- [x] todo: remove the rendered bench files; they look large right now and that can cause clone and even install slowdowns.  We should look into CI to auto render these into GH artifacts, or at least just keep them on a gh-pages branch.  Task for later I think.

Once complete, I believe this is ready to SQUASH merge.  (So we don't get the large rendered bench files in the main branch history.)

JavierMtzRdz:
Thanks for all of your detailed comments! I'm working on it!

PR #690, opened 2026-02-13
Join method for `epi_df`

### LINKED ISSUE #671: `epi_df` joins to data frames with richer keys yields invalid `epi_df`
- If `x` is an `epi_df` with some set of key cols, and `y` is a data frame with key cols that include all of `x`'s key cols plus some more, and we join the two by `x`'s key cols, we should be outputting an `epi_df` with `y`'s key cols.  But instead, we output an `epi_df` with only `x`'s key cols, even when this violates the one-row-per-epikeytime constraint (we don't even decay to tibble).
- `check_ukey_unique` on errors is printing a string instead of actually generating an error (thus why the below needs to `cat` its result) [check whether this is idiomatic according to checkmate/etc.; if so, make assert_ variant etc.]

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(epiprocess)
#> Loading required package: epidatasets
#> Registered S3 method overwritten by 'tsibble':
#>   method               from 
#>   as_tibble.grouped_df dplyr
#> 
#> 
#>     filter
target <- tibble(geo_value = 1, time_value = 1, target = 1) %>% as_epi_df()
sensors <- tibble(geo_value = 1, time_value = 1, sensor_name = c("A", "B"), sensor_value = 1:2)
joined <- left_join(target, sensors, by = c("geo_value", "time_value"))
joined
#> An `epi_df` object, 2 x 5 with metadata:
#> * geo_type  = hhs
#> * time_type = integer
#> * as_of     = 2025-06-03 10:24:05.131905
#> 
#> # A tibble: 2 × 5
#>   geo_value time_value target sensor_name sensor_value
#>       <dbl>      <dbl>  <dbl> <chr>              <int>
#> 1         1          1      1 A                      1
#> 2         1          1      1 B                      2
key_colnames(joined)
#> [1] "geo_value"  "time_value"
epiprocess:::check_ukey_unique(joined, key_colnames(joined)) %>% cat()
#> There cannot be more than one row with the same combination of geo_value
#> and time_value.  Problematic rows:
#> An `epi_df` object, 2 x 5 with metadata:
#> * geo_type  = hhs
#> * time_type = integer
#> * as_of     = 2025-06-03 10:24:05.131905
#> 
#> # A tibble: 2 × 5
#>   geo_value time_value target sensor_name sensor_value
#>       <dbl>      <dbl>  <dbl> <chr>              <int>
#> 1         1          1      1 A                      1
#> 2         1          1      1 B                      2
```

<sup>Created on 2025-06-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

JavierMtzRdz:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

### Join method for `epi_df`

Here's how `*_join.epi_df` performs across different scenarios:
- When keys are unique and valid, the output remains an `epi_df`:
  - Joining two `epi_df` objects with different keys correctly updates the combined `other_keys` metadata.
  - The original grouping structure of the input `epi_df` is preserved.
  - When `*_type` internal attributes do not match, it issues a warning. 
- The output converts to a standard tibble if:
  - The join results in multiple rows for the same key combination. This includes cases where two `epi_df` objects are not joined by `geo_value` or `time_value`.
  - A key column is missing from the result. A warning is issued.
  - The join introduces `NA` values into key columns. A warning is issued suggesting the use of `inner_join`.
  - `cross_join()` is used.

I also added an internal `force_meta` to easily adjust `geo_type` and `time_type` when using a pipe, as shown in the example below. 
 
```r
week <- tibble::tibble(geo_value = "ca", time_value = as.Date("2020-01-01"), val2 = 2) %>%
    as_epi_df() %>%
    epiprocess:::force_meta(time_type = "week")
```

- Partially fixes #671


R/methods-epi_df.R
```diff
@@ -548,3 +548,112 @@ sum_groups_epi_df <- function(.x, sum_cols, group_cols = "time_value") {
...
+  decay_epi_df(NextMethod())
+}
+
+# Helper to merge keys and validate result
+merge_epi_df_join <- function(res, x, y) {
+  meta <- attr(x, "metadata")
+
+  # NA checks for essential keys
+  has_na_geo <- anyMissing(res$geo_value)
+  has_na_time <- anyMissing(res$time_value)
```
brookslogan: todo: anyMissing is sort of deprecated; replace with vctrs::vec_any_missing or anyNA; vctrs one is a bit more general, which might make it easier to generalize the types we accept for geo & time values later.

JavierMtzRdz:
Resolved conflicts and replaced `anyMissing`. Waiting for tests to complete.

brookslogan (approved):
I think I left this PR hanging thinking it contained the `bind_{rows,cols}` that we should probably get over, but those are over in the `binds` branch.  So this should be good to go.

PR #692, opened 2026-02-28
`epi_df` behavior for `drop_na` and `pivot_*`

JavierMtzRdz:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
- [x] Styling and documentation checks. Make a PR comment with:
  process.

### `epi_df` behavior for `drop_na` and `pivot_*` verbs

This PR includes methods for `drop_na` and `pivot_*` verbs.

#### `drop_na`
 Always returns an epi_df, as it only subsets rows and does not remove essential key columns.

#### `pivot_wider`
`pivot_wider` method preserves the `epi_df` class as long as the `geo_value` and `time_value` columns are present in the output. It also automatically updates the `other_keys` metadata by removing any keys that were pivoted away or dropped during the transformation. However, the object will silently decay into a `tibble` if `geo_value` or `time_value` are used in `names_from`, as this removes them from the column set.

#### `pivot_longer`
The `pivot_longer` method automatically promotes the new `names_to` column to a key in the `other_keys` metadata. If the resulting data is not unique on its keys even after this promotion, the method will issue a warning and the object will decay into a `tibble`. Similar to `pivot_wider`, the result will silently become a `tibble` if the essential `geo_value` or `time_value` columns are pivoted away.

- Partially resolves issue #391.


R/methods-epi_df.R
```diff
@@ -548,3 +548,43 @@ sum_groups_epi_df <- function(.x, sum_cols, group_cols = "time_value") {
   ) %>%
     arrange_canonical()
 }
+
+#' @method drop_na epi_df
+#' @importFrom tidyr drop_na
+#' @export
+drop_na.epi_df <- function(data, ...) {
+  res <- NextMethod()
+  reconstruct_light_edf(res, data)
```
brookslogan:
```suggestion
  reclass(res, attr(data, "metadata"))
```
for performance.  `reconstruct_light_edf` is "light" because it does not perform expensive grouped_df reconstruction checks if we're grouped, but it's still not **that** light (maybe it should be called "medium" or "shallow"): it's still performing checks of many `epi_df` invariants.  I think we're guaranteed that `drop_na` won't do anything of concern.

R/methods-epi_df.R
```diff
@@ -548,3 +548,43 @@ sum_groups_epi_df <- function(.x, sum_cols, group_cols = "time_value") {
...
+}
+
+#' @method pivot_wider epi_df
+#' @importFrom tidyr pivot_wider
+#' @export
+pivot_wider.epi_df <- function(data, ...) {
+  res <- NextMethod()
+  reconstruct_light_edf(res, data)
```
brookslogan:
```suggestion
  res <- NextMethod(names_from = {{names_from}})
  names_from_chr <- names(tidyselect::eval_select(rlang::enquo(names_from), data, allow_rename = FALSE))
  print(names_from_chr)
  template <- vctrs::vec_ptype(data)
  attr(template, "metadata")$other_keys <- vctrs::vec_set_difference(attr(template, "metadata")$other_keys, names_from_chr)
  reconstruct_light_edf(res, template)
```

In order to appropriately drop `"signal"` from `other_keys` in this situation:
```r
tibble(geo_value = 1, time_value = 1, signal = c("cases", "signal"), value = c(5, 2)) %>%
  as_epi_df(other_keys = "signal") %>%
  pivot_wider(names_from = signal)
```

R/methods-epi_df.R
```diff
@@ -548,3 +548,43 @@ sum_groups_epi_df <- function(.x, sum_cols, group_cols = "time_value") {
...
+}
+
+#' @method pivot_longer epi_df
+#' @importFrom tidyr pivot_longer
+#' @export
+pivot_longer.epi_df <- function(data, ..., names_to = "name") {
+  res <- NextMethod()
+  res <- reconstruct_light_edf(res, data)
+  if (inherits(res, "epi_df")) {
+    new_keys <- setdiff(names_to, ".value")
```
brookslogan: praise: nice diligence checking this case!  Definitely not an option I remember off the top of my head...

R/methods-epi_df.R
brookslogan: actually, reconstruct_light_edf should be doing this medium check but it's not!  major bug

R/methods-epi_df.R
```diff
@@ -548,3 +548,43 @@ sum_groups_epi_df <- function(.x, sum_cols, group_cols = "time_value") {
...
+    attr(res, "metadata")$other_keys <- unique(
+      c(attr(res, "metadata")$other_keys, new_keys)
+    )
+    # Check if adding the new keys solved the uniqueness problem
```
brookslogan: suggestion: remove this if branch & just return `reconstruct_light_edf(res, template)`, where `template` incorporates the `new_keys`.  Maybe something like
```r
template <- vec_ptype(data)
template[new_keys] <- rep(list(character(), length(new_keys))
attr(template, "metadata")$other_keys <-  c(attr(res, "metadata")$other_keys, new_keys)
reconstruct_light_edf(res, template)
```
This currently won't decay if we have ukey uniqueness issues, but as far as I can tell, we only have those if the input already had them, so that seems fine.

brookslogan (changes requested):

R/methods-epi_df.R
brookslogan: ... or "not".  I guess we were planning to make it check epikeytime uniqueness, but maybe wanted to make this change at the same time as ensuring it in operations not using reconstruct_light_edf.

JavierMtzRdz: I've incorporated the suggested change, and I'll test out what it would look like if we made the `reconstruct_light_edf` check for ukey uniqueness.

R/methods-epi_df.R
JavierMtzRdz: Nice catch! I missed this case!

R/methods-epi_df.R
JavierMtzRdz: I was thinking about the situation where the input df doesn't satisfy key uniqueness, but probably this isn't the right place to flag that if the input is already problematic.

JavierMtzRdz (commented):
I have included the suggestions from your review! No relevant comments about it this time.

brookslogan (approved):

PR #693, opened 2026-03-03
Improve `autoplot` for when dealing with many unique keys + heatmap

JavierMtzRdz:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

### Improve `autoplot` for when dealing with many unique keys

This PR improves the usability and performance of `autoplot()` when handling many distinct key combinations. It introduces automatic subsampling for static plots and an interactive dropdown interface for easier exploration.

Static plots for `epi_df` with many keys now automatically subsample to 10 unique keys by default. This prevents the plot from becoming unreadable. When subsampling occurs, a warning is issued with suggestions for `.max_keys = Inf`, `interactive = TRUE`, or `.facet_filter`.

Added the `interactive` parameter to support plotly plots. In this scenario, faceted `epi_df` and `epi_archive` plots use dropdown menus.

I also included the heatmap function we previously discussed. 

Some examples below.
[reprex_pr_explanation.html](https://github.com/user-attachments/files/25749849/reprex_pr_explanation.html)

- Resolves #{issue number}


JavierMtzRdz:
Before reviewing this PR, it would be helpful to review #694. That PR automated pivoting for long-format data with `as_epi_*` and improved response selection in `autoplot`. I directed #694 to this branch to avoid issues when merging into dev.

R/autoplot.R
```diff
@@ -73,11 +79,15 @@ autoplot.epi_df <- function(
   .facet_by = c(".response", "other_keys", "all_keys", "geo_value", "all", "none"),
```
brookslogan, line 98: issue: while turning faceting into dropdowns in the interactive plot is nice for maximizing real estate for each plot & avoids issues with spacebar scrolling an inconvenient amount when using tall facet rows, I think the annoyance of having to interact with a long dropdown menu will still make us want to favor regular facet plots in the interactive version by default / explicit choice.  The facet version also lets us more quickly identify "interesting locations" (even if we had some sort of keyboard shortcut to quickly move between locations if using a dropdown).
suggestion:

add a .dropdown_menu_by (or better-named argument) to control what the dropdown menu selects between
don't turn faceting into dropdown menu by default
check that facet labels also have a some sort of prefix like "Key: " (except we need to tweak that name as it's inconsistent throughout the package), so that we can search "Key: al" etc. to jump to a particular facet without false positives.

R/autoplot.R
```diff
@@ -218,6 +256,181 @@ autoplot_check_viable_response_vars <- function(
...
+
+
+autoplot_plotly_dropdown <- function(
+  data, group_col, trace_col = NULL,
+  color_map = NULL, .base_color = "#3A448F",
+  xaxis_title = "Date", yaxis_title = "", legend_title = ""
+) {
+  rlang::check_installed("plotly")
+
+  unique_groups <- levels(droplevels(as.factor(data[[group_col]])))
```
brookslogan: question: can we remove the `droplevels` here & below without things going haywire?  The motivation is that, if the user encoded this as a factor, then they expect to explicitly see "nothing" for unused levels, rather than not seeing them at all.

R/autoplot.R
```diff
@@ -218,6 +256,181 @@ autoplot_check_viable_response_vars <- function(
...
+
+  # Prepare lines to be added to the plot
+  trace_specs <- list()
+  trace_group_idx <- integer(0)
+
+  # Identify each dropdown option
+  for (gi in seq_along(unique_groups)) {
+    g <- unique_groups[gi]
+    g_data <- data[data[[group_col]] == g, , drop = FALSE]
```
brookslogan: thought: this might be slow (quadratic in number of unique groups); we could potentially use a `group_map`

todo: remove `drop = FALSE` here & elsewhere and just ensure we're backed by tibbles / data.tables instead.  (Rest of package is written this way; don't want to mix and match.)

R/autoplot.R
```diff
@@ -218,6 +256,181 @@ autoplot_check_viable_response_vars <- function(
 }
 
 
+autoplot_subsample_keys <- function(
+  object, geo_and_other_keys, .max_keys, interactive, .facet_used
+) {
+  if (interactive || is.infinite(.max_keys)) {
+    return(object)
```
brookslogan, line 324: praise: limiting number of lines & facets in plain plots, and number of lines shown at start in interactive plots, avoids a lot of unreadable plotting situations!

suggestion: add a message similar to the one in the code below for when the number of lines shown at the start is limited.  Include a pointer about being able to double-click a legend entry twice to switch to showing all keys (and/or the .max_keys setting approach).

tests/testthat/test-autoplot.R
```diff
@@ -82,3 +82,90 @@ test_that("autoplot warns when some specified columns are not numeric, and lists
...
+    mutate(cases = rnorm(dplyr::n())) %>%
+    as_epi_df()
+
+  # Hint should appear if .facet_used = TRUE
+  expect_warning(
+    epiprocess:::autoplot_subsample_keys(
+      df, "geo_value",
+      .max_keys = 10, interactive = FALSE, .facet_used = TRUE
+    ),
+    regexp = "To plot specific keys, use `autoplot\\(..., .facet_filter = ...\\)`"
```
brookslogan: question: do you prefer this regexp/grepl-ing approach, or pairing together an `expect_{warning,error}(..., class = ...) with an `expect_snapshot([error = true])`?

brookslogan (commented):
For `autoplot.epi_df`: very nice!   I have a feature request regarding allowing choice/combination of faceting and dropdown selection, plus some minor suggestions/thoughts; see inline comments.

For `autoplot.epi_archive` and `plot_heatmap`: I still need to take a look.  Going to re-request review to keep this on my stack.

R/autoplot.R
JavierMtzRdz: Instead of `.dropdown_menu_by`, I used `.facet_to_dropdown`. When TRUE, it converts the facets into dropdown menus. I chose this approach because a simple implementation of `.dropdown_menu_by` might cause `dropdown` selections that leave facet boxes empty (without removing the facet boxes when they don't match the dropdown group selected). To ensure `.dropdown_menu_by` functions properly, we might need to create a subplot for each unique facet value, which could take additional time to implement. 

Regarding the prefix, I added an `autoplot_get_prefix` function that adds the key based on the `.facet_by`. Let me know if it matches your idea.

R/autoplot.R
JavierMtzRdz: I replaced this with `group_split()`+ `purrr::map()`, and removed `drop = FALSE`. I'll keep a note of avoiding `drop = FALSE`.

tests/testthat/test-autoplot.R
JavierMtzRdz: I recently learned about snapshot tests, but overall, I prefer `expect_*` + `expect_snapshot` when working with `cli` messages/warnings. However, when I'm implementing a feature, I first use `regexp` since some parts of the message can still change a bit, and it is faster than updating the snapshots. Turns out I sometimes forget to replace it x.x 

Would you like me to update the `regexp` above?

R/autoplot.R
```diff
@@ -327,16 +548,24 @@ autoplot.epi_archive <- function(object, ...,
...
+      snapshots, .base_color, inherits(.versions, "Date")
+    ))
+  }
+
+
   bp <- bp +
     ggplot2::geom_line(
       data = snapshots,
-      mapping = ggplot2::aes(y = .response, color = version, group = factor(version))
+      mapping = ggplot2::aes(y = .response, color = version, group = interaction(!!!all_avail, version))
```
brookslogan: suggestion: also use a `sep = ` setting here like with `autoplot.epi_df`?  Might also consider standardizing to a semicolon(?) rather than a slash but doesn't matter too much.

R/autoplot.R
```diff
@@ -354,9 +583,42 @@ autoplot.epi_archive <- function(object, ...,
   }
   # make the finalized layer last
   bp$layers <- rev(bp$layers)
+
   bp
 }
 
+autoplot_interactive_archive <- function(snapshots, .base_color, .versions_are_dates) {
```
brookslogan: question: why do we need `.versions_are_dates`?  I put in one example and it seemed like we might be able to just check `inherits(snapshots$version, "Date")`.

brookslogan (commented):
`autoplot.epi_archive` notes.  Still need to look at heatmap

R/autoplot.R
JavierMtzRdz: Done! Added a message to `autoplot_interactive_df` for those cases. Available [here](https://github.com/cmu-delphi/epiprocess/pull/693/changes#diff-3743e0b802a90dce0e37b63dd70b1e49071f833caeeb9e4e718213f15e1efc91R462).

R/autoplot.R
JavierMtzRdz: They should not interfere with the overall behaviour of the package function. I only used it to avoid generating empty extra plots. Do you want me to remove them?

R/extra_plots.R
```diff
@@ -0,0 +1,64 @@
+#' Plot a heatmap for an epi_df
+#'
+#' @inheritParams autoplot-epi
+#' @param x An `epi_df` object.
+#' @param fill <[`tidy-select`][dplyr::dplyr_tidy_select]> The column to use for the fill aesthetic.
+#'   If NULL, the first numeric non-key column is used.
+#' @param ... Additional arguments passed to [ggplot2::geom_tile()].
+#'
```
brookslogan: todo: add example

R/autoplot.R
JavierMtzRdz: In this case, it should not affect the final output, as the group is not used in the legends. Still, I've updated this line and the others to use semicolons.

R/autoplot.R
JavierMtzRdz: I was likely attempting to reuse another object and change the approach. I've taken it out.

R/extra_plots.R
```diff
@@ -0,0 +1,64 @@
...
+    )
+  }
+
+  fill_quo <- rlang::enquo(fill)
+  key_cols <- key_colnames(x)
+  non_key_cols <- setdiff(names(x), key_cols)
+
+  # --- Reuse autoplot utility for variable selection
+  vars <- autoplot_check_viable_response_vars(x, !!fill_quo, non_key_cols = non_key_cols)
+  fill_name <- names(vars)[1]
```
brookslogan: todo: don't auto-trim down to 1.  Either hard error if >1, or actually use them all (faceting?).  If the latter, it would be nice to also change the default here to plotting all acceptable variables (same applies to autoplot... but this is a separate issue and I'm not sure if you're already addressing it in a separate PR).

R/extra_plots.R
```diff
@@ -0,0 +1,64 @@
...
+      .y_axis = interaction(!!!rlang::syms(geo_and_other_keys), sep = " / ")
+    )
+
+  # Create plot
+  p <- ggplot2::ggplot(plot_df, ggplot2::aes(
+    x = time_value,
+    y = .y_axis,
+    fill = !!fill_quo
+  )) +
+    ggplot2::geom_tile(color = "white", linewidth = 0.1, ...) +
```
brookslogan: suggestion: consider thinner or no lines? They're a bit distracting with `plot_heatmap(cases_deaths_subset, case_rate_7d_av)`, and I'd imagine that if you had a handful of years that it might make it completely illegible.  They do seem potentially nice with a smaller number of time values, though...

tests/testthat/test-autoplot.R
```diff
@@ -82,3 +90,114 @@ test_that("autoplot warns when some specified columns are not numeric, and lists
...
+  df <- expand.grid(
+    geo_value = c("ak", "al"),
+    time_value = as.Date("2023-01-01") + 0:10
+  ) %>%
+    mutate(cases = 1:22) %>%
+    as_epi_df()
+
+  # Simple epi_df dropdowns
+  p1 <- autoplot(df, cases, .interactive = TRUE, .facet_by = "geo_value", .facet_to_dropdown = TRUE)
+  expect_snapshot(plotly::plotly_build(p1)$x$layout$updatemenus)
```
brookslogan: todo: please test out whether `testthat::expect_snapshot_file` or [this function](https://github.com/ryantibs/quantgen/blob/59718548fd6b327b5f1870f9c86fb079cc500b70/quantgen/tests/testthat/test-characterization-quantile_ensemble.R#L64) work here instead, so we can keep the text snap file more legible.  Or comment here what we should be looking for in the text snap file.

tests/testthat/test-autoplot.R
```diff
@@ -82,3 +90,114 @@ test_that("autoplot warns when some specified columns are not numeric, and lists
...
+    as_epi_df(other_keys = "other", as_of = as.Date("2023-01-04"))
+
+  p2 <- autoplot(
+    df_other, cases,
+    .facet_by = "other_keys", .facet_to_dropdown = TRUE,
+    .color_by = "none", .interactive = TRUE
+  )
+  pb2 <- plotly::plotly_build(p2)
+  expect_snapshot(pb2$x$layout$updatemenus)
+  expect_snapshot(purrr::map(pb2$x$data, ~ .x$visible))
```
brookslogan:
```suggestion
  expect_snapshot(purrr::map_lgl(pb2$x$data, ~ .x$visible))
```

brookslogan: or the map_chr %||% "TRUE" approach from below

tests/testthat/test-extra_plots.R
```diff
@@ -0,0 +1,52 @@
...
+  age = rep(rep(c("adult", "child"), each = 5), 2),
+  time_value = rep(test_date + 1:5, 4),
+  val = runif(20, 0, 15)
+) %>% as_epi_df(other_keys = "age")
+
+test_df_many <- dplyr::tibble(
+  geo_value = rep(as.character(1:100), each = 2),
+  time_value = rep(test_date + 1:2, 100),
+  val = runif(200, 0, 15)
+) %>% as_epi_df()
```
brookslogan, line 21: suggestion: wrap in
```r
withr::with_rng_version("3.5.0", withr::with_seed(1410852, {
  ...
}))
```

(1410852 is a fixed random seed here.  To make re-running and any more liberal usage of snapshotting not produce confusion.)

brookslogan (approved):
Looking good!  Have some minor things inline, and one larger task:

- [ ] todo: add interactive and heatmap examples to an existing or new vignette

Happy to look over the vignette (or other) changes again if you re-request review, but you can also just go ahead and just merge after addressing remaining notes.  Are we able to merge this on its own and then re-open any dependent PRs that are auto-closed?

JavierMtzRdz:
Thanks! I'll work on the notes. We can merge after that if necessary, but I'd prefer to wait until #694 is ready since that PR is directed to this branch, and merge squashing sometimes breaks the connection between the intermediate branch if we merge before the new one. However, we can merge if needed, and I can resolve the conflicts later. #694 includes the automated pivoting for long-format data with as_epi_* and improved response selection in autoplot.

R/extra_plots.R
JavierMtzRdz: Done!

R/extra_plots.R
JavierMtzRdz: Done, it now adjusts the linewidth based on the number of different dates. It's worth noting that even with `linewidth = 0`, it still maintains a thin dividing line.

tests/testthat/test-extra_plots.R
JavierMtzRdz: Done!

tests/testthat/test-autoplot.R
JavierMtzRdz: `map_lgl` is clearer here since the x$visible only returns booleans.. Below, it can return "legendonly"

JavierMtzRdz (commented):
@brookslogan I have solved all your previous comments, but there is an issue with `autoplot.epi_workflow` discussed [here](https://github.com/cmu-delphi/epiprocess/pull/693#discussion_r2957064748) that may need your attention.

R/autoplot.R
```diff
@@ -73,23 +95,20 @@ autoplot.epi_df <- function(
   .facet_by = c(".response", "other_keys", "all_keys", "geo_value", "all", "none"),
   .base_color = "#3A448F",
   .facet_filter = NULL,
-  .max_facets = deprecated()
```
JavierMtzRdz, line 76: I noticed that we removed `.max_facets` here, but `autoplot.epi_df` is used [here](https://github.com/cmu-delphi/epipredict/blob/5f5c470e43eda9c877f83d8f6df72ed49bedad0b/R/autoplot.R#L151) and references `.max_facets`. This appears to cause issues in `autoplot.epi_workflow`. Would you like me to (1) restore `.max_facets` here, (2) adjust `autoplot.epi_workflow`, or (3) perhaps rename `.max_keys` to `.max_facets`?

R/extra_plots.R
JavierMtzRdz: `autoplot_check_viable_response_vars` warns when no variable is provided and selects one, and #694 prioritizes choosing the `value` column when no column is provided. Therefore, I allowed facets when multiple variables are provided here.

tests/testthat/test-autoplot.R
JavierMtzRdz: I used `expect_snapshot_file` because it can also generate a separate file with the output. I saved it in JSON format to keep it readable, but this keeps the snapshot file legible. `expect_equal_custom_rds_snapshot` could be very useful for larger files.

R/autoplot.R
brookslogan: I think so?  By encoding as factor, the user is sort of asking to see the empty plots.

R/autoplot.R
brookslogan: Nice catch!
- I was tempted by (3), but I guess it may not actually be accurate; we're using `.max_keys` now to also control the number of lines in a non-faceted plot, right?
- So maybe (1), ~~but also we probably should rename `.max_keys` -> `.max_epikeys`, or `.max_geodemo_groups`, or `.max_lines`... When developing I like to use the term "epikeys", but it's not actually standard in the field, and we don't have documentation explaining it.  So maybe... `.max_lines`?  Other ideas?~~

brookslogan: Actually, forget the renaming part.  We already use "key" terminology here in options for other parameters (but in a way inconsistent with `key_colnames`... something to deal with at a different time.)  And `.max_lines` could be confused with "max lines per facet".

R/extra_plots.R
brookslogan: I love the faceting idea, and think we should probably be making `autoplot_check_viable_response_vars` default selection be ALL the numeric variables if there is no `value` column.  (Both for this and the autoplots.)

One issue with the current faceting, though: `free_y` doesn't actually work for heatmaps :'(.

suggestion: normalize each epikey x signal.  That both resolves this facet scaling issue, and issues with legibility when plotting a count signal across epikeys with very different populations.

(There's also `patchwork` or `ggExtra` or `cowplot` or manually messing with grobs to combine plots with their own color guides, rather than go through facet_*.  That might be better when dealing with 2 rate signals on different scales.  But that still doesn't address the "count" situation.)

tests/testthat/test-autoplot.R
brookslogan: Makes sense!  I forgot why I implemented the custom rds snapshot; I think maybe it was because I was dealing with testing some larger objects or I wanted to be able to just readRDS to see the reference or something.  Or a testthat version/edition issue.

R/autoplot.R
brookslogan: That totally makes sense for quick implementation.

Regarding the prefix, yes, that's the general idea I was thinking of.  But I don't like the feel of "Keys: " or "Other Keys: " or "All: ".
- Perhaps the prefix could just be something like "Facet: " or "For: " regardless of the `type`?
- Or:
  - just live with "Keys:"
  - "Other Keys:" -> paste(collapse = " / ", other_keys),
  - "All:" -> "Keys x Indicator" (I refuse to call it "response"...)

R/autoplot.R
JavierMtzRdz: Done!

R/autoplot.R
JavierMtzRdz: Done!

R/autoplot.R
JavierMtzRdz: Done!

tests/testthat/test-autoplot.R
JavierMtzRdz: Actually, I used `expect_equal_custom_rds_snapshot` to avoid adding jsonlite to suggests.

R/extra_plots.R
JavierMtzRdz: Done! I added this approach to #694.

PR #694, opened 2026-03-06
Automated pivoting for long-format data with `as_epi_*` and enhanced response selection in `autoplot`

### LINKED ISSUE #678: Make `autoplot.epi_archive` behave better when multiple value columns exist
Because plots automatically pop up, the warning about variable autoselection can be easily be overlooked.

Consider one of these options:
- Adding an axis or facet label with the selected variable name
- Plotting all value columns by default (using faceting) (#667 would need fixed first)
- Refusing to automatically select a variable if there are multiple candidates (as in current `revision_summary()`)

JavierMtzRdz:
### Checklist

Please:

      PR).
      brookslogan, nmdefries.
      number (the second number).
  process.

### Automated pivoting for long-format data with `as_epi_*` and enhanced response selection in `autoplot`

This PR introduces automated pivoting for long-format data and refines the column selection logic in `autoplot`.

Added `input_format` (defaulting to "auto") and `signal_var` parameters to enable both automatic detection and explicit configuration. This enhancement allows `as_epi_df` and `as_epi_archive` to identify when data is in "long" format (containing signal indicators) and convert it to "wide" format by default. Note that the pivot only works when a `value` column exists. Increasing flexibility would require adding another parameter to `as_epi_*`.

For `autoplot`, it now prefers the value column when multiple numeric columns are available.

- Partially fixes #678


R/archive.R
```diff
@@ -530,15 +538,24 @@ as_epi_archive <- function(
...
   assert_data_frame(x)
   x <- rename(x, ...)
   x <- guess_column_name(x, "time_value", time_column_names())
   x <- guess_column_name(x, "geo_value", geo_column_names())
   x <- guess_column_name(x, "version", version_column_names())
 
+  x <- pivot_epi_data(x, input_format, signal_var,
+    id_cols = c("geo_value", other_keys, "time_value", "version"),
+    caller_name = "as_epi_archive"
+  )
```
brookslogan: issue: this isn't valid due to the archive format
```r
tibble(geo_value = 1, time_value = 1, version = c(1,1,2), signal = c("a", "b", "a"), value = c(1, 11, 2)) %>%
  as_epi_archive()
#> as_epi_archive: pivoting long to wide based on `signal` column.
#> → An `epi_archive` object, with metadata:
#> ℹ Min/max time values: 1 / 1
#> ℹ First/last version with update: 1 / 2
#> ℹ Versions end: 2
#> ℹ A preview of the table (2 rows x 5 columns):
#> Key: <geo_value, time_value, version>
#>    geo_value time_value version     a     b
#>        <num>      <num>   <num> <num> <num>
#> 1:         1          1       1     1    11
#> 2:         1          1       2     2    NA
```
that `NA` should be `11`.

suggestion: instead, {vec,group}_split/nest/group_modify to turn each signal into an archive separately, then {R,r}educe with `epix_merge`.  Or just abort & tell the user to do that.

R/autoplot.R
```diff
@@ -226,28 +226,48 @@ autoplot_check_viable_response_vars <- function(
...
   vars <- tidyselect::eval_select(rlang::expr(c(...)), object)
   if (rlang::is_empty(vars)) { # find them automatically if unspecified
-    vars <- tidyselect::eval_select(names(allowed)[1], object)
-    cli::cli_warn(
-      "Plot variable was unspecified. Automatically selecting {.var {names(allowed)[1]}}.",
-      class = "epiprocess__unspecified_plot_var",
-      call = call
-    )
+    if (length(allowed) == 1L) {
+      vars <- tidyselect::eval_select(names(allowed)[1], object)
```
brookslogan: musing: sorry, I see you've addressed some things here that I pointed out in the base PR, and maybe induced some merge conflicts.

todo: add `allow_rename = FALSE` here and for the other `eval_select`s.  (Or allow + handle them, changing plot labels accordingly.  Perhaps by passing through `rename`.  I suspect that this would not be worth the effort, though.)

R/autoplot.R
```diff
@@ -226,28 +226,48 @@ autoplot_check_viable_response_vars <- function(
...
+      cli::cli_warn(
+        "Plot variable was unspecified. Automatically selecting {.var value}.",
+        class = "epiprocess__unspecified_plot_var",
+        call = call
+      )
+    } else {
+      cli::cli_abort(
+        c(
+          "Multiple candidate plot columns: {.var {names(allowed)}}.",
+          i = "Specify the column(s) to plot, e.g. `autoplot(x, {names(allowed)[1]})`."
```
brookslogan: praise: love this feedback & how it provides an example working command!

suggestion: change "i" to ">" when it's suggesting a user action

R/autoplot.R
```diff
@@ -226,28 +226,48 @@ autoplot_check_viable_response_vars <- function(
...
-        call = call
+        call = call,
+        qty = length(vars)
       )
     } else if (!all(ok)) {
       cli::cli_warn(
         c(
-          "Only the requested variables {.var {names(vars)[ok]}} are numeric.",
+          "{?The/Only the} requested variable{?s} {.var {names(vars)[ok]}} {?is/are} numeric.",
           i = "`autoplot()` cannot display {.var {names(vars)[!ok]}}."
```
brookslogan:
```suggestion
          "Only the requested variable{?s} {.var {names(vars)[ok]}} {?is/are} numeric.",
i = "`autoplot()` cannot display {.var {names(vars)[!ok]}}."
```
or
```suggestion
          "`autoplot()` cannot display {.var {names(vars)[!ok]}}, as they are not numeric",
i = "Only plotting {.var {names(vars)[ok]}}."
```

brookslogan: suggestion: perhaps change vars to be names(rlang::eval_select(......)) so we don't have all these names() calls over the place.  That also means in some place it can just be vars <- allowed[[1L]] instead of having to call eval_select at all.

R/epi_df.R
```diff
@@ -315,20 +330,30 @@ as_epi_df.grouped_df <- function(x, ...) {
 #' @order 1
 #' @method as_epi_df data.frame
 #' @export
-as_epi_df.data.frame <- function(x, as_of, other_keys = character(), ...) {
-  as_epi_df(x = tibble::as_tibble(x), as_of = as_of, other_keys = other_keys, ...)
+as_epi_df.data.frame <- function(x, as_of, other_keys = character(),
+                                 input_format = c("auto", "wide", "long"),
+                                 signal_var = NULL, ...) {
```
brookslogan, line 342: note: you can see we're violating the dot-prefix convention here.  I think some idea regarding limiting the renaming ability is already filed in an issue though.  Not something to tackle in this PR.

R/utils.R
```diff
@@ -1109,3 +1117,54 @@ check_ukey_unique <- function(x, ukey_names, end_cli_message = character()) {
...
+#' @importFrom tidyselect any_of
+#' @importFrom dplyr all_of
+#' @noRd
+pivot_epi_data <- function(x, input_format, signal_var, value_var = "value",
+                           id_cols = NULL, caller_name, call = rlang::caller_env()) {
+  if (input_format == "auto") {
+    candidates <- intersect(names(x), signal_column_names())
+    if (length(candidates) > 0) {
+      input_format <- "long"
+      signal_var <- signal_var %||% candidates[1]
```
brookslogan: issue: providing a non-NULL `signal_var` isn't enough to trigger "long" conversion, and yet it takes precedence over the "default candidates" which DO trigger it.
suggestion: make providing `signal_var` enough to trigger auto -> long; maybe by doing `candidates <- signal_var %||% candidates` above the if statement 

suggestion: abort if multiple candidates rather than using `candidates[1]`.

R/utils.R
```diff
@@ -1109,3 +1117,54 @@ check_ukey_unique <- function(x, ukey_names, end_cli_message = character()) {
...
+    if (!(signal_var %in% names(x))) {
+      cli::cli_abort("Column {.var {signal_var}} not found in `x`.", call = call)
+    }
+    if (!(value_var %in% names(x))) {
+      cli::cli_abort("Pivoting long to wide requires a {.var {value_var}} column.", call = call)
+    }
+    cli::cli_inform("{caller_name}: pivoting long to wide based on {.var {signal_var}} column.")
+    x <- tidyr::pivot_wider(
+      x,
+      id_cols = tidyselect::any_of(id_cols),
```
brookslogan: question: id_cols should be the geo x other x time...  could something weird happen if we let `NULL` get through?  Perhaps we need this to be required arg, and use `all_of` just for extra sanity checks?

brookslogan (changes requested):
Nice messaging!  Haven't looked at tests yet.  Some logic needs fixed (e.g., archive case).  (See above and below; I commented in two batches.)

R/epi_df.R
```diff
@@ -239,11 +245,20 @@ as_epi_df.tbl_df <- function(
   time_type = deprecated(),
   as_of,
   other_keys = character(),
+  input_format = c("auto", "wide", "long"),
```
brookslogan: musing: as I started actually typing in commands testing this feature out, I somehow started expecting `input_format = "long"` to be outputting something long, but with `"signal_name"`/etc. in the `other_keys`, rather than pivoting to wide.

Maybe `input_format` could be renamed to something better?  Maybe `from_format`?  That better implies that it may undergo significant transformation... but then it also suggests such transformation when the input format is already wide.  Other ideas?  Or just keep it at `input_format`?

R/archive.R
JavierMtzRdz: I created a `epix_pivot_wider` function using `map2` and `epix_merge` and added a test for this case.

R/autoplot.R
JavierMtzRdz: Good catch! I used `allow_rename = FALSE`.

R/autoplot.R
JavierMtzRdz: Done!

R/autoplot.R
JavierMtzRdz: Done!

R/epi_df.R
JavierMtzRdz: Noted!

R/epi_df.R
JavierMtzRdz: I adjusted the approach slightly by calling the parameter `signal_format`.
* When set to "auto," it switches to wide mode if `value` and `signal` columns are available and `signal` contains more than one element. 
* Setting it to "long" assigns `signal` to `other_keys` and issues a warning if the `signal` column does not exist. 
* When set to "wide," it pivots to wide.

R/utils.R
JavierMtzRdz: Sounds good!

R/utils.R
JavierMtzRdz: Done!

JavierMtzRdz (commented):
I have addressed all issues and suggestions in #693 and here. The most relevant comment from that PR was about [normalizing indicators](https://github.com/cmu-delphi/epiprocess/pull/693#discussion_r2967670885) when multiple are provided in `plot_heatmap`, and this was addressed here. Once the changes in this branch are approved, I will merge both into `dev`. 

In this PR, the most relevant comments were the invalid pivot [here](https://github.com/cmu-delphi/epiprocess/pull/694/changes/BASE..41ebc10f2a86f9c170c6e71a946c95ea0aeb8bf2#r2941691379) and the appropriate behaviour of input format [here](https://github.com/cmu-delphi/epiprocess/pull/694/changes/BASE..41ebc10f2a86f9c170c6e71a946c95ea0aeb8bf2#r2944184024). 

To perform the pivot for archives, I created a specific function using splitting and epix_merge [here](https://github.com/cmu-delphi/epiprocess/pull/694/changes#diff-58698176a34b0510ed60d4acddb1b589aef82df5964c66035c84628a6fc8c8bfR882). 

Regarding the `from_format` parameter, I adjusted the approach slightly by renaming it to `signal_format`.
* When set to "auto," it switches to wide mode if `value` and `signal` columns are available and `signal` contains more than one element. 
* Setting it to "long" assigns `signal` to `other_keys` and issues a warning if the `signal` column does not exist. 
* When set to "wide," it pivots to wide. 

I created specific functions to handle the process for [epi_df](https://github.com/cmu-delphi/epiprocess/pull/694/changes#diff-bfd9e8598d567a0eedbae6e9f9ae0962f1fe914f95fdc14b0f0ea2dd0ebbcc97R374) and [epi_archive](https://github.com/cmu-delphi/epiprocess/pull/694/changes#diff-58698176a34b0510ed60d4acddb1b589aef82df5964c66035c84628a6fc8c8bfR840). The auto selection logic and validation are carried out [here](https://github.com/cmu-delphi/epiprocess/pull/694/changes#diff-60bef5af43a1bd541299ce9bceeb226afb23deaa5e637b145c85e05ffd552b0cR782). 

In addition, I included some examples (rendered and as text) below to make testing easier.
[functionality_overview.html](https://github.com/user-attachments/files/26318411/functionality_overview.html)

```r
devtools::load_all(".")
library(dplyr)
library(epidatr)
library(ggplot2)

# Multi-signal epi_df, autopivot
df <- pub_covidcast(
  source = "hhs",
  signals = c(
    "confirmed_admissions_covid_1d",
    "confirmed_admissions_influenza_1d"
  ),
  geo_type = "state",
  time_type = "day",
  geo_values = "*",
  time_values = epirange(20220901, 20230101)
) %>%
  mutate(
    cat = stringr::str_sub(geo_value, 1, 1),
    value_noise = value + rnorm(n(), sd = value / 5),
    .by = geo_type
  ) |>
  as_epi_df(other_keys = "cat")

# Enhanced autoplot() for epi_df ---

# Static plots with multiple variables and advanced faceting
autoplot(df)
autoplot(df, confirmed_admissions_covid_1d)
autoplot(df, confirmed_admissions_covid_1d, confirmed_admissions_influenza_1d)
autoplot(df, confirmed_admissions_covid_1d, .facet_by = "all") # Facet by all keys + signals
autoplot(df, confirmed_admissions_covid_1d, .facet_by = "geo_value") # Facet by state only

# Interactive plots with plotly
autoplot(df, confirmed_admissions_covid_1d, .interactive = TRUE)

# Dropdown menus for cleaner exploration of many keys
autoplot(df, confirmed_admissions_covid_1d,
  .interactive = TRUE,
  .facet_by = "geo_value",
  .facet_to_dropdown = TRUE
)

# Enhanced plot_heatmap() ---

# Automatically standardizes
plot_heatmap(df, confirmed_admissions_covid_1d, confirmed_admissions_influenza_1d)

# Custom data with multiple other_keys ---
df2 <- expand.grid(
  geo_value = c("ca", "ny", "tx"),
  time_value = as.Date("2023-01-01") + 0:30,
  other = c("source_a", "source_b")
) %>%
  mutate(cases = rnorm(n())) %>%
  as_epi_df(other_keys = "other")

# Dropdown by 'other', facet by 'geo_value'
autoplot(df2, cases,
  .facet_by = "geo_value",
  .facet_to_dropdown = TRUE, .interactive = TRUE
)

# epi_archive support ---
# Fetching archive data
df_arch <- pub_covidcast(
  source = "hhs",
  signals = c("confirmed_admissions_covid_1d",
              "confirmed_admissions_influenza_1d"),
  geo_type = "state",
  time_type = "day",
  geo_values = "ar,ia,ca,ny,tx,ks,co,mi,pa",
  time_values = epirange(20221001, 20230101),
  issues = epirange(20221001, 20230101)
) %>%
  rename(version = issue) %>%
  as_epi_archive()

# Visualizing daily/weekly versions
autoplot(df_arch, confirmed_admissions_covid_1d,
         .versions = "week", .mark_versions = TRUE)

# Interactive archive with dropdown, selects all numeric
autoplot(df_arch,
         .interactive = TRUE, .facet_to_dropdown = TRUE)
```

R/epi_df.R
brookslogan: This feels more natural when you want to pull a single signal from epidatr and keep on calling things "value" for some short experimentation.

suggestion:
- Make `"auto"` with a <=1 unique signal-name col value also add the signal-name colname to `other_keys`, like with `"long"`.

musing: 
- Also in above case, should we be messaging something?  Or only in the auto-widening case?  I guess that would make some example usage kind of noisy.  We can try out the current no-messaging approach.
- Eventually we should also rename parameters related to "signal" to use "indicator" instead.  But I think this needs to hold off until new ETL stuff hits epidatr & we start using "indicator" there.

R/archive.R
brookslogan: `epix_pivot_wider` looks good.

todo: address usage of non-data.table-functions on data.tables.  r-lib and tidyverse packages aside from dtplyr aren't really aware of data.table's memory model or metadata.  E.g., `select`ing away a key column with dplyr doesn't update the key metadata like `[.data.table` does, so we end up with invalid intermediate data.table metadata.  Either
    - (a) convert to tibble via `%>% as.data.frame() %>% as_tibble()` (need both steps due to data.table bug in some versions) before the vec_split
    - (b) use only data.table ops on the data.table.  E.g., split/nest with `DT[, .SD, by = c(names_from)]` (the `c` here acts kind of like `.env$` in dplyr), and select/rename with `[` and/or `data.table::setnames`.
    - (c) add comment that `vec_split` appears not to violate data.table memory model, even with just 1 signal name, and that even though `dplyr::select` outputs a `data.table` with invalid `key` metadata, it's all right because `new_epi_archive` resets the key.

brookslogan (approved):
Thanks!  Couple of todos above; please merge when ready!

R/archive.R
brookslogan: Just going with (a) since it's easy, even if it's potentially slow.  We can fix perf separately if it's an issue.

R/epi_df.R
brookslogan: Pushed the "auto" change above.

